PR #20, opened 2017-08-05
Display script improvements

meyer9:
- Added context menu items for editing scripts
- Update display script every frame


cmeyer:
Updating display script was already merged.

Additions to context menu should be a separate pull request.

PR #181, opened 2018-12-04
added ui for string variables to ComputationPanel and Inspector

### LINKED ISSUE #180: Variable type "string" missing in computation panel
After adding a new variable via the respective button in the computation panel you cannot select type "string" for the variable. This also leads to the fact that no input field is created when a pre-defined computation specifies a variable of type "string", so there is no way to get usable string parameters in computations.

Brow71189:
this resolves #180


cmeyer:
8bb90aa70280b4189e3ac8a3ab60a156091c04aa

Issue #87, opened 2017-10-13
Interactive scripts display query UI until next query


cmeyer:
For instance, asking for an input will leave the input UI open until the next input.

This is a regression after changes to fix keyboard handling crashes. See ef34d304595d25b414ab52b4ff4c917a0185a17d.

Issue #152, opened 2018-09-21
Run Script dialog can produce multiple stacked cancel buttons


cmeyer:
From Dimitry: Apr 12, 2018

Astig purify does this regularly.

![image](https://user-images.githubusercontent.com/34664/45907464-d62b8100-bdac-11e8-81e4-15170fa80071.png)

PR #235, opened 2019-03-05
Fix prompts in scripts dialog

Brow71189:


Brow71189:
fixes #87 and maybe #152

Brow71189:
I put both buttons there on purpose because I don't think they are redundant. They do different things, the top one cancels the string enter field and the bottom one sets the interactive.cancelled flag.
Try getting out of this script in the master branch:
```
def script_main(api_broker):
    interactive = api_broker.get_interactive(version="1")
    while not interactive.cancelled:
        interactive.alert('Try cancelling me.')
```
Also, the always-present cancel button sometimes happened to be on top of a prompt and sometimes not. I would consider the always-present one part of the script dialog, so I think it should always be there and always in the same place.

cmeyer:
When the user is being prompted in the dialog above with the always-present Cancel button, pushing the 2nd Cancel button (bottom right) has no effect. So even if we decide to make both buttons present, it should have an immediate effect. Also, we could put it on the same line as the prompts and call it "Cancel Script" to avoid confusion with the prompt Cancel button.

Brow71189:
Ok that needs quite a bit of rework then, especially if we want the cancel button to always have an immediate effect. Although I actually like the behavior now, where you can use it to make sure the script will be cancelled (gracefully) after the next iteration of something that runs in a loop.

cmeyer:
Buttons that don't do anything are frustrating. To make it do something (I'm not sure I like this idea) clicking Script Cancel could put up text that says "Script will end at next cancellation point." Of course then the user would expect to uncheck it.

Overall, I like the original design which is that the Cancel Script button is only present when there is no prompt present. It's not a perfect solution, but it avoids extra, confusing buttons.

Another option (brainstorming): Get rid of cancel script altogether and use window closing and/or Ctrl+C to cancel.

Brow71189:
Script cancel by itself also does not do anything. It will just set the interactive.cancelled flag. So having it put out a message by default is also not great, I think. So also in the original design you will have a frustrating button that does not do anything if the user doesn't use the interactive.cancelled flag in hist script.
I'm not sure Ctrl+C will be good for everyone. I think there are people who like to click on things instead of using the keyboard. And also closing the window is not a good replacement. The nice thing about the cancel button is that you can stop the script, quickly change something and run it again without having to go through the long list of scripts.
But I can just make the cancel button non-permanent if that's your only issue with this branch and then we could get the other improvements into Swift soon.

Brow71189:
I made the script cancel button non-permanent.
Also I changed `add_task` to `queue_task` when resetting the message column (after the user acknowledges or rejects a prompt). This makes sure the command does always get executed.
Otherwise, when doing something like `print(confirm_ok_cancel(...))` there is a chance that `print` overwrites the task of resetting the message column.

Brow71189:
No I changed "queue_task" back to "add_task" when resetting message_column to its default state to make sure we can clear the tasks when the window is closed. The different id's for the tasks should make sure they get always executed.

cmeyer:
Squashed and merged. 9aa84d65e569222b6bed8e46c4ac5f3918426976.

PR #237, opened 2019-03-13
Improved console

Brow71189:
Lots of small improvements to the Swift builtin console:

- FIXED: When the cursor is in a line where the text is not white and you hit "enter", the new line has the same color as the line where the cursor was before.

- FIXED: Starting an assignment (e.g. a = api.) and hitting <TAB> removes everything until the equals sign when there is more than one suggestion for the completion. Similar behavior for starting a function call and hitting <TAB> after the opening parenthesis.

- IMPROVED: Overall behavior of <TAB> key, for instance you can now use tab to indent code when the cursor is not at the last position in the last line.

- IMPROVED: Behavior of up/down arrows when moving through history is now very similar to how bash does it. Most importantly unfinished commands (i.e. "enter" was not pressed) will be kept as a temporary history.

- CHANGED: The interpretation, auto-completion and history are now in a separate class. "ConsoleDialog" only handles display-related tasks, everything else is handled by the new class "ConsoleDialogController".

### DIFF
--- a/nion/swift/ConsoleDialog.py
+++ b/nion/swift/ConsoleDialog.py
@@ -122,7 +122,7 @@ class ConsoleWidget(Widgets.CompositeWidgetBase):
         if len(result) > 0:
             self.__text_edit_widget.set_text_color("red" if error_code else "green")
             self.__text_edit_widget.append_text(result[:-1])
-            self.__text_edit_widget.set_text_color("white")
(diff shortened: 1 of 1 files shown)


cmeyer:
Merged ffdfe460fb245522bdfcaffa4f440c63efbfa6eb

PR #244, opened 2019-04-12
Using 'jedi' for auto-completions if available.

Brow71189:


Brow71189:
I have no idea why the tests fail here. It seems to be completely unrelated (also they only fail for python 3.6).

cmeyer:
I don't think it's a good idea to suggest people modify their Python environments to install jedi as this is likely to be confusing and/or cause problems in the future. Nor do I think we should print "Congratulations" if they already have it installed. Again... confusing: Why am I being congratulated for opening a script window with an auto completer?

How about making startup silent for now and then adding an issue to print some helpful general information (such as the more important link or text to get more information about writing console scripts) at which point we can add status info about the auto-completer?

Brow71189:
Congratulations for using the superior auto-completion system is what I meant by that :)
I'll remove the messages and push an updated version.

ericpre:
Maybe add jedi as a required dependency and used it as default? It is a pure python library and available in the default channel of anaconda, so it should be transparent when installing nionswift.

cmeyer:
That's a possibility (adding jedi as a dependency), but I try to keep the required dependencies to a minimum; and it's for an extremely minor feature (for most users other than developers anyway), so I'd rather keep it separate for now.

Brow71189:
It was a request at the last Swift school, so people seem to want it. But
it might still be good to keep it separate for now, especially because it
uses a "hacky workaround" for a bug in jedi at the moment.

cmeyer:
Since this still seems to have a "hacky workaround" and it also pulls in another dependency (jedi), let's move this to a package. We can add a `Registry` based hook to plug-in an auto completer from a package. Then you can create the package independently of Swift and publish to PyPI/conda-forge. I'm closing this pull request.

PR #265, opened 2019-07-28
Fix bug where cursor position would not display on composite line plots

### LINKED ISSUE #200: Cursor position should be visible over composite line plot.

meyer9:
Fixes #200.


cmeyer:
Merged dcadae9bc32b0d19a3b13907a85bc3fa8c4cc3c8

PR #270, opened 2019-08-19
Add ability to copy SVG to external program

meyer9:
This copies SVG MIME data into the clipboard along with the previous Nion MIME types.


cmeyer:
Merged

PR #272, opened 2019-08-26
Added ability to reorder line plot layers by dragging legend

meyer9:
This allows users to reorder layers by clicking and dragging the legend.


meyer9:
This should be ready for review now. Currently, it only supports dragging a layer on one display item, however, it can be extended in the future to allow dragging a layer from one display item to another.

cmeyer:
It's no longer possible to drag one entire line plot onto another. In addition to fixing it, we should also add a test for that case to prevent future regressions.

meyer9:
Force pushed to remove merge commit and rebase with master instead.

meyer9:
Almost everything is done here except for the dynamic height and width of the legend to allow dropping display items to replace the current panel. Currently, the legend takes up almost the entire plot, but it should be updated as needed.

It can update when:
- the legend items are changed
- the drag position is changed
- the drag leaves the component
- the drag enters the component

meyer9:
Commit f5b283d may be easier than the current head to edit as it doesn't have as much WIP code.

meyer9:
I fixed the bug with dragging layers onto a line plot with a legend. The sizing is now updated as needed. There are a few more things I need to test/reimplement:


- [ ] Use midpoint of item as cutoff for where to place the new item.

The top half of the first item will insert the item before the first item, between the middle of the first item and the middle of the second item should place between the first item and second item, etc.

- [ ] Fix dragging between panels
  - [ ] Fix previewing drag in destination channel
  - [ ] Fix logic with dropping item in new location

meyer9:
Seems to be working consistently now, but I have to define the behavior of the legend when there is only one layer. Usually, the legend shouldn't show up, but then how would it deal with drag mechanics.

I'm thinking of doing it like this: hide the legend when there is only one display layer and then modify the line graph canvas item to allow dropping in layers. When a layer is dropped into a line graph, it moves the layer in (causing the legend to show up). Also, there probably should be a setting to show or hide the legend manually, but this should work for now.

meyer9:
Need to fix tests and ensure that the layer doesn't already exist when being added.

cmeyer:
This looks good as a first implementation of a difficult feature. A few minor changes for this PR set:

- rename `should_handle_for_mime_data` to `on_wants_drag_event` to be consistent with other method names

- rename `should_handle_for_mime_data` to be `wants_drag_event` or `handle_wants_drag_event` to be consistent with other naming

- `LineGraphLegendCanvasItem.delegate` should be `LineGraphLegendCanvasItem.__delegate` - keep fields private unless they need to be exposed.

- Put newlines at the end of `LinePlotCanvasItem.py` and `MimeTypes.py`

Optional for this PR but could be delayed until future:

- add tests for basic functionality

Bugs for future PRs (I will enter these as bugs after this PR is merged):

- dragging layer to another line plot without a legend works, but doesn't highlight the drag destination

- dragging layer between line plots doesn't undo/redo properly. needs a test and a fix.

- drag icon should be the color of the layer. right now it's just default/empty gray rect outline.

meyer9:
I fixed the basic problems and I'll add some basic tests this afternoon.

cmeyer:
Merged. c6c71455dd41acb5ebd0e9c3848a978aec69449e

PR #333, opened 2020-01-11
Add MANIFEST.in to include missing files in the source distribution.

ericpre:
Files are missing from the source distribution (.tar.tz) on pypi. Add MANIFEST.in to include them.


cmeyer:
See commit fbee4670ccc41cc42cdbdd2dc42c650dd33d640c

PR #334, opened 2020-01-21
Add support for folders to Run Script. Add script folders to sys.path.

Brow71189:
This is the implementation of what we discussed via Email last week.
New features:

- You can now add folders to the Run Script system. Every top-level python file in a folder will be treated as a script and shown underneath the folder
- Folders are automatically added to `sys.path` so that you can import from packages that are inside script folders
- Modules from script folders will automatically be reloaded (via `importlib.reload`) before a script is executed
- Scripts are now sorted by script name only instead of their full path. Additionally, the path to a script is only shown in small and gray font after the script name.

### DIFF
--- a/nion/swift/ScriptsDialog.py
+++ b/nion/swift/ScriptsDialog.py
@@ -8,6 +8,11 @@ import gettext
 import os
 import threading
 import traceback
+import typing
+import re
+import locale
+from importlib import reload
+import sys
 
 # typing
 from typing import AbstractSet
@@ -24,6 +29,7 @@ from nion.ui import Dialog
 from nion.ui import Widgets
 from nion.utils import Converter
 from nion.utils import Selection
+from nion.utils import Geometry
 
 _ = gettext.gettext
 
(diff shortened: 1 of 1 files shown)


Brow71189:
Changes in the additional commits:
- The script list refreshes when a new script/folder is added or deleted and when a folder is toggled
- When refreshing the scripts list, check if the files/folders in it still exist. If not, show a hint next to the script/folder name
- List nodes are serialized to a json string before saving them. Serialized nodes include a version number that can be used to make sure that reconstructing the class from its serialized representation is successful.

cmeyer:
Merged 8cfbd75d7c136ea945ffd55f0e795de9a62e8bb3..2dd6b020c1f63a878f43d0ef23a0be0b388cd6b7

...but the visual presentation needs some work. The twist downs should be in their own column; the file-scripts should have a file "icon" next to them; and the text should be aligned vertically with the folders. See how Windows explorer does it; or the Project panel in Swift.

Also, I'll soon add requests for right click to "reveal in Explorer", etc. See how context menus are handled in the Project panel.

Issue #46, opened 2017-09-04
Add optional calibration axis tick marks, labels, coordinates


cmeyer:
### See Also
- https://github.com/nion-software/nionswift/issues/1026

PR #354, opened 2020-02-21
Added support for RingGraphic to move and be used for other masks.

shabihsherjeel:


cmeyer:
This should already be supported by changing the coordinate system. The ring graphic is always centered at the origin of the coordinate system.

I think it introduces all sorts of problems if it isn't symmetric around the origin. What's the use case of this new functionality [being moveable]?

shabihsherjeel:
This is part of the libertem integration hackathon we had with Andreas. I do not know exactly what kind of processing this needs to do. But Andreas, @Brow71189, might be able to explain this better.

Brow71189:
We would like to be able to use the mask as a virtual detector for 4D STEM Maps. I know that you can change the origin of the coordinate system to move the graphic but this is a very awkward way to move a virtual detector.
I haven't looked at the code yet, but the idea was to have an extra ring graphic that is movable. Like this the original fourier mask would be unaffected.
Do you think having a moveable ring graphic causes problems by itself, or just when using it as a fourier mask.

shabihsherjeel:
As of right now, the code replaces the RingGraphic class. I did this to make it the same tool that is clicked from the toolbar. I could add modes or make a new class if you wish to keep the default "Fourier mask" functionality.

Brow71189:
I think it would be good to have a separate class and then a separate
toolbar button and menu entry.

Sherjeel Shabih <notifications@github.com> schrieb am Sa., 22. Feb. 2020,
13:16:

cmeyer:
Another possibility is to have a tool to adjust the underlying coordinate system - we could provide a display of the coordinate system and a tool to drag it around for adjustment (but not on actual FFT data). One complexity of that would be that we would be modifying an intrinsic property of the data (it's specified origin) and we need work that out with the 4D STEM data sets (where would that modification be stored?).

In general, are the masking tools only useful in Fourier/diffraction/frequency domain? Or are they useful spatial or temporal domains too? I've been assuming frequency domain only, in which case a proper origin is required.

I think the tests are failing because `get_mask` doesn't take into account the position. And if we decide that we want to allow masking tools to allow position offsets (not my vote right now), then the position should be treated as an offset from the origin, not as the origin itself. Doing that allows the position to be initialized to 0, 0 instead of 0.5, 0.5.

Brow71189:
I think the masking tools are only useful in the diffraction/frequency
domain. There might be an application for the spatial domain, too but I
can't think of an example right now.
In any case I think the ring graphic is complementary to the ellipse
graphic when you think of them as virtual detectors (ellipse -> BF, ring
tool -> annular detector). So from a user's point of view they should also
behave the same way which means you should be able to position them freely.
I think adjusting the coordinate offset introduces more problems than it
solves (like modifying the original data, etc). We could make the masks
always be centered on the coordinate origin when they detect that the data
they are working on is the result of a Fourier transform, and let them move
freely otherwise.

Chris Meyer <notifications@github.com> schrieb am Sa., 22. Feb. 2020, 16:44:

cmeyer:
If we allow the ring graphic to be offset, then I think we should allow all of the mask graphics (spot, ring, wedge, lattice) to be offset from the origin. And I think it should be implemented as an offset, not a fixed position. This allows us to keep the calibrated origin functionality as-is. In fact, we should probably have a base class `MaskGraphic` which has the offset field.

Anyway, my concern in doing this is that we're allowing users to too easily mess up the regular use case, which is the mask filters on FFT data, in which case the offset should _always_ be zero. It may be difficult to detect which type of data the mask is being applied in the graphic itself.

Also, before we give up a tool to drag the origin - I've been considering making the "displayed calibration" something separate from the "intrinsic calibration". The "displayed calibration" would be stored with the display (or display data channel) instead of the data. This may be useful when the user wants to change how they show calibrations (i.e. spatial vs angular vs temporal for scans). I haven't thought this fully through, but it might be possible to use this as the mechanism to change the origin of the STEM 4D data.

cmeyer:
Closing this since it doesn't pass tests and isn't something I think should be supported a the moment anyway. Please see comments above and  #362 and #46.

PR #361, opened 2020-03-26
Add support for running 'pick' on sequences of spectrum images.

Brow71189:
This is in preparation for and updated `MultiAcquire` that will allow users to acquire sequences of spectrum images.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -3,6 +3,10 @@
 Release Notes
 =============
 
+Version 0.15.0 (UNRELEASED)
+---------------------------
+* (2020-3-26) Allow pick computations for sequences of spectrum images.
+
 Version 0.14.8, November 27, 2019
 ---------------------------------
 * (2019-11-25) Improve display of sequence measurements.
--- a/nion/swift/model/DocumentModel.py
+++ b/nion/swift/model/DocumentModel.py
@@ -2123,6 +2123,43 @@ class DocumentModel(Observable.Observable, ReferenceCounting.ReferenceCounted, D
     def make_data_item_with_computation(self, processing_id: str, inputs: typing.List[typing.Tuple[DisplayItem.DisplayItem, typing.Optional[Graphics.Graphic]]], region_list_map: typing.Mapping[str, typing.List[Graphics.Graphic]]=None) -> DataItem.DataItem:
         return self.__make_computation(processing_id, inputs, region_list_map)
 
+    def __digest_requirement(self, requirement: typing.Mapping[str, typing.Any], data_item: DataItem.DataItem) -> bool:
+        requirement_type = requirement["type"]
+        if requirement_type == "datum_rank":
+            values = requirement.get("values")
+            if not data_item.datum_dimension_count in values:
+                return False
+        if requirement_type == "datum_calibrations":
+            if requirement.get("units") == "equal":
+                if len(set([calibration.units for calibration in data_item.xdata.datum_dimensional_calibrations])) != 1:
+                    return False
+        if requirement_type == "dimensionality":
+            min_dimension = requirement.get("min")
+            max_dimension = requirement.get("max")
+            dimensionality = len(data_item.dimensional_shape)
+            if min_dimension is not None and dimensionality < min_dimension:
+                return False
+            if max_dimension is not None and dimensionality > max_dimension:
+                return False
+        if requirement_type == "is_sequence":
+            if not data_item.is_sequence:
+                return False
+
+        if requirement_type == "bool":
+            operator = requirement["operator"]
+            for operand in requirement["operands"]:
+                requirement_satisfied = self.__digest_requirement(operand, data_item)
+                if operator == "not":
+                    return not requirement_satisfied
+                if operator == "and" and not requirement_satisfied:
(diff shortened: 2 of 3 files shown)


Brow71189:
For some reason the tests are always run with numpy 1.15 for python 3.6 and 3.7. They fail because the `where` keyword argument was introduced in numpy 1.17 (https://docs.scipy.org/doc/numpy/reference/generated/numpy.sum.html#numpy.sum)

nion/swift/test/DocumentModel_test.py
```diff
@@ -1977,7 +1977,7 @@ def test_undelete_connection(self):
             display_item = document_model.get_display_item_for_data_item(data_item)
             pick_data_item = document_model.get_pick_new(display_item)
             pick_display_item = document_model.get_display_item_for_data_item(pick_data_item)
-            self.assertEqual(1, len(document_model.connections))
+            self.assertEqual(2, len(document_model.connections))
```
cmeyer: Why would there be two connections now instead of one?

Brow71189: That's because this test is using `pick`, which has a second connection now. This is used to connect the sequence index of the spectrum image to the sequence index of the pick result. So a sequence of spectrum images will result in a sequence of pick data. The connection is always initialized, also for non-sequence spectrum images. See here for the code: https://github.com/Brow71189/nionswift/blob/073dae61216736879b72a21748bf46b6f578c3bd/nion/swift/model/DocumentModel.py#L2535

cmeyer: I'm going to take a bit of time to think about this. I'm trying to eliminate connections overall - but I need to review my previous notes on this.

Brow71189: Sure if there is a better replacement for connections I'm happy to use it. But I guess we need them in some way in order to do things like connecting a pick graphic to a display slice. Or are planning to replace this with a different system?

docs/release_notes.rst
```diff
@@ -3,6 +3,11 @@
 Release Notes
 =============
 
+Version 0.14.9 (unreleased)
+---------------------------
+* (2020-3-26) Allow pick computations for sequences of spectrum images.
+* (2020-3-26) Add boolean logic for computation requirements.
```
cmeyer: This isn't visible to users so it doesn't need to be here.

docs/release_notes.rst
```diff
@@ -3,6 +3,11 @@
 Release Notes
 =============
 
+Version 0.14.9 (unreleased)
```
cmeyer: This will be version 0.15.0 (UNRELEASED)

nion/swift/model/DocumentModel.py
```diff
@@ -2363,8 +2384,14 @@ def __make_computation(self, processing_id: str, inputs: typing.List[typing.Tupl
...
-                    connection = Connection.PropertyConnection(display_data_channel0, connection_src_prop, new_regions[connection_dst], connection_dst_prop, parent=new_data_item)
-                    self.append_connection(connection, project=project)
+                    connection = None
+                    if "region" in connection_dst:
+                        connection = Connection.PropertyConnection(display_data_channel0, connection_src_prop, new_regions[connection_dst], connection_dst_prop, parent=new_data_item)
+                    elif connection_dst == "display_data_channel":
+                        dst_display_data_channel = new_display_item.display_data_channel
+                        connection = Connection.PropertyConnection(display_data_channel0, connection_src_prop, dst_display_data_channel, connection_dst_prop, parent=new_data_item)
+                    if connection:
+                        self.append_connection(connection, project=project)
```
cmeyer: This code is confusing -- and I think it fails for the `map_connection` which has a `connection_dst` of 'pick_point'. Can you put the `connection_dst == "display_data_channel"` first and then use `else` for the original version? This might need some tests.

Brow71189: I made the change you requested. As far as I can see the tests should be written for `nion/swift/model/Processing.py`, right? At least if we want to test that `map_connections` are established correctly. I couldn't find any place where the `Processing.py` code is actually used, so I'm not sure how exactly to write tests for that.

nion/swift/test/DocumentModel_test.py
cmeyer: I'm planning to deprecate all of the persistent Connections and replace with implicit connections that get built at load time. The computation has all of the required information to build the connection dynamically at load time. It's less general, but much lower maintenance. For now, this patch looks ok with two connections, though. I will update it when I make the implicit connection change.

nion/swift/model/DocumentModel.py
cmeyer: Processing is used for the FFT windows at the moment; but they don't hit this code. I have an internal branch that uses this code - so I will add tests for it. Thanks.

nion/swift/test/DocumentModel_test.py
cmeyer: Note to future self: this will switch to implicit connections. Ok for now.

cmeyer:
Waiting for https://github.com/nion-software/niondata/pull/14

PR #371, opened 2020-04-14
Change 'time.gmtime(0)' to 'time.localtime(0)' in 'FileStorageSystem.py'. Fixes #370.

### LINKED ISSUE #370: Crash at start because of 'time.gmtime(0)'  in 'FileStorageSystem.py'
Depending on the timezone that is set and the exact implementation on the platform the error below prevents Swift from starting.

My system: Windows 10 and timezone UTC+01

See also #363, https://docs.python.org/3/library/time.html#time.mktime and https://stackoverflow.com/questions/2518706/python-mktime-overflow-error.

```
File "c:\cygwin64\home\andi\git\nionswift\nion\swift\Application.py", line 166, in start
document_model = DocumentModel.DocumentModel(profile=profile)
File "c:\cygwin64\home\andi\git\nionswift\nion\swift\model\DocumentModel.py", line 591, in __init__
self.__profile.read_profile()
File "c:\cygwin64\home\andi\git\nionswift\nion\swift\model\Profile.py", line 477, in read_profile
project_reference.read_project(self.projects, self.profile_context)
File "c:\cygwin64\home\andi\git\nionswift\nion\swift\model\Profile.py", line 147, in read_project
self.__project.prepare_read_project()  # sets up the uuid, used next.
File "c:\cygwin64\home\andi\git\nionswift\nion\swift\model\Project.py", line 197, in prepare_read_project
self._raw_properties = self.__storage_system.read_project_properties()  # combines library and data item properties
File "c:\cygwin64\home\andi\git\nionswift\nion\swift\model\FileStorageSystem.py", line 533, in read_project_properties
data_items_copy = sorted(properties_copy.get("data_items", list()), key=data_item_created)
File "c:\cygwin64\home\andi\git\nionswift\nion\swift\model\FileStorageSystem.py", line 530, in data_item_created
earliest_datetime = datetime.datetime.fromtimestamp(time.mktime(time.gmtime(0))).isoformat()
OverflowError
:
mktime argument out of range
```

Brow71189:


nion/swift/model/FileStorageSystem.py
```diff
@@ -527,7 +527,7 @@ def read_project_properties(self) -> typing.Dict:
                 properties_copy.setdefault("data_items", list()).append(data_item_properties)
 
         def data_item_created(data_item_properties: typing.Mapping) -> str:
-            earliest_datetime = datetime.datetime.fromtimestamp(time.mktime(time.gmtime(0))).isoformat()
+            earliest_datetime = datetime.datetime.fromtimestamp(time.mktime(time.localtime(0))).isoformat()
```
cmeyer, line 530: Can you try `earliest_datetime = datetime.datetime.utcfromtimestamp(time.mktime(time.gmtime(0))).isoformat()` instead? All times are stored as GMT so I think this will be a cleaner solution as long as it works.

Brow71189: This also doesn't work because the error happens in `time.mktime`, so it doesn't matter which function from the `datetime` module we're using.

Brow71189: The actual problem is that on timezones with a "positive" offset (i.e. UTC+01, like central europe), `time.mktime(time.gmtime(0))` would be a negative number, which is not supported internally, hence the `OverflowError`.

cmeyer: Sorry I missed these comments - what about just `earliest_datetime = datetime.datetime.utcfromtimestamp(0).isoformat()`

cmeyer: My 2nd suggestion doesn't work. I've found a way to reproduce this error. I'll fix it separately.

cmeyer:
See 039b6bfcb9d33087ad4e6eaa02c8952ea031a3ca.

PR #372, opened 2020-04-14
Removed requirement '2d to 3d' from sequence integrate, trim and extract.

Brow71189:
This is required in order to be able to use these computations in sequences of spectrum images or 4D data sets. See also nion-software/niondata#15.


cmeyer:
Like the other bug, I'm ok with this since I don't think the old code checked data types accurately -- but in the future we might want to ensure that we're integrating the data dimensions only (I think?).

Brow71189:
I think we want to sum along the sequence axis, which means all other dimensions get summed.
I would also like to add a `concatenate` function (or rather change the exsiting one) that allows to concatenate an arbitrary number of data items. But we should probably discuss this separately.

PR #380, opened 2020-04-24
Ensure negative scale in line plots is handled properly. Fixes #130.

### LINKED ISSUE #130: 1D data items  (line plots) with negative scale show no data
To reproduce the issue type the following into the builtin python console:

```python
import numpy as np
di=api.library.create_data_item_from_data(np.arange(100))
```
creates a data item with a line. Everything is fine so far. If you then do:

```python
c=api.create_calibration(scale=-1)
di.set_dimensional_calibrations([c])
```

After this command the data disappears from the data item. You can still do everything as normal, but you just do not see the plot. So the data seems to be there  but is just not shown (also the labels at the axis are shown correctly). If you switch the display style to pixels or fractional the data appears. In calibrated mode it disappears again.

The issue occurs in Swift 0.12 and 0.13. It is not there in Swift 0.11 and earlier.

Brow71189:
I would also like to add a test for this, but I have no idea how you check that a plot actually appears in a line plot. If you can give me some hints I will write a test and add it to the pull request.


cmeyer:
Merged 3c0a8fb10f077e38932c3fffd0f1ca92a2b7f18c. I added a test if you want to look. I haven't tested drawing functions too often; but this one seemed easy -- it basically tries to draw the graph and checks that sufficient drawing commands were generated (it was had zero drawing commands when it was failing before; I chose 100 as a threshold for success). In the future we might consider improved tests for drawing.

PR #385, opened 2020-04-27
Make sure tick labels are not wider than drawing area for log line plots.

Brow71189:
The function that calculates the width of the y-axis tick labels widget was not aware of log-scale data. Since the data that gets drawn is the log of the actual data this width was too small for large values (for 10,000 it would calculate the width for a max value of 5). Now the system is aware of log-scaled data and calculates the width corretly.

### DIFF
--- a/nion/swift/LineGraphCanvasItem.py
+++ b/nion/swift/LineGraphCanvasItem.py
@@ -152,6 +152,18 @@ class LineGraphAxes:
         return self.__calibrated_data_max
 
     @property
+    def calibrated_value_max(self):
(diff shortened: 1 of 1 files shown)


Brow71189:
That's a weird test failure which I don't think is related to the change I made...

Brow71189:
This one is independent from the nionutils change, so you can merge it.

Chris Meyer <notifications@github.com> schrieb am Mo., 27. Apr. 2020, 18:22:

cmeyer:
Thanks!

PR #386, opened 2020-04-27
Interval graphic cursor style

Brow71189:


cmeyer:
This makes it much easier to understand what's going to happen - although the 8-pixel change seems too large and also makes it difficult to drag narrow intervals. See #389.

PR #391, opened 2020-04-29
Fixed a bug that caused data < 1 not being drawn in log line plots. Added test.

Brow71189:


cmeyer:
This change looks good and I'm ok merging it "as is".

However, for handling negative/positive data, our algorithm still clips to 1.0 if any negative data is present. Matplotlib handles it differently:

image

Our code: `show(numpy.arange(-2,5,0.01))`

image

Before merging this, can you confirm this is how you choose to do things (clipping to 1.0)?

I think the 1.0 clipping was put in place to sensibly handle background subtracted EELS spectra that for one reason or another went negative during the background subtraction.

Brow71189:
Actually if you move the y-axis up (so that you can also see smaller numbers), you will see that it looks pretty much like matplotlib, except that there is some weird "wrong" data where the original data is negative. I'll look into that and try to fix it.

cmeyer:
Ok - I'm merging this and adding a new bug.

cmeyer:
#392

Issue #382, opened 2020-04-27
Add key to toggle a "cursor" channel graphic on and off.


cmeyer:
Also select it and focus so that arrow keys can immediately be used to move it around. Perhaps a key to expand channel to an interval and a key to collapse to a channel.

Need to propose key mappings first.

cmeyer:
Cursor should display the value at the cursor position. See #383

Brow71189:
What do you think about a double click as a shortcut to add a channel graphic?

See also my comments in #381:

> 
> 
> Some thoughts on additional functionality:
> 
>     1. The label should default to the data value at this position. Right now, chances are high that people think the "Add channel graphic" button does not work because the tiny little red line is easy to miss.
> 
>     2. We need a way for the user to choose to which piece of data the line "belongs" (for multi-line plots). This is also something we have to think through for interval graphics.
> 
>     3. I think we could use a double-click as a shortcut to add a channel graphic. That's easy to remember and does not happen accidentally too often.
> 
>     4. We should probably also allow the user to change the appearance of it, i.e. change lenght/width of the line.

cmeyer:
I'm not immediately sold on double-click. Double-clicks are currently used to reset things - like double clicking the axes will reset them; double clicking histogram will reset. Traditionally double-clicks are "select + action", e.g. launch an application. Somehow double clicking to place a cursor doesn't seem consistent with either of these.

Right click: yes. That would be immediately useful.

I'm more inclined to do something like AutoCAD: a keypress (or tool button in the tool palette) to enter 'cursor position mode'; now the channel cursor follows the mouse until the mouse is clicked or the keypress is repeated. If the operation ends while the mouse is outside the line plot, nothing changes (cursor reverts to previous state).

cmeyer:
2+. We should decide whether an interval covers a single data layer or can be useful for multiple layers simultaneously. For instance, the interval could be data layer independent; and applying it to a specific data layer could be an additional operation. This is closest to how it works now.

Issue #338, opened 2020-02-07
Provide a way for plug-ins to access mouse location and listen to click events


Brow71189:
Someting on the lines of `document_controller.courser_position_changed_event`. I can see different ways to implement this, but I guess the most useful and simplest to use way would be to have an event in, say, document controller whose listener function would have the following form:
```
listener_fn(position: Geometry.FloatPoint, display_item: DisplayItem.DisplayItem)
```
For mouse clicks one could add `modifiers` to the arguments to give the plug-in acces to keyboard modifiers, too.

Brow71189:
As a sidenote:
Right now you have to do something like this to get the mouse position on a click:
```python
dc = api.application.document_controllers[0]

original_mouse_clicked = dc._document_controller.selected_display_panel.canvas_widget.on_mouse_clicked

def on_mouse_clicked(x, y, modifiers):
    original_mouse_clicked(x, y, modifiers)
    canvas_item = dc._document_controller.selected_display_panel.root_container._RootCanvasItem__mouse_tracking_canvas_item
    if canvas_item:
        canvas_item_point = dc._document_controller.selected_display_panel.root_container.map_to_canvas_item(Geometry.IntPoint(y=y, x=x), canvas_item)
        if hasattr(canvas_item, 'map_widget_to_image'):
            image_point = canvas_item.map_widget_to_image(canvas_item_point)
            # The following line prints the mouse click position in the image in pixels relative to top-left corner
            print(image_point)
   
dc._document_controller.selected_display_panel.canvas_widget.on_mouse_clicked = on_mouse_clicked
```

cmeyer:
What is the intended use of this? Mouse cursor handling has an extreme effect on UI responsiveness, so I'm apprehensive to have packages modifying it. Is the intended use something that should go into Swift anyway? Or is there another way to provide a similar functionality?

Brow71189:
The use case for this specific example is what the guys in Vienna wanted to do (see the email I forwarded you on Feb 3).
They want to know where in an image the user clicked to select and atom that was found by their algorithm. I guess they are drawing some sort of annotations into the image, since having hundreds of graphics in a data_item is too slow for their application.
I guess an alternative solution would be to speed up handling many graphics in an image. Because then the user would just select one and one could get the desired position from there.

Brow71189:
I get that cursor tracking can have a huge impact on UI performance, but how about mouse clicks? Most use cases of this issue would probably be covered by supporting listening to mouse click events.

[...]

PR #404, opened 2020-05-07
Add some graphics operations to context menu.

Brow71189:
This does not add a keyboard shortcut as requested in #382 but the context menu functionality we discussed there.
The context menu entries are dynamically adapted to the display type of the data panel and whether a graphic is currently selected or not.

### DIFF
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -2070,6 +2070,40 @@ class DisplayPanelManager(metaclass=Utility.Singleton):
 
         display_panel_type = display_panel.display_panel_type
 
+        if display_panel_type == "data_item" and display_panel.display_canvas_item:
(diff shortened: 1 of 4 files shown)


cmeyer:
Everything looks good on this... except it doesn't quite work as expected due to an upstream issue.

If you right click on a display that doesn't have keyboard focus, the action goes to the one with keyboard focus, not the one you clicked on. I'm delaying this merge until I address that issue.

cmeyer:
Also related #338 Plug-in tools

cmeyer:
Again, sorry for the long non-action on this PR. I'm closing this for a few reasons:
- it doesn't merge cleanly anymore
- it had issues in the original version that I can't tell if they were addressed or not
- it needs more thought and design
- #1043
- #1042

PR #413, opened 2020-05-13
Add complex display type chooser for images and line plots.

### LINKED ISSUE #64: Add combo box to select the complex display mode to inspector

### LINKED ISSUE #131: Display 1d complex data should allow choice of complex to display conversion

Brow71189:
Fixes #64. Fixes #131.

For line plots I put the selector in the `DisplayLayer` sections, because it can be useful to be able to have different complex display types in one plot. For instance, if you want to see the real and imaginary part at the same time, you could create a composite line plot and select "imaginary" and "real" for its two data channels.

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -729,6 +729,10 @@ class LinePlotDisplayLayersInspectorSection(InspectorSection):
                 self.content_widget.add(content_row)
                 self.content_widget.add(fill_color_row)
                 self.content_widget.add(stroke_color_row)
+                # complex display type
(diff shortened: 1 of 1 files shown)


cmeyer:
Merged in 9de0fff491a8dc253513cf498e0fa8b300f772f5 and 518ec7331deebba5d46c1f28de39aa110ad05167.

The 2nd commit is to fix a bug in how data was indexed. See the last few commits to understand. Mainly it was a bug on data shaped like 2x32 with multiple layers. This whole display layer mechanism will be reworked when I do composite raster images.

PR #414, opened 2020-05-14
Change LinePlot tickers to new classes.

Brow71189:


Brow71189:
The tests fail here because the change I made depends on my pull request in nion-software/nionutils#5.

I also added the second commit to this pull request here because it is a very small change that is related to log scale plots and not worth an extra branch + commit.
It fixes the awkward behavior where log scale plots would always zoom around 1 because this is 0 internally on the log scale.
For linear plots it is fine to zoom around 0, but on a log scale, 1 does not have any special meaning, so I think we always want to zoom around the bottom of the plot just like when 0 is outside the plot range on linear plots.
At least that was my feeling while working on the new tickers.

cmeyer:
This change looks great, with a few caveats (see below).

In the interest of moving this project forward, I merged these changes.

Merged 43a8a55469f8df617060bd717d09cd5d94b51cb7

However, I think more work needs to be done. I filed new bug reports here:

#417 text too dense vertically
#418 limits too wide

PR #415, opened 2020-05-14
Add capability to handle double clicks for DisplayPanelManager listeners.

Brow71189:


cmeyer:
I'll think about this but I will almost certain just close this PR. The image display clicks were a very-early-in-the-development hack and probably not something we should be using or extending.

cmeyer:
#338 Plug-in tools

cmeyer:
This shouldn't be needed after b9c0033b3cabcc0762e33263293a9cf8035cd79f. Please file another issue if you still have problems.

PR #420, opened 2020-05-15
Add superscript scientific notation on line plot axes.

cmeyer:
This was just too fun to pass up. Take a look @Brow71189.

image


cmeyer:
Do we need a checkbox to make this optional?

PR #457, opened 2020-05-25
Add option to highlight display panel and to show a popup message.

Brow71189:
This is required for the new features added in nion-software/nion-instrumentation#173 and nion-software/nionswift-instrumentation-kit#24.

It allows DisplayPanels to be "highlighted", which means they get an additional colored frame drawn around them. They can also show small popup messages that overlay the content of the DisplayPanel.


cmeyer:
Closing. Use window reports. See also b2f3adba2dda7cf3340ab385f1d0164e18dcc3e4 support for modal actions (tools).

PR #459, opened 2020-06-02
Fix import and export of data_items with 'large_format' flag. Add test.

### LINKED ISSUE #379: Importing large files should use "large_format" for the data_item that gets created.

Brow71189:
Fixes #379.


Brow71189:
This also ensures that data without associated metadata and with more than two dimension will be imported as "large_format" data.

FYI: No existing tests are failing with these changes. Also the DM3 IO tests are passing.

cmeyer:
Thanks!

PR #637, opened 2021-03-22
Add zarr handler and use it as default for new large-format data items.

Brow71189:
H5py 3.x is still only available on conda-forge which prevents us from using it in our default installation. But we actually desperately need the change they made which releases the GIL when reading/writing from/to disk.
But even the new version does not support multi-threaded reading/writing which is a big performance bottleneck.
To improve performance with large datasets I'd like to explore new options for storage backends. Here I made an attempt at a new handler using [Zarr](https://zarr.readthedocs.io/en/stable/) as a backend. It supports multi-threading and compression which allows for better performance with large datasets. Additionally it provides a lot of flexibility in how you arrange the data on disk. In this implementation I used the [NestedDirectoryStore](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.NestedDirectoryStore) layout which arranges the data in a set of nested directories. One can also choose from different database formats which offer even better performance especially when choosing smaller chunk sizes. Here I'm using chunks with 100 MB uncompressed size which make accessing large portions of the data very fast but can negatively impact speed when accessing very small portions.
I'm also using compression which reduces the data size on disk by at least a factor of 2. For a test dataset consisting of real expertimental data I'm getting a compression ratio of about 10 which is great when dealing with big multi-dimensional datasets.

This pull request is not neccessarily intended to be merged as-is but more as an experiment to try an alternative for our hdf5 storage backend.

### DIFF
--- a/nion/swift/model/FileStorageSystem.py
+++ b/nion/swift/model/FileStorageSystem.py
@@ -770,7 +770,7 @@ class FileProjectStorageSystem(ProjectStorageSystem):
                 # test_delete_and_undelete_from_file_storage_system_restores_data_item_after_reload
                 try:
                     file_path.unlink()
-                except PermissionError:
+                except (PermissionError, IsADirectoryError):
                     try:
                         shutil.rmtree(file_path)
                     except:
@@ -802,7 +802,7 @@ class FileProjectStorageSystem(ProjectStorageSystem):
                     os.makedirs(os.path.dirname(target_storage_handler.reference), exist_ok=True)
                     try:
                         shutil.copyfile(storage_handler.reference, target_storage_handler.reference)
-                    except PermissionError:
+                    except (PermissionError, IsADirectoryError):
                         shutil.copytree(storage_handler.reference, target_storage_handler.reference, dirs_exist_ok=True)
                     target_storage_handler.write_properties(Migration.transform_from_latest(copy.deepcopy(properties)), datetime.datetime.now())
                     logging.getLogger("migration").info(f"Copying data item ({index + 1}/{count}) {data_item_uuid} to new library.")
--- a/nion/swift/model/FileStorageSystem.py
+++ b/nion/swift/model/FileStorageSystem.py
@@ -800,7 +800,10 @@ class FileProjectStorageSystem(ProjectStorageSystem):
             with contextlib.closing(self._make_storage_handler(old_data_item, file_handler)) as target_storage_handler:
                 if target_storage_handler and storage_handler.reference != target_storage_handler.reference:
                     os.makedirs(os.path.dirname(target_storage_handler.reference), exist_ok=True)
-                    shutil.copyfile(storage_handler.reference, target_storage_handler.reference)
+                    try:
+                        shutil.copyfile(storage_handler.reference, target_storage_handler.reference)
+                    except PermissionError:
+                        shutil.copytree(storage_handler.reference, target_storage_handler.reference, dirs_exist_ok=True)
                     target_storage_handler.write_properties(Migration.transform_from_latest(copy.deepcopy(properties)), datetime.datetime.now())
                     logging.getLogger("migration").info(f"Copying data item ({index + 1}/{count}) {data_item_uuid} to new library.")
                     return ReaderInfo(properties, [False], self._is_storage_handler_large_format(target_storage_handler),
--- a/nion/swift/model/FileStorageSystem.py
+++ b/nion/swift/model/FileStorageSystem.py
@@ -14,6 +14,7 @@ import uuid
 
 from nion.swift.model import DataItem
 from nion.swift.model import HDF5Handler
+from nion.swift.model import ZarrHandler
 from nion.swift.model import Migration
 from nion.swift.model import Model
 from nion.swift.model import NDataHandler
@@ -653,7 +654,7 @@ class ProjectStorageSystem(PersistentStorageSystem):
 
 class FileProjectStorageSystem(ProjectStorageSystem):
 
-    _file_handlers = [NDataHandler.NDataHandler, HDF5Handler.HDF5Handler]
+    _file_handlers = [NDataHandler.NDataHandler, HDF5Handler.HDF5Handler, ZarrHandler.ZarrHandler]
 
     def __init__(self, project_path: pathlib.Path, project_data_path: pathlib.Path = None):
         super().__init__()
@@ -717,7 +718,7 @@ class FileProjectStorageSystem(ProjectStorageSystem):
         return self.__find_storage_handlers(self.__project_data_path)
 
     def _is_storage_handler_large_format(self, storage_handler: StorageHandler.StorageHandler) -> bool:
-        return isinstance(storage_handler, HDF5Handler.HDF5Handler)
+        return isinstance(storage_handler, (HDF5Handler.HDF5Handler, ZarrHandler.ZarrHandler))
 
     def _remove_storage_handler(self, storage_handler: StorageHandler.StorageHandler, *, safe: bool = False) -> None:
         assert self.__project_data_path is not None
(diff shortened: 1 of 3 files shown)


TomaSusi:
I also had a look at Zarr recently and it does seem like probably the best alternative out there.

Brow71189:
> 
> 
> I also had a look at Zarr recently and it does seem like probably the best alternative out there.

Thanks for the feedback. Zarr is still pretty young but I agree that it looks very promising.

cmeyer:
I haven't looked at this code closely - and I'm generally positive about this format -- but I have a few comments.

First, you can modify the tests so that they all run with a specific storage handler. This is a good way to make sure it works with a good cross section of use cases. To do this, modify the `_file_handlers` at the top of the `FileProjectStorageSystem` class.

Next, generally, I'm extremely hesitant to introduce a new storage format. Since this is how we store intrinsic data for our projects, we are committing to support this storage format for the _lifetime of Swift_. It is not an export format; it is a fundamental storage technique for our project. This isn't a reversible decision. We become highly dependent on the format, the library, the implementation, its availability within Python, its implementation in other languages, etc. So this might be something we want to go slow on to make sure we get it right.

At the very minimum, I think we should have a set of criteria by which we evaluate potential new storage formats. Off the top of my head, this would include a performance suite, a survey availability of high quality implementations outside of Python, an understanding of who funds the project, a history of up-to-date releases within the Python ecosystem, the age and maturity of the library, a survey of the history of significant bugs, etc. (side comment: I'm not sure `h5py` lives up to these criteria; which is one of your reasons for migrating).

As an alternative, we could also try to reduce the impact of the storage formats by allowing users to specify which formats they are comfortable with for their long term data. This approach would need more thought too.

NetCDF is another format to consider.

sk1p:
> I also had a look at Zarr recently and it does seem like probably the best alternative out there.

It's interesting - just a few days ago, I also had a look at this format, and compared to HDF5, I'm quite happy with the API and performance. With zstd compression, I observed up to a ratio of about 22 on real data, with read performance only slightly degraded compared to uncompressed reading. When reading from my slow (~500MB/s) SSD, the compressed format should actually win out in that case, but I didn't do comprehensive benchmarking yet.

I also looked into using `tensorstore` as a wrapper around `zarr`, which adds a "lazy slicing" API on top, among other things. With "naked" zarr, I had some concurrency issues when writing data - I'm sure I just didn't follow the locking protocols in the correct way.

Brow71189:
> 
> 
> I haven't looked at this code closely - and I'm generally positive about this format -- but I have a few comments.
> 
> First, you can modify the tests so that they all run with a specific storage handler. This is a good way to make sure it works with a good cross section of use cases. To do this, modify the `_file_handlers` at the top of the `FileProjectStorageSystem` class.
> 
That's already in the pull request. I just didn't update `test_requirements.txt` yet which is why all tests fail.

> Next, generally, I'm extremely hesitant to introduce a new storage format. Since this is how we store intrinsic data for our projects, we are committing to support this storage format for the _lifetime of Swift_. It is not an export format; it is a fundamental storage technique for our project. This isn't a reversible decision. We become highly dependent on the format, the library, the implementation, its availability within Python, its implementation in other languages, etc. So this might be something we want to go slow on to make sure we get it right.
> 
I totally agree that this is not something we should decide lightly. I'm just trying to explore new options because HDF5 does not seem to be the best format for our purpose. We've been waiting forever for improvements that help with large files but they still haven't arrived. And even with the changes in hpy 3.x it still only offers a small subset of features from what other formats offer.

> At the very minimum, I think we should have a set of criteria by which we evaluate potential new storage formats. Off the top of my head, this would include a performance suite, a survey availability of high quality implementations outside of Python, an understanding of who funds the project, a history of up-to-date releases within the Python ecosystem, the age and maturity of the library, a survey of the history of significant bugs, etc. (side comment: I'm not sure `h5py` lives up to these criteria; which is one of your reasons for migrating).
> 
I like the idea of defining a set of criteria. That would allow us to systematically explore and rate different storage options. I thing I'll start something like this as a side project because I'm still convinced we need to at least consider some options besides hdf5/h5py.
In terms of age and maturity zarr will of course not be able to compete with h5py. I'm also not aware of any non-python implementations of the format. One thing to consider though is that zarr is not a specific implementation of a storage format, it is rather a specification for how to arrange chunked data within an arbitrary storage backend. So if there is a stable implementation of the chosen storage backend available, reading the zarr-array data is fairly easy in any language.

> As an alternative, we could also try to reduce the impact of the storage formats by allowing users to specify which formats they are comfortable with for their long term data. This approach would need more thought too.
> 
I'm not sure this is going to be very practical. It might be interesting for certain "expert users". But I don't think a normal user would know what to choose from for example "ndata", "hdf5" and "zarr". If you haven't really looked into these formats in a lot of detail you just can't know which one is the best.

> NetCDF is another format to consider.
Once we have a set of criteria we should just evaluate this option, too.

sk1p:
> I'm also not aware of any non-python implementations of the format

FWIW the zarr driver in tensorstore is written in C++: https://github.com/google/tensorstore/tree/master/tensorstore/driver/zarr including a C++ API you can build against: https://google.github.io/tensorstore/installation.html#c-api

Brow71189:
> 
> 
> > I also had a look at Zarr recently and it does seem like probably the best alternative out there.
> 
> It's interesting - just a few days ago, I also had a look at this format, and compared to HDF5, I'm quite happy with the API and performance. With zstd compression, I observed up to a ratio of about 22 on real data, with read performance only slightly degraded compared to uncompressed reading. When reading from my slow (~500MB/s) SSD, the compressed format should actually win out in that case, but I didn't do comprehensive benchmarking yet.
> 

Interesting that you also explored zarr independently. I didn't look into zstd before myself but now I also benchmarked zarr + zstd.
Before I've always used blosc with blosclz or lz4 compressor. Blosc has the nice feature that it supports seamless multi-threading. I also used zstd through blosc now for this reason. I get similar read speeds with zstd than with blosclz or lz4 but the write speed is a lot slower. The compression ratio of zlib is about a factor of 2 better than blosclz.

Reading from "slow" drives is actually a very good use case for compressed formats. So far I've always tested storage candidates on very fast SSDs (~ 2 GB/s nominal speed). On such drives I don't see a huge perfomance increase when using compression, but depending on the exact settings compression can still give you some improvements. Seeing these results I would definitely expect a quite big increase in performance for slower drives.

cmeyer:
See #662 #663 and #664 (plug-in storage, IO handlers, and expanded data item).

ndellby:
Don't these formats have compiled C readers without dependencies, so that even if the format goes obsolete, or as is more likely in the Python world updated and then deprecated and then unsupported, there will be a permanent plug-in that can read it.

cmeyer:
> Don't these formats have compiled C readers without dependencies, so that even if the format goes obsolete, or as is more likely in the Python world updated and then deprecated and then unsupported, there will be a permanent plug-in that can read it.

Many format do have just that. However, it looks like "not yet" for `zarr`. (Note: I didn't research this thoroughly).

https://github.com/zarr-developers/community/issues/9

Brow71189:
Another way to at least detect incompatible future versions of the format would be to include a small data item into our test suite and make a test that reads it. That test would fail when a future version is not backwards-compatible anymore thus notifying us about this problem

joshmoore:
> However, it looks like "not yet" for `zarr`. 

Sorry for jumping into your thread, but I saw the reference from zarr-developers/community#9 NetCDF-C has recently (4.8.0) added support for Zarr which may also cover this other point:

> https://github.com/nion-software/nionswift/pull/637#issuecomment-804222854 NetCDF is another format to consider.

(see also: https://twitter.com/zarr_dev/status/1379875020059635713)

cmeyer:
I'm closing this PR as outdated. We can use it for a reference if we return to zarr.

PR #678, opened 2021-04-16
Fix #388. Sort data item references first.

### LINKED ISSUE #388: Ordering in data panel during acquisition is inconsistent
Sorting is done with all live items first reverse-alphabetically; then by date. However, live items include dependents of acquisition items. So an FFT named "zFFT" can appear first in the list.

An improved sorting would be data channels sorted first alphabetically then by channel index; then live items alphabetically; then by date/uuid.

cmeyer:


cmeyer:
This PR has bad performance: `is_data_item_reference` is called for each displayed item when sorting.

This can be merged when there is a better solution for determining if a display item is representing a reference. This could happen as part of the data item rewrite where liveness and other characteristics will have to be moved out of the data item class anyway.

PR #682, opened 2021-04-20
Fix bug when dragging channel graphic outside data range. Fixes #644.

### LINKED ISSUE #644: Dragging a channel graphic outside of the data region breaks the line plot display
To reproduce take some line plot display and adjust the x-axis so that it goes outside the region where data is available. Now add a channel graphic to the line plot and try to drag it outside the data range. This will cause lots of error messages to run across the terminal and make the line plot display unresponsive to anything you're trying to do. The only way to recover is to delete the channel graphic from the built-in python console.
Note that interval graphics don't show this bug. They can be freely moved also outside of the data area in the same way as all other graphics in Swift can.

Brow71189:
I also tried to write a test for this, but I couldn't figure out how to make the system actually repaint the graphics on the line plot.

My test would look something like this:

```python
def test_channel_graphic_outside_data_range_does_not_fail(self):
        with TestContext.create_memory_context() as test_context:
            document_controller = test_context.create_document_controller()
            document_model = document_controller.document_model
            display_panel = document_controller.selected_display_panel
            data_item = DataItem.DataItem(numpy.arange(10))
            document_model.append_data_item(data_item)
            display_item.set_display_property('right_channel', 25)
            channel_graphic = Graphics.ChannelGraphic()
            channel_graphic.position = 0.99
            display_item.add_graphic(channel_graphic)
            display_panel.set_display_panel_display_item(display_item)
            display_panel.layout_immediate((640, 480))
            document_controller.periodic()
            display_panel.display_canvas_item.refresh_layout_immediate()
``` 

And then there needs to be some call that triggers drawing, I guess.

### DIFF
--- a/nion/swift/LineGraphCanvasItem.py
+++ b/nion/swift/LineGraphCanvasItem.py
@@ -621,7 +621,6 @@ class LineGraphRegionsCanvasItem(CanvasItem.AbstractCanvasItem):
             self.update()
 
     def _repaint(self, drawing_context):
-
         # draw the data, if any
         axes = self.__axes
         data = self.__calibrated_data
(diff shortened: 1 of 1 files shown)


cmeyer:
Merged 22faad92a26c01e2b21082c4f57eb068aaa29eba.

For testing, I usually try to isolate the drawing code to a single function. I started doing that here a173637f6eaa5b6835104ead2314ec29164e4138 but I decided the drawing code in the line plot needs more rework than I was willing to do at the moment.

The general procedure for isolating the drawing code is to move all the references to `self` to the top of the `_repaint` function and then move any common variables to the top also. Then what's left is the pure drawing code - it's easy to make it into a function that not dependent on the class.

In this case, though, I decided not to proceed because it would entail passing the regions, which are `RegionInfo` objects and defined in a higher level file, to the drawing code. Also, this drawing code probably needs a larger rethink - to make it more object oriented and also to have a common object to retain any cached drawing info that is to avoid repeated calculations required to draw multiple intervals/channels.

Brow71189:
I also thought about just testing the drawing function but abandoned the idea because of the `RegionInfo` objects being created somewhere else.
Thanks for the detailed explanation though!

Issue #399, opened 2020-05-04
Option to display exponents in axes as using a superscript


cmeyer:
![image](https://user-images.githubusercontent.com/34664/80991334-b368a780-8dec-11ea-9785-899571d3bbd6.png)

PR #696, opened 2021-06-01
Fix #681. Normalize string used in axis label size calculation.

### LINKED ISSUE #681: Line plot drawing unstable when dragging y-axis
`show(numpy.random.randn(100,) * -20000)` then drag the y-axis. The left edge of the plot / right edge of the y-axis captions jitters back and forth.

dhauge:
This is not complete yet, but I wanted to create the PR to start a discussion about the general approach. As mentioned in the issue, the problem is that
- the content size calculation for the label is based on an integer calculated from the min/max calibrated bounds
- that integer is not rounded in any way, so as the axis is dragged the digits are changed almost pixel-by-pixel
- if the font is not monospace, the calculated width varies depending on which digits show up in the number
- thus the calculated width will tend to vary every pixel or so of drag.

The implemented fix avoids this by replacing all digits with `'0'` (because in most fonts it seems to be at least as wide as other digits), so it only varies roughly if the order of magnitude changes. This might be a slight overestimation of the width for the exact number, but it shouldn't be too much of one at least when the tick labels are rounded so most digits are `'0'`. 

A few questions
- How much of an expectation is there that the layout of a line plot does not vary between versions by a few pixels.
-- This question would apply both for this fix, and for potential further fine-tuning in later releases.
- Is there any reason to fine-tune the normalization so, say, it does not normalize the `'1'`in the exponential form `' x 10'` .
- Is there a reasonable way in tests to create a `UISettings.UISettings` instance so `calculate_scientific_notation_drawing_width` can be unit tested in isolation, or is the preferred way to test it via an actual line plot being rendered. 
- I'm not sure I'll do it yet here, but what is the preferred route for refactors related to an issue? The code related to calculating the size and rendering effectively copies the following logic in several places
```
    labels = label.lower().split("e")
    if len(labels) == 2:
        labels[0] = labels[0] + " x 10"
        labels[1] = str(int(labels[1]))
```
It is not that complex, but I think rearranging the code so all paths go through a shared path to break apart the exponential case would make the overall logic simpler. I'm guessing this should be a separate PR, but does it need an issue created, or should it be related to the issue that motivated it?
- The same label normalization when fitting a label-based canvas item to content could potentially be applied to the other components that call `self.__ui_settings.get_font_metrics`, which I guess is just `LineGraphLegendCanvasItem`. I'm not sure the problem will be as apparent in that component, unless there is a similar drag operation that will update the legend label in a similar manner to the vertical axis dragging, so there might not be an apparent need, but it naively seems to me that calculation of size for numbers should be done consistently.

### TODO
- ~~Decide whether normalization should be fine-tuned to match string more.~~
- ~~Add unit tests~~

### DIFF
--- a/nion/swift/LineGraphCanvasItem.py
+++ b/nion/swift/LineGraphCanvasItem.py
@@ -11,6 +11,7 @@
 import collections
 import gettext
 import math
+import re
 import typing
 
 # third party libraries
@@ -969,12 +970,20 @@ class Exponenter:
             drawing_context.fill_text(label, width, y)
 
 
+_LABEL_SIZE_CALCULATION_NORMALIZE_RE = re.compile(r"\d")
+
+
+def _normalize_label_for_size_calculation(label):
+    return _LABEL_SIZE_CALCULATION_NORMALIZE_RE.sub("0", label)
+
+
 def calculate_scientific_notation_drawing_width(ui_settings: UISettings.UISettings, fonts: typing.Tuple[str, str], label: str) -> int:
     labels = label.lower().split("e")
     if len(labels) == 2:
         labels[0] = labels[0] + " x 10"
         labels[1] = str(int(labels[1]))
-    return sum(ui_settings.get_font_metrics(font, label).width for font, label in zip(fonts, labels))
+    return sum(ui_settings.get_font_metrics(font, _normalize_label_for_size_calculation(label)).width
+               for font, label in zip(fonts, labels))
 
 
 
--- a/nion/swift/test/LineGraphCanvasItem_test.py
+++ b/nion/swift/test/LineGraphCanvasItem_test.py
@@ -1,6 +1,7 @@
 # standard libraries
 import logging
 import math
+import operator
 import typing
 import unittest
 
@@ -12,13 +13,43 @@ from nion.data import Calibration
 from nion.data import DataAndMetadata
 from nion.swift import Application
 from nion.swift import LineGraphCanvasItem
+from nion.swift import LinePlotCanvasItem
 from nion.swift.model import DataItem
 from nion.swift.model import DisplayItem
 from nion.swift.model import Graphics
 from nion.swift.test import TestContext
+from nion.ui import CanvasItem
 from nion.ui import DrawingContext
 from nion.ui import TestUI
 
+
+def _enumerate_child_canvas_items_postorder(
+        canvas_item: CanvasItem.AbstractCanvasItem,
+        depth: int
+) -> typing.Iterator[typing.Tuple[CanvasItem.AbstractCanvasItem, int]]:
+    if isinstance(canvas_item, CanvasItem.CanvasItemComposition):
+        for child_canvas_item in canvas_item.canvas_items:
+            yield from _enumerate_child_canvas_items_postorder(child_canvas_item, depth+1)
+
+    yield canvas_item, depth
+
+
+def _find_first_descendent_of_type_postorder(
+        canvas_item: CanvasItem.AbstractCanvasItem,
+        canvas_item_type
+) -> typing.Optional[CanvasItem.AbstractCanvasItem]:
+    return next(
+        filter(
+            (lambda _: isinstance(_, canvas_item_type)),
+            map(operator.itemgetter(0), _enumerate_child_canvas_items_postorder(canvas_item, 0))
+        ),
+        None
+    )
+
+def _print_canvas_item_tree_preorder(canvas_item: CanvasItem.AbstractCanvasItem):
+
+    for ci, depth in _enumerate_child_canvas_items_postorder(canvas_item, 0):
+        print("0x{:016x} - {} {} - {}".format(id(ci), depth * '   ', type(ci).__name__, ci.canvas_bounds))
 
 def create_memory_profile_context() -> TestContext.MemoryProfileContext:
     return TestContext.MemoryProfileContext()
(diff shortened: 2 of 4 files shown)


cmeyer:
> How much of an expectation is there that the layout of a line plot does not vary between versions by a few pixels.
> -- This question would apply both for this fix, and for potential further fine-tuning in later releases.

When you say "between versions" - do you mean by platform versions or something else? If by platform, the answer is that ideally it wouldn't vary; but I think the default font used is different anyway so I would say, at least for now, pixel-by-pixel consistency between platforms is not an important requirement.

> Is there any reason to fine-tune the normalization so, say, it does not normalize the '1'in the exponential form ' x 10' .

I think our [me, @Brow71189] reasoning there was to match the behavior of matplotlib. However, I'm not sure if we matched it exactly; and I'm also not sure doing that is reasonable in all cases. I think either matching matplotlib or choosing something that is reasonably aesthetic are both good choices.

> Is there a reasonable way in tests to create a UISettings.UISettings instance so calculate_scientific_notation_drawing_width can be unit tested in isolation, or is the preferred way to test it via an actual line plot being rendered.

The source in nionswift is split into nion.swift namespace, which can reference UI; and nion.swift.model which cannot. However, there are some blurred lines - where does drawing code go when it draws something in the model but only references the DrawingContext to do drawing? In order to push that decision out to the future, I created UISettings, which is a bridge between the drawing code in the model that requires font sizing and the UI code.

In tests, I haven't been too strict about that separation between model and UI. The tests, more or less, all assume access to both UI and model at once.

Another consideration is that in the tests, there is no window system - so font metrics are usually calculated simply:
https://github.com/nion-software/nionui/blob/1e8efaf507c322aa6a896490e28db893029c6a59/nion/ui/TestUI.py#L1081

This might be a case where a hack is a good start; once I can look at it, I might be able to offer a better suggestion, if there is one.  You might be able to pass the TestUI directly as a UISettings, since get_font_metrics is available in either and works the same.

dhauge:
> > How much of an expectation is there that the layout of a line plot does not vary between versions by a few pixels.
> > -- This question would apply both for this fix, and for potential further fine-tuning in later releases.
> 
> When you say "between versions" - do you mean by platform versions or something else? If by platform, the answer is that ideally it wouldn't vary; but I think the default font used is different anyway so I would say, at least for now, pixel-by-pixel consistency between platforms is not an important requirement.
> 

I was talking about between versions of the code, on the same platform. Given that the cause of the bug is that the exact size depends on the less significant digits of the vertical bounds, if the axis label width is expected to stay the same across moderate changes in vertical bounds, and if the axis label width is to be deterministically calculated from the containing component size and not have any hysteresis, then the bug fix necessitates that some layouts will change with this fix.

> > Is there any reason to fine-tune the normalization so, say, it does not normalize the '1'in the exponential form ' x 10' .
> 
> I think our [me, @Brow71189] reasoning there was to match the behavior of matplotlib. However, I'm not sure if we matched it exactly; and I'm also not sure doing that is reasonable in all cases. I think either matching matplotlib or choosing something that is reasonably aesthetic are both good choices.
> 

I'll take a look at matplotlib. Does it avoid the problem by not making fine-tuned vertical axis dragging easy?

...

cmeyer:
> > > How much of an expectation is there that the layout of a line plot does not vary between versions by a few pixels.
> > > -- This question would apply both for this fix, and for potential further fine-tuning in later releases.
> > 
> > 
> > When you say "between versions" - do you mean by platform versions or something else? If by platform, the answer is that ideally it wouldn't vary; but I think the default font used is different anyway so I would say, at least for now, pixel-by-pixel consistency between platforms is not an important requirement.
> 
> I was talking about between versions of the code, on the same platform. Given that the cause of the bug is that the exact size depends on the less significant digits of the vertical bounds, if the axis label width is expected to stay the same across moderate changes in vertical bounds, and if the axis label width is to be deterministically calculated from the containing component size and not have any hysteresis, then the bug fix necessitates that some layouts will change with this fix.
> 

I'm not sure if I'm answering this completely, but I would say that as the line plot is resized either vertically or horizontally, or if the axes are dragged slowly, meaning they don't rapidly change magnitude of the displayed data, the width of the vertical axis labels should not jump around. It should appear smooth in most cases.

An approach I've used is to calculate the sizes based on data that is 50% beyond the display limits. I'm not sure how that works when rescaling, though. Maybe calculating the widths based on some guesses about what the user could potentially do (i.e. dragging up/down by 50%; rescaling in/out by 100%?)

An obvious but unaesthetic solution is to pad the label column to the maximum possible label width in all cases. But I'd like it to be more dynamic than that, so as to minimize padding in most cases and make it look good.

> > > Is there any reason to fine-tune the normalization so, say, it does not normalize the '1'in the exponential form ' x 10' .
> > 
> > 
> > I think our [me, @Brow71189] reasoning there was to match the behavior of matplotlib. However, I'm not sure if we matched it exactly; and I'm also not sure doing that is reasonable in all cases. I think either matching matplotlib or choosing something that is reasonably aesthetic are both good choices.
> 
> I'll take a look at matplotlib. Does it avoid the problem by not making fine-tuned vertical axis dragging easy?
> 
> ...

I don't know if matplotlib offers draggable axes at all. I was only referring to how it displayed exponents.

dhauge:
> > > > How much of an expectation is there that the layout of a line plot does not vary between versions by a few pixels.
> > > > -- This question would apply both for this fix, and for potential further fine-tuning in later releases.
> > > 
> > > 
> > > When you say "between versions" - do you mean by platform versions or something else? If by platform, the answer is that ideally it wouldn't vary; but I think the default font used is different anyway so I would say, at least for now, pixel-by-pixel consistency between platforms is not an important requirement.
> > 
> > 
> > I was talking about between versions of the code, on the same platform. Given that the cause of the bug is that the exact size depends on the less significant digits of the vertical bounds, if the axis label width is expected to stay the same across moderate changes in vertical bounds, and if the axis label width is to be deterministically calculated from the containing component size and not have any hysteresis, then the bug fix necessitates that some layouts will change with this fix.
> 
> I'm not sure if I'm answering this completely, but I would say that as the line plot is resized either vertically or horizontally, or if the axes are dragged slowly, meaning they don't rapidly change magnitude of the displayed data, the width of the vertical axis labels should not jump around. It should appear smooth in most cases.
> 
> An approach I've used is to calculate the sizes based on data that is 50% beyond the display limits. I'm not sure how that works when rescaling, though. Maybe calculating the widths based on some guesses about what the user could potentially do (i.e. dragging up/down by 50%; rescaling in/out by 100%?)
> 
> An obvious but unaesthetic solution is to pad the label column to the maximum possible label width in all cases. But I'd like it to be more dynamic than that, so as to minimize padding in most cases and make it look good.
> 

I'm not sure the padding can be minimized much more than the implementation in this PR does while still scrolling smoothly, at least without slower, more complex calculation, that looks at the exact labels that will be displayed. If scrolling a few pixels, say, changes the least significant digit from `'1'`to `'2'`, then the width will change a pixel. And at least in this test, the displayed label is at a tick that has `'0'`in the least significant digits, which effectively uses the max with at that digit. 

The calculation can be made more precise by slowing it down and adding complexity, by basically looking at what labels will actually be displayed (rather than using the label corresponding to the min/max calibrated value, which more than likely will not have a labelled tick associated with it). However, it would be a little tricky to do this precisely, because we would not be able to just check the min max. For example, if the labelled ticks were
- `'5310000``
- ...
- `'5220000'`
Then if just the max was used, `'5310000'`, then the calculated with will be less than that for the lower value `'5220000'`. This difference in width would probably not be noticed, and I'm not sure at this point whether a case can be constructed so the width difference would produce some portion being cut off.

Also, you say that you want the size to be calculated based on data 50% the display limit, but the current code in fact does 500% : (The `* 5` at the end)
```
            max_width = max(max_width, calculate_scientific_notation_drawing_width(ui_settings, self.__fonts, axes.y_ticker.value_label(axes.calibrated_value_max + y_range * 5)))
            max_width = max(max_width, calculate_scientific_notation_drawing_width(ui_settings, self.__fonts, axes.y_ticker.value_label(axes.calibrated_value_min - y_range * 5)))
```

cmeyer:
This PR looks good to me already. It would be nice to have a test to go along with it and then I'll merge.

dhauge:
I should have a test by tomorrow.

cmeyer:
For reference, see #399 (closed) option to display exponents in axes with superscript.

dhauge:
I added a test case that did not depend on any changes to other projects. I would like to move the text size calculation in `VarWidthSystemFontUserInterface` into the base `TestUI.UserInterface` implementation, and as you suggest in the email incorporate more realistic char widths calculated from a real font, but I think that should be done in a separate PR. If you think that change should be made as part of this issue, I can create a PR for the `nionswift.ui` branch, then create a new PR for this project that leverages the new `TestUI.UserInterface` implementation.

cmeyer:
Discussed offline: move get_font_metrics to TestUI as the default implementation and do a PR in `nionui`. Then push this PR again.

dhauge:
* Reverted changes from previous commit that added a local override to calculate var-width text metrics. All tests will fall back to the `UserInterface` defined in `TestUI` in the `nionui` project.
* This implementation was modified to calculate text widths in a way that approximates what would be returned for a var with font. This will be available in the next version of `nionui`
* Added a comment to update the `nionui` dependency to use this new implementation.

The tests will still succeed when the old `TestUI.UserInterface` implementation is used which calculates text widths as if the font was fixed width. These tests, however, will not effectively test the fix implemented for this issue until the new var-width font metrics are used.

PR #791, opened 2022-01-29
Implementation of get_mask function for point graphics and line graphics

mshakeelt:
added masking functionality for point graphics and line graphics.

### DIFF
--- a/nion/swift/model/Graphics.py
+++ b/nion/swift/model/Graphics.py
@@ -23,6 +23,8 @@ if typing.TYPE_CHECKING:
 
 DragPartData = typing.Tuple[typing.Any, ...]
 DragPartDataPlus = typing.Tuple[typing.Any, ...]
+NormPointType = typing.Tuple[float, float]
+NormSizeType = typing.Tuple[float, float]
 
 _ = gettext.gettext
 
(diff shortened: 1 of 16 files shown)


nion/swift/model/Graphics.py
```diff
@@ -1706,7 +1749,16 @@ def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapp
     def label_position(self, mapping: CoordinateMappingLike, font_metrics: UISettings.FontMetrics, padding: float) -> typing.Optional[Geometry.FloatPoint]:
         p = Geometry.FloatPoint.make(mapping.map_point_image_norm_to_widget(self.position))
         return p + Geometry.FloatPoint(-self.cross_hair_size - font_metrics.height * 0.5 - padding * 2, 0.0)
+    
+    @property
+    def used_role(self) -> typing.Optional[str]:
```
cmeyer, line 1754: The used role only needs to be overridden for graphics that can only be used for a specific role. In this case, point and line may not have a specific role; their role is only set to 'mask' if the user chooses to 'add to mask' on the graphic. So... this method should not be implemented on either the point or the rectangle.

nion/swift/model/Graphics.py
```diff
@@ -1706,7 +1749,16 @@ def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapp
...
+        return "fourier_mask"
 
+    def get_mask(self, data_shape: DataAndMetadata.ShapeType, calibrated_origin: typing.Optional[Geometry.FloatPoint] = None) -> DataAndMetadata._ImageDataType:
+        mask = numpy.zeros(data_shape)
+        x, y = self.position
+        mask[int(x*data_shape[0]), int(y*data_shape[1])] = 1
```
cmeyer, line 1760: There should be a check to ensure `x` and `y` are in the range [0,1]. If either are out of range, no value should be set.

nion/swift/model/Graphics.py
```diff
@@ -1501,6 +1501,49 @@ def label_position(self, mapping: CoordinateMappingLike, font_metrics: UISetting
         p2 = mapping.map_point_image_norm_to_widget(self.end)
         return Geometry.FloatPoint(y=(p1.y + p2.y) * 0.5, x=(p1.x + p2.x) * 0.5)
 
+    @property
+    def used_role(self) -> typing.Optional[str]:
+        return "fourier_mask"
+    
+    def draw_line(self, r0: int, c0: int, r1: int, c1: int) -> tuple(numpy.array, numpy.array, numpy.array):
```
cmeyer, line 1508: Can you move `draw_line` out of this class (i.e. don't pass `self`)? If the code were left as it were, the return type should be `typing.Tuple[numpy.typing.NDArray[typing.Any], numpy.typing.NDArray[typing.Any], numpy.typing.NDArray[typing.Any]]`. And it would be ideal to define `_ImageDataType = numpy.typing.NDArray[typing.Any]` at the top of the file and make the return type be `typing.Tuple[_ImageDataType, _ImageDataType, _ImageDataType]`.

However, I think this function should be simplified further so that it has the signature:

`def make_line_mask(start: Geometry.FloatPoint, end: Geometry.FloatPoint) -> _ImageDataType: ...`

Notice: it takes a start and end point and return a single mask image.

cmeyer (changes requested):
Additionally, I will want you to add a few test functions to ensure that the point and line are functioning properly (drawing into a 4x4 image, for example), and that both line and point handle out-of-bounds coordinates properly. I can help you write the initial tests after the next commit.

nion/swift/model/Graphics.py
mshakeelt: Got it. I have removed this method from line and point graphics.

nion/swift/model/Graphics.py
mshakeelt: I have implemented the check. points outside the range will not create the mask.

nion/swift/model/Graphics.py
mshakeelt: I have taken the method outside the class and simplified it as mentioned in the comment above!

PR #802, opened 2022-03-03
Reordered User Guide Outline

ejensen28:
All of the same document are still there but the order has been slightly modified. display_panels was moved above user_interface, and data_management was moved to the bottom.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -2,16 +2,16 @@
 
 .. _user-guide:
 
+##########
 User Guide
-==========
+##########
 
 .. toctree::
     introduction
-    user_interface
     display_panels
+    user_interface
     data
-    data_management
     graphics
     processing
+    data_management
     :maxdepth: 2
-    :caption: User Guide:


docs/user_guide.rst
```diff
@@ -2,16 +2,16 @@
 
 .. _user-guide:
 
+##########
 User Guide
-==========
+##########
```
cmeyer, line 7: Let's do header formatting in a different PR.

How to proceed: checkout this branch (or launch the appropriate codespace), unedit this change, commit the unedited change with `git commit -a --amend --no-edit`, then force push to the branch `git push --force` (you need to make sure you're on the right branch before force pushing! it is dangerous and can potentially overwrite your other PRs. I usually run `git push` and have it show me the error before `git push --force`, chat with me if in doubt).

cmeyer: Actually... just do a new commit and don't force push. Unedit the change, commit it, push it. I'll "squash" it when merging so the two commits will just be joined into one.

PR #806, opened 2022-03-11
New Overview Section

ejensen28:
This replaces the original Overview section with the new one.

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -12,11 +12,11 @@ Click here to get started fast with our :ref:`basic-use`.
 
 User Interface
 ==============
+The user interface is comprised of a central section called the workspace and the surrounding area for utility panels.
+
 .. image:: graphics/workspace.png
   :width: 320
-  :alt: Example of the Swift Workspace
-
-The user interface is comprised of a central section called the workspace and the surrounding area for utility panels. 
+  :alt: Example of the Swift Workspace 
 
 The central workspace is for display panels which can either display a data item or can display thumbnail images of multiple data items side by side in one of the browser formats. For more information, see :ref:`display-panels`.
 
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -18,8 +18,12 @@ The user interface is comprised of two main sections: a central section called t
   :width: 320
   :alt: Example of the Swift Workspace 
 
+Workspace
+---------
 The central workspace shows display panels which can either display a single data item or thumbnail images of multiple data items in one of the browser formats. For more information, see :ref:`display-panels`.
 
+Utility Panels
+--------------
 The utility panels surrounding the central workspace allow you to make adjustments to a selected data item in a display panel. For more information on their specific functions, see :ref:`user-interface`.
 
 Data and Display
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -14,8 +14,7 @@ User Interface
 ==============
 .. image:: graphics/workspace.png
   :width: 320
-  :alt: Example of the Swift workspace
-  :align: center
+  :alt: Example of the Swift Workspace
 
 The user interface is comprised of a central section called the workspace and the surrounding area for utility panels. 
 
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -4,7 +4,7 @@
 ********
 Overview
 ********
-Nion Swift is open source scientific image processing software for controlling instruments, and managing the acquisition, organization, visualization, and analysis of data. The software is designed to be used in conjunction with electron microscopes. However, it is also useful in other data visualization applications. Whether you already have your data, or you want to see the data as it is acquired, Nion Swift helps you visualize your data while it is being acquired and during processing and analysis.
+Nion Swift is open source scientific image processing software for controlling instruments, and managing the acquisition, organization, visualization, and analysis of data. The software is designed to be used in conjunction with electron microscopes. However, it is also useful in other data visualization applications. Nion Swift helps you visualize your data both while it is being acquired and during processing and analysis.
 
 Click to download Nion Swift or for :ref:`installation` instructions.
 
@@ -12,25 +12,25 @@ Click here to get started fast with our :ref:`basic-use`.
 
 User Interface
 ==============
-The user interface is comprised of a central section called the workspace and the surrounding area for utility panels.
+The user interface is comprised of two main sections: a central section called the workspace, and the surrounding area for utility panels.
 
 .. image:: graphics/workspace.png
   :width: 320
   :alt: Example of the Swift Workspace 
 
-The central workspace is for display panels which can either display a data item or can display thumbnail images of multiple data items side by side in one of the browser formats. For more information, see :ref:`display-panels`.
+The central workspace is for display panels which can either display a single data item or thumbnail images of multiple data items in one of the browser formats. For more information, see :ref:`display-panels`.
 
 The utility panels surrounding the central workspace allow you to make adjustments to a selected data item in a display panel. For more information on their specific functions, see :ref:`user-interface`.
 
 Data and Display
 ================
-Data is stored in a project as data items. Data items can be dragged into display panels to be displayed. Data items that are displayed in a display panel can be annotated with graphics, processed with computations, and more. These displayed data items are referred to as display items. Display items can be modified and viewed to see various aspects or values of the original data item. 
+Data is stored in a project as data items. Data items can be dragged into display panels to be displayed. Data items that are displayed in a display panel can be annotated with graphics, processed with computations, and more. These displayed data items are referred to as display items. 
 
 Read more in :ref:`data-items`.
 
 Graphics and Masks
 ==================
-Use graphics to highlight portions of the data, select areas for processing, or to focus on specific sections of data.
+Use graphics to highlight portions of the data, select areas for processing, or focus on specific sections of data.
 
 Masking allows you to block out part of an image and manipulate only a specific section of the data. Use a preset mask or add graphics to a mask to precisely define your area of interest.
 
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -4,7 +4,7 @@
 ********
 Overview
 ********
-Nion Swift is an image processing software for controlling instruments, and managing the acquisition, organization, visualization, and analysis of data. The software was designed to be used in conjunction with Nion's electron microscopes, but the functionality and the ability to extend its capabilities using Python mean that Nion Swift is useful for any number of data visualization projects. Whether you already have your data, or you want to see the data as it is acquired, Nion Swift helps you visualize your data in exactly the way you want.
+Nion Swift is open source scientific image processing software for controlling instruments, and managing the acquisition, organization, visualization, and analysis of data. The software is designed to be used in conjunction with electron microscopes. However, it is also useful in other data visualization applications. Whether you already have your data, or you want to see the data as it is acquired, Nion Swift helps you visualize your data while it is being acquired and during processing and analysis.
 
 Click to download Nion Swift or for :ref:`installation` instructions.
 
@@ -12,6 +12,11 @@ Click here to get started fast with our :ref:`basic-use`.
 
 User Interface
 ==============
+.. image:: graphics/workspace.png
+  :width: 320
+  :alt: Example of the Swift workspace
+  :align: center
+
 The user interface is comprised of a central section called the workspace and the surrounding area for utility panels. 
 
 The central workspace is for display panels which can either display a data item or can display thumbnail images of multiple data items side by side in one of the browser formats. For more information, see :ref:`display-panels`.
@@ -20,21 +25,21 @@ The utility panels surrounding the central workspace allow you to make adjustmen
 
 Data and Display
 ================
-Data is stored in a project as data items. Data items can be dragged into display panels to be displayed. Data items that are displayed in a display panel can be manipulated with graphics, computations, and more. These displayed data items are referred to as display items. Display items can be modified and viewed to see various aspects or values of the original data item. 
+Data is stored in a project as data items. Data items can be dragged into display panels to be displayed. Data items that are displayed in a display panel can be annotated with graphics, processed with computations, and more. These displayed data items are referred to as display items. Display items can be modified and viewed to see various aspects or values of the original data item. 
 
 Read more in :ref:`data-items`.
 
 Graphics and Masks
 ==================
-Nion Swift provides plenty of useful graphics to narrow down your view of the data. Use graphics to highlight portions of the data, select areas for processing, or to focus on specific sections of data.
+Use graphics to highlight portions of the data, select areas for processing, or to focus on specific sections of data.
 
-Masking allows you to block out the rest of an image and manipulate only the data you want. Choose from a variety of mask shapes to precisely define your area of interest.
+Masking allows you to block out part of an image and manipulate only a specific section of the data. Use a preset mask or add graphics to a mask to precisely define your area of interest.
 
 See :ref:`graphics` for more information.
 
 Processing and Analysis
 =======================
-Processing data by adding line profiles to an image, running Python scripts, filtering data, etc. can give you deeper insight into your data. Use the :ref:`inspector panel` to adjust various settings to tailor processes to your specific needs. Customizing processes allows for easier and more specific analysis of data.
+Processing data by adding line profiles to an image, running Python scripts, filtering data, etc. can give you deeper insight into your data. Use the :ref:`inspector panel` to adjust processing settings. Customizing processing allows for easier and more specific analysis of data.
 
 See :ref:`processing` for more information.
 
@@ -42,6 +47,6 @@ Projects and File Management
 ============================
 Imported and gathered data items are stored as a project in the form of an index file on your computer. Projects can be transferred just like any other file and can be opened in other installations of Nion Swift provided the versions are compatible.
 
-Import data into a project by dragging a data file into the workspace or by using the menu item [File > Import Data…]. Similarly, you can export data by using the menu item [File > Export…].
+Import data into a project by dragging a data file into the workspace or by using the menu item [File > Import Data…]. Export data by using the menu item [File > Export…].
 
 See :ref:`data-management` for more information.
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -12,46 +12,36 @@ Click here to get started fast with our :ref:`basic-use`.
 
 User Interface
 ==============
-The user interface is split up into a central section of the window called the workspace and the surrounding area for utility panels. 
+The user interface is comprised of a central section called the workspace and the surrounding area for utility panels. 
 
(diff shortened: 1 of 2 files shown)


ejensen28:
Some of the reference links are just linked to the broader sections because the new sections they are supposed to be referencing aren't implemented yet. These will be fixed in the future once those sections are ready.

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
-Key Features
+Click here to get started fast with our :ref:`basic-use`.
 
-- Data handling of 1D plots, 2D images, 1D and 2D collections of plots and images, and sequences.
-- Live computations that can be chained during acquisition or live parameter adjustment.
-- Open source, cross platform (macOS, Windows, Linux), and Python based scientific data processing.
-- Low latency, high throughput data acquisition platform.
+User Interface
+==============
+The user interface is split up into a central section of the window called the workspace and the surrounding area for utility panels.
```
cmeyer, line 15: This could be reworded. "The user interface window is comprised of a central section called the workspace and individual utility panels typically attached to the left/right edges of the window." I say 'typically' because they can be floating. But we can gloss over that point for now.

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
-- Open source, cross platform (macOS, Windows, Linux), and Python based scientific data processing.
-- Low latency, high throughput data acquisition platform.
+User Interface
+==============
+The user interface is split up into a central section of the window called the workspace and the surrounding area for utility panels. 
 
-Data Items and Display Items
-----------------------------
-|AppName| operates on data arrays, which include 2D images and 1D spectra, which are generally called *Data Items*.
+The central workspace is for display panels which can either display a single piece of data (also called a data item) or can display multiple pieces of data side by side in one of the browser formats. For more information, see :ref:`display-panels`.
```
cmeyer, line 17: What do you think of just "display data" instead of "display a single piece of data (aka a data item)".

Also, some clarification: the browsers are literally for browsing and selecting the data to be displayed in a panel, not for displaying them side-by-side.

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
 
-Data items represent data arrays with anywhere from one to five dimensions. The data array associated with a data item can be 1D or 2D data and then organized into 1D or 2D collections or 1D sequences.
+The utility panels surrounding the central workspace allow you to make adjustments to a selected data item in a display panel. For more information on their specific functions, see :ref:`user-interface`.
 
-A spectrum image would be a 2D collection of 1D data. This would be typical data acquired from a STEM microscope spectrum imaging acquisition.
+Data and Display
+================
+A piece of data stored in the program is called a data item. Data items can be dragged into display panels to be displayed. Once displayed, they become display items. Display items can be modified and viewed to see various aspects or values of the original data item. Read more in :ref:`data-items`.
```
cmeyer, line 23: Again, I'd avoid "piece of data" since it is just a renaming of "item of data" and just use something simpler like "Data stored in the application is called..." or "Data items stored in ...".

Also, I'd like to stick with "application" rather than "program"; but ideally avoid those altogether.

"they become display items" -- well, not really: they _are_ display items. "become" implies some sort of action.

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
-Data Items are displayed using one or more *Display Items*.
+If a data item includes complex components, it will need to be reduced into a scalar image or line plot in order to be displayed. For more info, see :ref:`data-items`.
 
-See :ref:`data-items`.
+Sometimes, data items can be part of a larger 'stack' of data which will require higher level organization that the data panel can provide. See :ref:`data-items` for more info.
 
-Projects
---------
-Data items are organized into projects. Projects are stored on disk and take the form of an index file and an associated data directory.
+The INFO PANEL shows you the location and value of the data in a display panel under your cursor. You can choose how you would like the :ref:`data-items` (coordinate system) to be defined in the inspector panel.
```
cmeyer, line 29: Did you forget a link to the Info Panel here? (INFO PANEL)

Overall, this section seems to be mixing technical details (i.e. converting complex to scalar, displaying values in the info panel, etc.) with an overview (describing what the overall purpose is).

docs/introduction.rst
ejensen28: Right, I should've specified that the browsers display thumbnail images of data items side by side. I'll go ahead and reword this. Should I add that as a new commit to this pull request?

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
 
-You can have any number of projects, but only a single one can be open at once. When you open a project, it is associated with a main document window.
+Click on any data item, display panel, or graphic to select it. You can select more than one item of the same kind using [ctrl + click] on Windows or [cmd + click] on macOS.
```
cmeyer, line 31: Does this belong in the overview?

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
-Data items represent data arrays with anywhere from one to five dimensions. The data array associated with a data item can be 1D or 2D data and then organized into 1D or 2D collections or 1D sequences.
+The utility panels surrounding the central workspace allow you to make adjustments to a selected data item in a display panel. For more information on their specific functions, see :ref:`user-interface`.
 
-A spectrum image would be a 2D collection of 1D data. This would be typical data acquired from a STEM microscope spectrum imaging acquisition.
+Data and Display
+================
+A piece of data stored in the program is called a data item. Data items can be dragged into display panels to be displayed. Once displayed, they become display items. Display items can be modified and viewed to see various aspects or values of the original data item. Read more in :ref:`data-items`.
 
-Data Items are displayed using one or more *Display Items*.
+If a data item includes complex components, it will need to be reduced into a scalar image or line plot in order to be displayed. For more info, see :ref:`data-items`.
```
cmeyer, line 25: Too technical for an overview.

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
 
-See :ref:`data-items`.
+Sometimes, data items can be part of a larger 'stack' of data which will require higher level organization that the data panel can provide. See :ref:`data-items` for more info.
```
cmeyer, line 27: This sentence either needs to be removed or clarified. What's the point of this? As it reads now, it's describing something that can't be done; but I have a feeling you had another point in mind.

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
-----------
-The main window is organized around a main area called a workspace. A workspace has a layout which includes one or more display panels in which display items can be visualized.
+Selecting multiple data items will allow you to move them at the same time. However, multiple data items cannot be dragged into the same display panel.
 
-You can create additional workspace for different uses and easily switch between them.
+Selecting multiple graphics in the same display panel will allow you to move them at the same time. The inspector panel will display options for all of the selected graphics.
 
-In addition, the main window has a variety of utility panels attached or associated with the main window. Some examples of utility panels are the data browser panel, the inspector panel, and the info and histogram panels.
+Graphics and Masks
+==================
```
cmeyer, line 40: As a counter example, this section is a good "overview".

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
+Nion Swift provides plenty of useful graphics to narrow down your view of the data. Use graphics to highlight portions of the data, select areas for processing, or to focus on specific sections of data.
 
-See :ref:`user-interface`.
+Masking allows you to block out the rest of an image and see only what you want. Choose from a variety of mask shapes to precisely define your area of interest.
```
cmeyer, line 43: ... FYI not only "see" but use in processing. i.e. you might want to only set the masked portion of an image to 0's. You can probably come up with a better wording of my explanation :)

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
 
-Display Panels
---------------
-Workspaces are split into one or more display panels. Each display panel can show one display item.
+See :ref:`graphics` for more information.
 
-A typical display item would show a single data item, but you can create multiple display items for each data item to visualize the data differently at the same time.
+Processing and Analysis
+=======================
+Processing and analysis greatly expand the functionality of Nion Swift. The software provides many options for filtering, transforming, and more. Computations allow for specific and customizable processes, and there is even support for python scripting to add your own computations.
```
cmeyer, line 49: The first sentence seems "sales"-ey. I noticed you removed the sales-like stuff at the top of this page; so you should remove it here too. Words/phrases like "greatly" and ".. there is even support for .." are probably out of scope for a reference guide.

Also, one minor but important note: Python should be capitalized in all edits.

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
-Graphics
---------
-Graphics are annotations or other graphical items that can be attached to display items.
-
-Examples of graphics would be a point, line, or rectangle on a 2D image or a interval or channel on a 1D plot.
-
-Additionally, Fourier mask filter graphics, such as band pass filters or lattice filters, can be attached to data in the Fourier domain.
-
-See :ref:`graphics`.
+See :ref:`data-management` for more information.
```
cmeyer, line 57: One additional note: the old overview included a section on "projects". I think we should still include that. Was there a reason you dropped it from this section? The reason I think it's important is to introduce users to the term "project" a.s.a.p. I'm open to other suggestions, but that's my thought.

docs/introduction.rst
```diff
@@ -1,73 +1,57 @@
...
+=======================
+Processing and analysis greatly expand the functionality of Nion Swift. The software provides many options for filtering, transforming, and more. Computations allow for specific and customizable processes, and there is even support for python scripting to add your own computations.
 
-In addition, some display items may display multiple data items at once: a line plot may have multiple layers, each displaying a different data item or even different reductions of the same data item.
+See :ref:`processing` for more information.
 
-See :ref:`display-panels`.
+File Management
+===============
+Make use of Nion Swift without live acquisition by importing your own data files into the program, take your files with you by exporting your data, save your project files and use them on any other installation of Nion Swift, make copies of your data to create a backup, and contact us about recovery options.
```
cmeyer, line 55: The sentence reads like bullet points.

"program" -> "application".

One idea to avoid using "program/application" is to say something like "importing your own data files into a project".

docs/introduction.rst
```diff
@@ -1,73 +1,47 @@
 .. include:: defs.rst
 .. _introduction:
 
+********
 Overview
-========
-|AppName| is open source scientific image processing software integrating hardware control, data acquisition, visualization, processing, and analysis using Python. |AppName| is easily extended using Python. It runs on Windows, Linux, and macOS.
+********
+Nion Swift is an image processing software for controlling instruments, and managing the acquisition, organization, visualization, and analysis of data. The software was designed to be used in conjunction with Nion's electron microscopes, but the functionality and the ability to extend its capabilities using Python mean that Nion Swift is useful for any number of data visualization projects. Whether you already have your data, or you want to see the data as it is acquired, Nion Swift helps you visualize your data in exactly the way you want.
```
cmeyer, line 7: Let's refer to "software" as a plural object, and I still like the initial phrase since it works well in search engines. "Nion Swift is open source scientific image processing software for ..."

Also, I'd back off the "Nion electron microscope" and "extend with Python" only because those are (1) too specific; (2) difficult to explain quickly -- "The software is designed to be used in conjunction with electron microscopes. However, it is also useful in other data visualization applications." (n.b. I'd prefer 'applications' rather than 'projects' to avoid confusion with 'Projects' defined below).

Finally, the last sentence (and several others in this document) still read too sales-y. I would stick more to matter-of-fact sentence structure: "Nion Swift helps you visualize and process your data while it is being acquired and during processing and analysis."

docs/introduction.rst
```diff
@@ -1,73 +1,47 @@
...
-Key Features
+Click here to get started fast with our :ref:`basic-use`.
 
-- Data handling of 1D plots, 2D images, 1D and 2D collections of plots and images, and sequences.
-- Live computations that can be chained during acquisition or live parameter adjustment.
-- Open source, cross platform (macOS, Windows, Linux), and Python based scientific data processing.
-- Low latency, high throughput data acquisition platform.
+User Interface
+==============
+The user interface is comprised of a central section called the workspace and the surrounding area for utility panels.
```
cmeyer, line 15: We might consider adding a graphic here, but very small... something perhaps 320x240 that gives an overall screenshot. You could reduce the one on the main page for this.

docs/introduction.rst
```diff
@@ -1,73 +1,47 @@
...
+The central workspace is for display panels which can either display a data item or can display thumbnail images of multiple data items side by side in one of the browser formats. For more information, see :ref:`display-panels`.
 
-Data items represent data arrays with anywhere from one to five dimensions. The data array associated with a data item can be 1D or 2D data and then organized into 1D or 2D collections or 1D sequences.
+The utility panels surrounding the central workspace allow you to make adjustments to a selected data item in a display panel. For more information on their specific functions, see :ref:`user-interface`.
 
-A spectrum image would be a 2D collection of 1D data. This would be typical data acquired from a STEM microscope spectrum imaging acquisition.
+Data and Display
+================
+Data is stored in a project as data items. Data items can be dragged into display panels to be displayed. Data items that are displayed in a display panel can be manipulated with graphics, computations, and more. These displayed data items are referred to as display items. Display items can be modified and viewed to see various aspects or values of the original data item.
```
cmeyer, line 23: "Data items that are display in a display panel can be annotated with graphics, processed with computations, and more."

docs/introduction.rst
```diff
@@ -1,73 +1,47 @@
...
+================
+Data is stored in a project as data items. Data items can be dragged into display panels to be displayed. Data items that are displayed in a display panel can be manipulated with graphics, computations, and more. These displayed data items are referred to as display items. Display items can be modified and viewed to see various aspects or values of the original data item. 
 
-Data Items are displayed using one or more *Display Items*.
+Read more in :ref:`data-items`.
 
-See :ref:`data-items`.
+Graphics and Masks
+==================
+Nion Swift provides plenty of useful graphics to narrow down your view of the data. Use graphics to highlight portions of the data, select areas for processing, or to focus on specific sections of data.
```
cmeyer, line 29: "Nion Swift provides graphics to annotation and select parts of your data." (n.b. "plenty of useful" doesn't add much and every prepositional clause makes it more difficult for non-native-English speakers).

You could also combine both sentences here since they say the same thing. I like the second sentence.

docs/introduction.rst
```diff
@@ -1,73 +1,47 @@
...
 
-Projects
---------
-Data items are organized into projects. Projects are stored on disk and take the form of an index file and an associated data directory.
+Masking allows you to block out the rest of an image and manipulate only the data you want. Choose from a variety of mask shapes to precisely define your area of interest.
```
cmeyer, line 31: "block out the rest of" -> "block out part of".

"You can choose..." ?

docs/introduction.rst
```diff
@@ -1,73 +1,47 @@
...
 
-You can have any number of projects, but only a single one can be open at once. When you open a project, it is associated with a main document window.
+See :ref:`graphics` for more information.
 
-See :ref:`data-management`.
+Processing and Analysis
+=======================
+Processing data by adding line profiles to an image, running Python scripts, filtering data, etc. can give you deeper insight into your data. Use the :ref:`inspector panel` to adjust various settings to tailor processes to your specific needs. Customizing processes allows for easier and more specific analysis of data.
```
cmeyer, line 37: "tailor processes to" -> "adjust processing for"

"Customizing processes..." -> "Customizing processing..."

Use the term "processing"

docs/introduction.rst
```diff
@@ -1,73 +1,47 @@
...
 
-Workspaces
-----------
-The main window is organized around a main area called a workspace. A workspace has a layout which includes one or more display panels in which display items can be visualized.
+See :ref:`processing` for more information.
 
-You can create additional workspace for different uses and easily switch between them.
+Projects and File Management
+============================
+Imported and gathered data items are stored as a project in the form of an index file on your computer. Projects can be transferred just like any other file and can be opened in other installations of Nion Swift provided the versions are compatible.
```
cmeyer, line 43: This needs a bit of rewording. My suggestion is to start with the key point:

"Data is organized into projects. A project file..."

Also, what do you think about putting this earlier in the document? I'm not sure where it fits; but it's something to think about.

cmeyer (changes requested):
Looking pretty good - I have more comments, though. Let's do another pass.

docs/introduction.rst
cmeyer: Reiterating this comment.

The old text said "Data items are organized into projects. Projects are stored on disk and take the form of an index file and an associated data directory."

I would keep it simple but reword something like this:

"Nion Swift organizes data items into projects. Projects are folders/directories on your computer which contain both an index file for tracking relationships between data items and the individual files for each data item. Projects can be archived and/or moved to other computers and opened in other installations of Nion Swift."

Overall, this may take a few passes to get right, but I think this is a good starting paragraph.

Also, I don't want to lose this key piece of information:

"You can have any number of projects, but only a single one can be open at once. When you open a project, it is associated with a main document window."

docs/introduction.rst
```diff
@@ -1,73 +1,51 @@
...
 
-Data Items and Display Items
-----------------------------
-|AppName| operates on data arrays, which include 2D images and 1D spectra, which are generally called *Data Items*.
+.. image:: graphics/workspace.png
+  :width: 320
+  :alt: Example of the Swift Workspace 
 
-Data items represent data arrays with anywhere from one to five dimensions. The data array associated with a data item can be 1D or 2D data and then organized into 1D or 2D collections or 1D sequences.
+The central workspace is for display panels which can either display a single data item or thumbnail images of multiple data items in one of the browser formats. For more information, see :ref:`display-panels`.
```
cmeyer, line 21: This could be reworded to use active voice: "The central workspace shows display panels..." or something like that.

docs/introduction.rst
```diff
@@ -1,73 +1,51 @@
...
 
-See :ref:`data-management`.
+See :ref:`graphics` for more information.
 
-Workspaces
-----------
-The main window is organized around a main area called a workspace. A workspace has a layout which includes one or more display panels in which display items can be visualized.
+Processing and Analysis
+=======================
+Processing data by adding line profiles to an image, running Python scripts, filtering data, etc. can give you deeper insight into your data. Use the :ref:`inspector panel` to adjust processing settings. Customizing processing allows for easier and more specific analysis of data.
```
cmeyer, line 41: Similar to other comments I've made, I think this is too "sales-y". I'd rather this just be an explanation followed by a few examples. I'm not convinced my explanation is notably better, but I think it provides a better starting point:

"Nion Swift can automatically perform processing and analysis that is performed when data or parameters change as the result of acquisition, user-initiated parameter changes, or regions of interest edits."

"Use the inspector panel..."

I would leave out the mention of the line profile since it is very specific and not easy to explain in this section.

You could include an example to the effect of "you can do processing on an image by making a rectangle to serve as the region of interest, then choosing Processing > Filter > Sobel..." but that is probably too much detail for this section.

cmeyer (changes requested):

cmeyer:
Merged by hand in 40134c6d50594e6744424b5f17bc13aa3f861cd7

Issue #842, opened 2022-05-24
Dragging Nx1 data items to line plot does not allow them to be dropped


cmeyer:
Nx1 or 1xN data items are displayed as line plots; but dragging them to another line plot and/or dragging other line plots to them fails.

This is probably due to the `get_drop_regions_map` method in `LinePlotCanvasItem.py` checking whether `is_datum_1d` but there may be more to it. Instead `get_drop_regions_map` should use check whether `display.display_data_channel` exists and whether `display.display_data_channel.is_display_1d_preferred` returns True. This is what `DisplayItem.used_display_type` uses to determine the default display type.

PR #843, opened 2022-05-25
Preliminary fix for #842

gosselind1:
Modified if statement to:
1. Check if both graphs are of the line type
2. The utilized units of both graphs are identical


cmeyer:
Merged with minor edits to commit message and to remove whitespace.

Issue #657, opened 2021-03-24
Line plot colors should default to solid strokes and transparent (50%) fills (use existing colors)


cmeyer:
Problem we're trying to solve is shown in this figure. If the red plot is entirely behind the blue plot, there is no way to see the red plot. So the idea is to draw the fills first, then draw the strokes. We may also have to modify the defaults so that a default line plot always draws the stroke; currently the stroke default is to not draw.

<img width="465" alt="image" src="https://user-images.githubusercontent.com/34664/170125750-897ae4d9-6555-4014-87fc-90ee7e3fa14f.png">

ND 2021-03-24:

There seems to be a lot of bug sorting going on, this one definitely
should get in there.

Line profiles default to filled, this is super annoying when combining
plots, every single plot has to have filled turned off and outline
turned on one at a time.

One can argue about what is the right solution, but the minimal solution
is to:

A-default style for all plots include the line, whether it should
include the fill is discussable but it should always include the line.

B-all lines should be plotted on top of all fills, this way no plot is
ever hidden.

cmeyer:
BPS 2021-06-04: When dragging line plots on top of each other, the default display settings are red and blue fill, with no stroke color. This is pretty much always a poor choice, since overlapping data gets hidden. I'd propose to make the default red and blue stroke colors, with no fill color.

cmeyer:
Fills/strokes drawn for each layer, with transparency:

<img width="461" alt="image" src="https://user-images.githubusercontent.com/34664/170789321-2faea481-5700-4ea1-85c5-9027d40de73d.png">

Fills drawn for each layer then strokes drawn for each layer, no transparency:

<img width="458" alt="image" src="https://user-images.githubusercontent.com/34664/170789510-68ad609e-dfd6-4be6-81d2-bd124ec3ed66.png">

PR #844, opened 2022-05-27
Issue#657

gosselind1:
Simple fix of #657 by utilizing  double drawing and transparency


gosselind1:
Implementing alternative implementation with transparency

PR #845, opened 2022-05-30
Alternative Fix to Issue#657

gosselind1:
After discussion and updated requirements, this merge request modifies the prepare_display function to inject transparency into layers without defined transparency, with the exception of the last layer.

### DIFF
--- a/nion/swift/model/DisplayItem.py
+++ b/nion/swift/model/DisplayItem.py
@@ -1997,7 +1997,7 @@ class DisplayItem(Persistence.PersistentObject):
 
     def __get_unique_display_layer_color(self) -> str:
         existing_colors = {display_layer.fill_color for display_layer in self.display_layers}
-        for color in ('#1E90FF', "#F00", "#0F0", "#00F", "#FF0", "#0FF", "#F0F", "#888", "#800", "#080", "#008", "#CCC", "#880", "#088", "#808", "#964B00"):
+        for color in ('#1E90FF', "#FF0000", "#00FF00", "#0000FF", "#FFFF00", "#00FFFF", "#FF00FF", "#888888", "#880000", "#008800", "#000088", "#CCCCCC", "#888800", "#008888", "#880088", "#964B00"):
             if not color in existing_colors:
                 return color
         return '#1E90FF'
--- a/nion/swift/model/DisplayItem.py
+++ b/nion/swift/model/DisplayItem.py
@@ -1313,8 +1313,8 @@ def display_layer_factory(lookup_id: typing.Callable[[str], str]) -> DisplayLaye
 
 
 class DisplayItem(Persistence.PersistentObject):
-    DEFAULT_COLORS = ('#1E90FF', "#FF0000", "#00FF00", "#0000FF", "#FFFF00", "#00FFFF", "#FF00FF", "#888888", "#880000",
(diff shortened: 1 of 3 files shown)


gosselind1:
New push removes transparent layers, and instead gives outlines by default on any layer past the first one

nion/swift/model/DisplayItem.py
```diff
@@ -1996,11 +2002,13 @@ def append_display_data_channel(self, display_data_channel: DisplayDataChannel,
             self.__add_display_layer_auto(display_layer, display_data_channel)
 
     def __get_unique_display_layer_color(self) -> str:
+        # get a set of currently used colors w/o alpha, and then check if they're in a static color list
         existing_colors = {display_layer.fill_color for display_layer in self.display_layers}
-        for color in ('#1E90FF', "#F00", "#0F0", "#00F", "#FF0", "#0FF", "#F0F", "#888", "#800", "#080", "#008", "#CCC", "#880", "#088", "#808", "#964B00"):
-            if not color in existing_colors:
+        existing_colors = set(map(lambda c: "#" + c[3:] if isinstance(c, str) and len(c) == 9 else c, existing_colors))
```
cmeyer, line 2007: minor change here: can you using `DrawingContext.color_without_alpha` to convert the existing colors to hex values and then compare the hex values in all cases? That way it will handle strings like "red" and "rgb(255,0,0)" as the same. (on second look, I think you'll need to add a similar function in `DrawingContext` - something like `canonical_color_without_alpha` that always returns hex colors of length 9. That will be a separate PR.

nion/swift/model/DisplayItem.py
```diff
@@ -1310,6 +1310,12 @@ def display_layer_factory(lookup_id: typing.Callable[[str], str]) -> DisplayLaye
 
 
 class DisplayItem(Persistence.PersistentObject):
+    DEFAULT_COLORS = ('#1E90FF', "#FF0000", "#00FF00", "#0000FF", "#FFFF00", "#00FFFF", "#FF00FF", "#888888", "#880000",
+                      "#008800", "#000088", "#CCCCCC", "#888800", "#008888", "#880088", "#964B00")
+                      # While one might this as a set, making this a set destroys the order, and would require that
```
cmeyer, line 1315: Can you simplify this to something like "this list should is ordered" or just remove the comment?

cmeyer (changes requested):

nion/swift/model/DisplayItem.py
cmeyer: One additional change that would be nice - you can use the named colors instead of the hex colors in the DEFAULT_COLORS list. The last color is the list can use "sienna" as the named color.

cmeyer: Also, the first color is "dodgerblue"

cmeyer:
As per previous discussion, I decided to introduce a simplified `Color` class myself with a method `matches_without_alpha`. Can you revisit this PR and rewrite in terms of the new `Color` class?

Usage would be something like this to determine if a color was already being used:

`if Color.Color(proposed_color_str).matches_without_alpha(Color.Color(existing_color_str))`

Also, this will require an updated dependency, so please include a change to `setup.py` like this:
https://github.com/nion-software/nionui/commit/105cdf44961f99bc8e6561546050117c73e63320#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52

PR #857, opened 2022-06-20
Fix #390

### LINKED ISSUE #390: Line plot should allow resizing intervals from the center
The alt key should resize from center of interval (`modifiers.alt`). This matches resizing rectangles on images.

Note: this may be a slightly more complex project since it involves tracking the anchor point. It will need test cases and need to handle the user changing the modifier in the middle of a drag.

gosselind1:
Modified IntervalGraphics.adjust_part to resize both fences when pressing the alt key.
Added a test to verify nominal operation of interval resizing from the center.


cmeyer:
Merged 6260b1e255ef5892da3437ecd865c9643bd81b6e (with commit message change).

Issue #495, opened 2020-07-30
Add option to put legend outside of line plot rectangle


cmeyer:
![image](https://user-images.githubusercontent.com/34664/88870558-86991600-d1ca-11ea-8918-ec4b6f4f1154.png)

cmeyer:
To accomplish this, a different canvas item layout can be used. Here is some starter code to demonstrate the idea:

```
diff --git a/nion/swift/LinePlotCanvasItem.py b/nion/swift/LinePlotCanvasItem.py
index 22a83cb1..9dfd9f9c 100755
--- a/nion/swift/LinePlotCanvasItem.py
+++ b/nion/swift/LinePlotCanvasItem.py
@@ -165,7 +165,15 @@ class LinePlotCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
         line_graph_background_canvas_item = CanvasItem.CanvasItemComposition()
         #line_graph_background_canvas_item.update_sizing(line_graph_background_canvas_item.size.with_minimum_aspect_ratio(1.5))  # note: no maximum aspect ratio; line plot looks nice wider.
         line_graph_background_canvas_item.add_canvas_item(CanvasItem.BackgroundCanvasItem("#FFF"))
-        line_graph_background_canvas_item.add_canvas_item(line_graph_group_canvas_item)
+
+        legend_row = CanvasItem.CanvasItemComposition()
+        legend_row.layout = CanvasItem.CanvasItemRowLayout()
+        legend_row.add_canvas_item(line_graph_group_canvas_item)
+        self.__line_graph_legend_canvas_item2 = LineGraphCanvasItem.LineGraphLegendCanvasItem(ui_settings, typing.cast(LineGraphCanvasItem.LineGraphLegendCanvasItemDelegate, delegate))
+        legend_row.add_canvas_item(self.__line_graph_legend_canvas_item2)
+
+        line_graph_background_canvas_item.add_canvas_item(legend_row)
+
         line_graph_background_canvas_item.add_canvas_item(self.__overlap_controls)
 
         self.__display_controls = CanvasItem.CanvasItemComposition()
@@ -258,6 +266,9 @@ class LinePlotCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
         else:
             self.__line_graph_legend_canvas_item.visible = False
 
+        self.__line_graph_legend_canvas_item2.size_to_content()
+        self.__line_graph_legend_canvas_item2.visible = True
+
[...]

PR #859, opened 2022-06-22
Added the Ability to Have Legend Outside of LinePlot

gosselind1:
This is the first of a few PRs to address #495. Specifically, in this set of commits the ability for the LinePlot to have its legend drawn outside of the graph area has been added, but the functionality is not yet exposed to the UI.

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -1981,9 +1981,10 @@ class ImageDisplayInspectorSection(InspectorSection):
 
 
 def make_legend_position(document_controller: DocumentController.DocumentController, display_item: DisplayItem.DisplayItem) -> typing.Tuple[UserInterface.BoxWidget, Event.EventListener]:
+    # This function is currently only utilized by lineplots
     ui = document_controller.ui
     legend_position_row = ui.create_row_widget()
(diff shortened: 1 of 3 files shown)


cmeyer:
This looks good so far. A couple of requests - can you add a test that sets the legend position to these two values and forces it to render? If I start with a multi-line plot and assign it a script reference using Ctrl+Shift+K and then open a script window with Ctrl+K and then run the script as below ("r01" will be specific to what is assigned to the display), it triggers an exception in the inspector. I'd like to see a test that triggers that error also and then also include the trivial fix for that error in this PR. The fix is probably one line, using a default to look up the menu item in the inspector.

`r01._display_item.set_display_property('legend_position', 'outer-right')`

I didn't find a really good similar test, but you could copy/distill `test_move_display_layer_from_display_to_display_undo_redo` down to a few lines and it probably comes the closest. A good strategy is: reproduce the error with the instructions above; write the test to expose the same error; then fix the problem (in `Inspector.py`) and then ensure the test passes with the fix.

cmeyer:
Closed with rebase and some minor modifications. a6da0bd7f8bee5c4350f6b07063202a46ea972aa.

PR #860, opened 2022-06-27
Fix #513 Modfied LinePlot Interval to Bind to Selected Area

### LINKED ISSUE #513: View to intervals is inconsistent when intervals straddle left-most or right-most channels

gosselind1:
Modified LinePlot intervals to show selected are even when area is outside of graph's bounds
Fixes #513


cmeyer:
Merged with minor commit message edit 697f10fbf4cdc81e26db1511655d9f0470ac20d0

PR #863, opened 2022-06-30
Issue#861

gosselind1:
Fixed bugs introduced in 697f10f
Included some tests to validate operation of __view_to_intervals

This PR builds off of the original branch for #513, so some modification might be required to merge.


cmeyer:
Merged with modifications: 7d49523f5ae4698716c0fea9ab115db88c8eb131

The modifications are:

- ensure "out of bounds" interval is entirely out of bounds (2.0 -> 2.5) instead of almost out of bounds (1.0 -> 1.5).
- do not use random values as this potentially leads to occasional test false negatives.
- do not access private line plot variables; instead use the `_axes` to figure out left/right drawn channels.
- squashed commits and shorter, concise commit message.

PR #865, opened 2022-07-01
Additional Changes to LinePlot Interval Testing

gosselind1:
Fixed a minor bug with the interval code, though it seems to not really be a bug due to it not failing any tests despite being the "wrong" code.
Additionally consolidated and expanded interval zooming tests.


cmeyer:
Merged in commits 6e831a425909859a2208b87b46c9eaae31170f1a and b8000b0af7067c40ceb357608d6add185cfad455 with minor edits:

- Fixed first line of commits to have proper capitalization.
- Eliminated use of subtest - these tests are fast enough and don't warrant a subtest.

Additional note: I disfavor subtests because the testing apparatus is not able to predict how many tests run with them and gets its progress bar wrong. Generally a minor thing, but I run the tests a lot, so this is important to me.

PR #867, opened 2022-07-07
Fix #657 Utility Class Approach

### LINKED ISSUE #657: Line plot colors should default to solid strokes and transparent (50%) fills (use existing colors)

gosselind1:
Implementation for #657 utilizing the new color class.

Right now it is utilizing the svg color set within the utility class as a proper default color list did not seem to make the cut.
In addition, the logic __get_unique_display_color is a bit more complicated than I would like due to being unable to get a converted color / color string directly

### DIFF
--- a/nion/swift/LinePlotCanvasItem.py
+++ b/nion/swift/LinePlotCanvasItem.py
@@ -24,7 +24,6 @@ from nion.swift.model import Utility
 from nion.ui import CanvasItem
 from nion.utils import Geometry
 from nion.utils import Registry
-from nion.utils import Color
 
 if typing.TYPE_CHECKING:
     from nion.swift.model import Persistence
@@ -512,7 +511,7 @@ class LinePlotCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
             if self.__display_frame_rate_id:
                 Utility.fps_tick("prepare_"+self.__display_frame_rate_id)
 
-            colors = list({*Color.svg_color_map.values()})
+            colors = list(DisplayItem.DisplayItem.DEFAULT_COLORS)
 
             max_layer_count = len(self.__line_graph_stack.canvas_items)
 
--- a/nion/swift/LinePlotCanvasItem.py
+++ b/nion/swift/LinePlotCanvasItem.py
@@ -512,7 +512,7 @@ class LinePlotCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
             if self.__display_frame_rate_id:
                 Utility.fps_tick("prepare_"+self.__display_frame_rate_id)
 
-            colors = Color.svg_color_map.values()
+            colors = typing.cast(typing.List[str], [*{Color.svg_color_map.values()}])
 
             max_layer_count = len(self.__line_graph_stack.canvas_items)
 
--- a/nion/swift/LinePlotCanvasItem.py
+++ b/nion/swift/LinePlotCanvasItem.py
@@ -512,7 +512,7 @@ class LinePlotCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
             if self.__display_frame_rate_id:
                 Utility.fps_tick("prepare_"+self.__display_frame_rate_id)
 
-            colors = typing.cast(typing.List[str], [*{Color.svg_color_map.values()}])
+            colors = list({*Color.svg_color_map.values()})
 
             max_layer_count = len(self.__line_graph_stack.canvas_items)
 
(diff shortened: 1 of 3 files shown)


gosselind1:
That windows 3.9 build is failing on one of the tests that have been exclusively failing on my machine
(reproduced by testing against an up-to-date main on my problemed machine)

gosselind1:
I also seem to have accidentally added an unintended 'v' character in the setup.cfg file. Holding off on removing it so that failing test remains up.

nion/swift/LinePlotCanvasItem.py
```diff
@@ -511,7 +512,7 @@ def calculate_scalar_xdata(xdata_list: typing.Sequence[typing.Optional[DataAndMe
             if self.__display_frame_rate_id:
                 Utility.fps_tick("prepare_"+self.__display_frame_rate_id)
 
-            colors = ('#1E90FF', "#F00", "#0F0", "#00F", "#FF0", "#0FF", "#F0F", "#888", "#800", "#080", "#008", "#CCC", "#880", "#088", "#808", "#964B00")
+            colors = list({*Color.svg_color_map.values()})
```
cmeyer, line 515: I'd like to keep the existing color list rather than choosing the first colors listed in the svg_color_map. But we should move the list to `DisplayItem.py` so that it is only defined in one place.

nion/swift/model/DisplayItem.py
```diff
@@ -1998,11 +1999,17 @@ def append_display_data_channel(self, display_data_channel: DisplayDataChannel,
             self.__add_display_layer_auto(display_layer, display_data_channel)
 
     def __get_unique_display_layer_color(self) -> str:
-        existing_colors = {display_layer.fill_color for display_layer in self.display_layers}
-        for color in ('#1E90FF', "#F00", "#0F0", "#00F", "#FF0", "#0FF", "#F0F", "#888", "#800", "#080", "#008", "#CCC", "#880", "#088", "#808", "#964B00"):
-            if not color in existing_colors:
-                return color
-        return '#1E90FF'
+        existing_colors = [display_layer.fill_color for display_layer in self.display_layers]
+        unused_colors = list({*Color.svg_color_map.values()})
```
cmeyer, line 2003: I think this code can be simplified... just get the existing colors and then iterate through the color list looking for the first one that doesn't match anything in the existing colors and use it.

nion/swift/model/DisplayItem.py
```diff
@@ -1998,11 +1999,17 @@ def append_display_data_channel(self, display_data_channel: DisplayDataChannel,
...
+        matches = [False] * len(unused_colors)
+        for color_str in existing_colors:
+            source_color = Color.Color(color_str)
+            for index in range(len(matches)):
+                matches[index] |= source_color.matches_without_alpha(Color.Color(unused_colors[index]))
+        try:
+            return unused_colors[matches.index(False)]
+        except ValueError:
```
cmeyer, line 2011: As a side note -- I don't like using exceptions for regular code flow since it makes debugging more difficult. However, if you use the simplified algorithm above it won't be necessary anyway.

setup.cfg
```diff
@@ -25,7 +26,7 @@ install_requires =
     tzlocal
     imageio
     pillow
-    nionutils>=0.4.0,<0.5.0
+    nionutils>=0.4.0,<0.5.0v
```
cmeyer, line 29: Can you fix this and force rebase?

cmeyer (changes requested):

nion/swift/model/DisplayItem.py
gosselind1: I revised the code to be a bit more simple, but I think any more simplification would require the ability to sort and or directly have casting to  9 character hex colors

cmeyer:
I merged this (with changes) here: d3a7fb5ff16821b8c5b546395ebab9ac5959a713
I then simplified the color choosing algorithm here: f870cbdb88a60fd14984c77af74452d32309db54

These two changes do not close #657 so it is still open.

PR #868, opened 2022-07-08
Fix #640 Non-Fixed Place Import Menus Now Return to Previous Location

### LINKED ISSUE #640: Canceling import dialog clobbers the saved default import folder
2020-12-09 TL: There is an annoying bug when you are importing data to Swift. If you ever hit cancel, then it reverts to some nonsense defaults like ../../.../jbr/bin

The last folder location used to import data is always a more useful place to start.

gosselind1:
Import menus will now return to their previous location on reopen if nothing was selected for import

### DIFF
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -379,6 +379,8 @@ class Application(UIApplication.BaseApplication):
             filter_str = "Projects (*.nsproj);;Legacy Libraries (*.nslib);;All Files (*.*)"
             import_dir = ui.get_persistent_string("open_directory", ui.get_document_location())
             paths, selected_filter, selected_directory = ui.get_file_paths_dialog(_("Add Existing Library"), import_dir, filter_str)
+            if len(paths) == 0:
+                selected_directory = import_dir
             ui.set_persistent_string("open_directory", selected_directory)
             if len(paths) == 1:
                 project_reference = profile.open_project(pathlib.Path(paths[0]))
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -749,6 +749,8 @@ class DocumentController(Window.Window):
         filter += ";;All Files (*.*)"
         import_dir = self.ui.get_persistent_string("import_directory", "")
         paths, selected_filter, selected_directory = self.get_file_paths_dialog(_("Import File(s)"), import_dir, filter)
+        if len(paths) == 0:
+            selected_directory = import_dir
         self.ui.set_persistent_string("import_directory", selected_directory)
         self.receive_files(paths, display_panel=self.next_result_display_panel())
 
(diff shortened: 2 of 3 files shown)


nion/swift/Application.py
```diff
@@ -379,6 +379,8 @@ def show_open_project_dialog(self) -> None:
             filter_str = "Projects (*.nsproj);;Legacy Libraries (*.nslib);;All Files (*.*)"
             import_dir = ui.get_persistent_string("open_directory", ui.get_document_location())
             paths, selected_filter, selected_directory = ui.get_file_paths_dialog(_("Add Existing Library"), import_dir, filter_str)
+            if len(paths) == 0:
+                selected_directory = import_dir
             ui.set_persistent_string("open_directory", selected_directory)
```
cmeyer, line 384: This seems like a bit of pretzel logic ... why not just move the `ui.set_persistent_string` inside the `if` statement on the next line?

nion/swift/DocumentController.py
```diff
@@ -749,6 +749,8 @@ def import_file(self) -> None:
         filter += ";;All Files (*.*)"
         import_dir = self.ui.get_persistent_string("import_directory", "")
         paths, selected_filter, selected_directory = self.get_file_paths_dialog(_("Import File(s)"), import_dir, filter)
+        if len(paths) == 0:
```
cmeyer, line 752: Similar ... `if len(paths) > 0:` and then put lines 754+755 inside the `if` statement.

nion/swift/ScriptsDialog.py
```diff
@@ -403,6 +403,8 @@ def add_clicked() -> None:
             assert self.__profile
             add_dir = self.ui.get_persistent_string("import_directory", "")
             file_paths, filter_str, directory = self.get_file_paths_dialog(_("Add Scripts"), add_dir, "Python Files (*.py)", "Python Files (*.py)")
+            if len(file_paths) == 0:
```
cmeyer, line 406: Same...

cmeyer (changes requested):

nion/swift/Application.py
gosselind1: I could have something like:
`ui.set_persistent_string("open_directory", selected_directory if len(paths) else import_dir)` instead.
The main catch with the current logic setup is that the application saves the last folder the user used, and if we encapsulate `ui.set_persistent_string("open_directory", selected_directory)` in the if, the new directory will never be set. At which point it'd probably be simpler to use ui.get_document_location() as the import folder dialog does.

cmeyer: I've made the minimal changes I suggested here:
https://github.com/nion-software/nionswift/pull/871

Do you see any problem with that approach?

gosselind1: No that looks fine. I assume that was your original intent too, so I must have misunderstood your initial comment here.

cmeyer:
Final commit: 11e1a99c1e7a53415c6a5665fac07bbfa4668189

PR #870, opened 2022-07-14
Make view to interval work properly on stacked line plots

### LINKED ISSUE #866: Stacked line plots do not correctly view-to-interval
Steps to reproduce:
1. Generate two lineplots
2. Drag one over the other to create a stack
3. Add an interval graphic
4. Press enter to zoom in

Expected Behavior:
Zooms into a window looking similar to this:
image

Actual Behavior:
Zooms into a window which only accounts for the first layer (the light blue one in this case)
image

gosselind1:
This should be a proper fix to #866 assuming I have correctly understood the particularities of the scale attribute.

### DIFF
--- a/nion/swift/LinePlotCanvasItem.py
+++ b/nion/swift/LinePlotCanvasItem.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 # standard libraries
 import copy
+import logging
 import math
 import threading
 import typing
@@ -460,7 +461,7 @@ class LinePlotCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
             x_sale_factor = top_layer_x_scale / layer_x_scale
             layer_left = int(numpy.floor((((left * top_layer_width) + top_layer_x_offset) * x_sale_factor) - layer_x_offset))
             layer_left = max(0, min(layer_left, top_layer_width))
-            layer_right = int(numpy.floor((((right * top_layer_width) + top_layer_x_offset) * x_sale_factor) - layer_x_offset))
+            layer_right = int(numpy.ceil((((right * top_layer_width) + top_layer_x_offset) * x_sale_factor) - layer_x_offset))
             layer_right = min(top_layer_width, max(layer_right, 0))
 
             # get values of interval region for this particular stacked_data_layer
@@ -491,8 +492,8 @@ class LinePlotCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
 
         x_padding = (right - left) * 0.5
         # top layer width is used here because all other width/position values have been made relative to it
-        display_left_channel = int((left - x_padding) * top_layer_width)
-        display_right_channel = int((right + x_padding) * top_layer_width)
+        display_left_channel = numpy.ceil((left - x_padding) * top_layer_width)
+        display_right_channel = numpy.floor((right + x_padding) * top_layer_width)
 
         y_min = min(y_minimums) * 1.2
         y_max = max(y_maximums) * 1.2
--- a/nion/swift/LinePlotCanvasItem.py
+++ b/nion/swift/LinePlotCanvasItem.py
@@ -420,25 +420,26 @@ class LinePlotCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
 
     def __view_to_intervals(self, data_and_metadata: DataAndMetadata.DataAndMetadata, intervals: typing.List[typing.Tuple[float, float]]) -> None:
         """Change the view to encompass the channels and data represented by the given intervals."""
-        # X dimensional factors are controlled by dimensional_calibrations
-        # Y dimensional factors are controlled by intensity_calibration
 
-        if len(intervals) > 0:  # left and right are absolute positions on the display
+        # get the farthest left and right interval bounds to zoom into
+        if len(intervals) > 0:
             left = intervals[0][0]
             right = intervals[0][1]
             for interval in intervals:
                 left = min(interval[0], left)
                 right = max(interval[1], right)
-        else:
-            left = right = 0.0
+        else:  # default to the whole area of the display if no intervals exist
+            left = 0.0
+            right = 1.0
 
-        # Isolate static properties (at draw-time) of top-level layer
         top_layer_width = data_and_metadata.data_shape[-1]
         top_layer_x_offset = data_and_metadata.dimensional_calibrations[0].offset
         top_layer_x_scale = data_and_metadata.dimensional_calibrations[0].scale
         top_layer_y_scale = data_and_metadata.intensity_calibration.scale
         top_layer_y_offset = data_and_metadata.intensity_calibration.offset
 
+        # Because we don't know if an interval will contain values, we initialize two lists
+        # one for interval ranges, one for the whole stack_data_layer.
         selected_y_minimums: typing.List[float] = []
         selected_y_maximums: typing.List[float] = []
         global_y_minimums: typing.List[float] = []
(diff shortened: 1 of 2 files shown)


nion/swift/LinePlotCanvasItem.py
```diff
@@ -430,31 +430,46 @@ def __view_to_intervals(self, data_and_metadata: DataAndMetadata.DataAndMetadata
...
-            data_min = numpy.amin(channel_data[..., :])
-            data_max = numpy.amax(channel_data[..., :])
+        top_level_data_size = int(data_and_metadata.data_shape[-1])
+        top_level_offset = data_and_metadata.intensity_calibration.offset
+        top_level_scale = data_and_metadata.intensity_calibration.scale
+
+        y_minimums = []
+        y_maximums = []
+
+        for display_value in self.__display_values_list:
```
cmeyer, line 440: It's better to just use the `__xdata_list` directly rather than asking for the adjusted data values. The `__xdata_list` is the actual drawn data at any given moment.

nion/swift/LinePlotCanvasItem.py
```diff
@@ -430,31 +430,46 @@ def __view_to_intervals(self, data_and_metadata: DataAndMetadata.DataAndMetadata
...
+            adjusted_data_and_metadata = display_value.adjusted_data_and_metadata
+            assert adjusted_data_and_metadata is not None
+            adjusted_data = adjusted_data_and_metadata.data
+            data_size = adjusted_data_and_metadata.data_shape[-1]
+            offset = adjusted_data_and_metadata.intensity_calibration.offset
+            scale = adjusted_data_and_metadata.intensity_calibration.scale  # Scale *should* handle frequencies
+            scaling_factor = scale / top_level_scale
+            offset_factor = int(numpy.round(top_level_offset - (offset * scaling_factor)))
+
+            left_channel = int(numpy.floor(top_level_data_size * left * scaling_factor) + offset_factor)
```
cmeyer, line 451: Beware of mixing calibrations -- channels (x-values) are calibrated using `dimensional_calibrations` (0 index in most cases); intensity (y-values) are calibrated using `intensity_calibration`. `scaling_factor` is based on an intensity calibration (y-value), but used to calculate a channel (x-value).

cmeyer: Also, you'll want to just use the `Calibration` object methods directly:
https://github.com/nion-software/niondata/blob/f17642bc0cdee2c5d8dc77144486171d005fba3c/nion/data/Calibration.py#L112-L140

nion/swift/test/DisplayPanel_test.py
```diff
@@ -2564,6 +2564,41 @@ def test_interval_zoom(self):
                 self.assertAlmostEqual(axes.drawn_left_channel, (min(interval) - padding) * 1024)
                 self.assertAlmostEqual(axes.drawn_right_channel, (max(interval) + padding) * 1024)
 
+    def test_stacked_lineplot_interval_zoom(self):
```
cmeyer, line 2567: I'm curious why this test passes when the code is mixing intensity/dimensional calibrations. It's probably worth revisiting this test before making further changes to ensure it is testing the right thing.

It's easy to construct data that fails: just make one line plot 16x1 with a value of 1 (or random) and intensity calibration of 1.0 scale and units of 'z'. Then make another 16x1 with value of 1 (or random) and intensity calibration of 8.0 scale and units of 'z'. Drag the 2nd onto the first, make an interval, and view to interval. It will fail.

cmeyer (changes requested):

nion/swift/LinePlotCanvasItem.py
```diff
@@ -430,35 +432,66 @@ def __view_to_intervals(self, data_and_metadata: DataAndMetadata.DataAndMetadata
         else:
             left = right = 0.0
```
cmeyer, line 433: This line leads to undesired behavior - it should set `left = 0.0` and `right = 1.0`. Can you add a test for this? Hitting <enter> with no interval selected should reset the view to its default.

nion/swift/LinePlotCanvasItem.py
```diff
@@ -420,8 +420,10 @@ def convert_to_calibrated_size_str(f: float) -> str:
 
     def __view_to_intervals(self, data_and_metadata: DataAndMetadata.DataAndMetadata, intervals: typing.List[typing.Tuple[float, float]]) -> None:
         """Change the view to encompass the channels and data represented by the given intervals."""
+        # X dimensional factors are controlled by dimensional_calibrations
```
cmeyer, line 423: I don't think these comments belong here. They're a general knowledge about the calibrations/line plot and I think they're just extra here since they could literally be applied to almost any function in this file.

nion/swift/LinePlotCanvasItem.py
```diff
@@ -420,8 +420,10 @@ def convert_to_calibrated_size_str(f: float) -> str:
...
+        # Y dimensional factors are controlled by intensity_calibration
 
-        if len(intervals) > 0:
+        if len(intervals) > 0:  # left and right are absolute positions on the display
```
cmeyer, line 426: Likewise with this comment - I don't understand what it means. Can you either clarify or remove it? The intervals are relative coordinates, typically ranging from 0.0 to 1.0; they're not absolute.

What would be helpful, though, is a comment before this section describing that this section of code is finding the min/max values of the set of intervals.

nion/swift/LinePlotCanvasItem.py
```diff
@@ -430,35 +432,66 @@ def __view_to_intervals(self, data_and_metadata: DataAndMetadata.DataAndMetadata
...
-        if channel_data[..., left_channel:right_channel].size > 0:
-            data_min = numpy.amin(channel_data[..., left_channel:right_channel])
-            data_max = numpy.amax(channel_data[..., left_channel:right_channel])
-        else:
-            data_min = numpy.amin(channel_data[..., :])
-            data_max = numpy.amax(channel_data[..., :])
-
-        y_min = 0.0 if data_min > 0 else data_min * 1.2
-        y_max = 0.0 if data_max < 0 else data_max * 1.2
+        # Isolate static properties (at draw-time) of top-level layer
```
cmeyer, line 435: ... again, not sure what this comment means. Is it necessary?

nion/swift/LinePlotCanvasItem.py
```diff
@@ -430,35 +432,66 @@ def __view_to_intervals(self, data_and_metadata: DataAndMetadata.DataAndMetadata
...
+        top_layer_x_scale = data_and_metadata.dimensional_calibrations[0].scale
+        top_layer_y_scale = data_and_metadata.intensity_calibration.scale
+        top_layer_y_offset = data_and_metadata.intensity_calibration.offset
+
+        selected_y_minimums: typing.List[float] = []
+        selected_y_maximums: typing.List[float] = []
+        global_y_minimums: typing.List[float] = []
+        global_y_maximums: typing.List[float] = []
+
+        for stacked_data_layer in self.__xdata_list:
```
cmeyer, line 447: Overall, this algorithm is confusing. Can you write an explanation of how it is supposed to work? For instance, why do you have selected and global min/max arrays? What are the units of the min/max values stored in either array? What is the purpose of the `if layer_values.size <= 0:` line? etc.

Also, assigning a variable (y_min/maximums) to one of the the two arrays (selected_y_min/maximums vs global_y_min/maximums) and then appending values to it is what I would consider "tricky" code. Can you rewrite that to be more straightforward?

nion/swift/test/DisplayPanel_test.py
```diff
@@ -2564,6 +2564,48 @@ def test_interval_zoom(self):
...
```
cmeyer, line 2567: This test is good, but it is very specific. Can you change the name of the test to reflect that: "test_stacked_lineplot_interval_zoom_with_different_intensity_calibration_scales". Some other tests might be:
- test_stacked_lineplot_interval_zoom_with_different_intensity_calibrations_with_no_selected_intervals
- test_stacked_lineplot_interval_zoom_with_different_intensity_calibrations_with_different_intensity_units
- test_stacked_lineplot_interval_zoom_with_different_intensity_calibrations_with_multiple_intervals

cmeyer (changes requested):
This is getting closer, but I still think it needs some work, particularly with the comments/explanation. The actual algorithm looks pretty close too, but, again, read the comments about questions I have about the algorithm.

cmeyer:
Merged this after making some significant edits to simplify the algorithm, adding comments, and cleaning up the tests.

Issue #591, opened 2021-01-16
Exporting data items with forward slashes results in wrong filename


cmeyer:
Forward slashes aren't allowed in Windows filenames.

Exporting a single file with forward slashes works on macOS. It is not allowed (some versions of Qt?) on Windows.

Exporting a batch with forward slashes in the prefix silently fails on both macOS and Windows.

Also, how are backslashes and other disallowed characters handled?

Is there any way to escape a filename and remove disallowed characters that works cross platform?

---

Edit 2024-05-14: This issue only applies to batch exports or possibly the default file name used when exporting a single file. In those places, we should ensure that if we generate a filename from a title, it is sanitized for use as a filename. See #874 for some initial, abandoned work. We can assume that a filename the user chooses a filename in a save file dialog is sanitized; it may still fail but will be reported to the user.

---

```[tasklist]
### Tasks
- [ ] Propose an algorithm to sanitize an arbitrary title string
- [ ] Consider platform specific restrictions, unicode, special characters including slashes, spaces, parethensis, etc.
- [ ] Consider restrictions on the length, i.e. limit to a length of 64?
- [ ] Implement algorithm, include tests (input strings -> expected output strings)
- [ ] Integrate algorithm into export dialog where we propose the initial title
- [ ] Integrate algorithm into export dialog where we allow user to export group of items
- [ ] Export batch code is in `do_export` in `ExportDialog.py`
- [ ] For extra credit, move `do_export` functionality into an `Action` object in `DocumentController.py` for scripting
```

cmeyer:
My initial thought is that the `path` should be checked and cleaned up around here:

https://github.com/nion-software/nionswift/blob/6dd93014207b8f805cd700f06d236aa10105768a/nion/swift/DocumentController.py#L774

gosselind1:
I've dug a bit through this one. Addressing the big question of if there is an easy cross-platform way to sanitize paths, the answer is no.

Though there are some attempts at resolving this, ex: the first answer of https://stackoverflow.com/questions/9532499/check-whether-a-path-is-valid-in-python-without-creating-a-file-at-the-paths-ta, not only does it note it has issues, but it also misses issues such as restricted file names, ex: "NUL", along with not even considering UNC paths on the windows side. (As an additional note, most OSs will prevent or attempt to prevent invalid or disliked file names within their graphical and text shells for the most part. Windows for example will not let you use a file with "/" in it when utilizing the explorer selection window.)

[...]

PR #874, opened 2022-08-06
Export Filename Sanitization and Directory Expansion for #591

gosselind1:
This set of patches fixes the bigger issue with #591 that it used to allow for non-valid characters to be attempted to be written.

Now, the program should sanitize file names down to the POSIX portable character set, plus spaces, though this process does occur silently. If slashes are included in the provided file name, the exporter will now attempt to expand those slashes as directories, and create the directories rather than attempt to write a file with slashes in its name.

One of the other issues still remaining about #591 is that swift will not notify the user if certain if checks within the exporting code fail, and result in the file not being written. This would probably require a unification of the exporting logic paths into a single one

### DIFF
--- a/nion/swift/test/DocumentController_test.py
+++ b/nion/swift/test/DocumentController_test.py
@@ -2,6 +2,8 @@
 import contextlib
 import gc
 import logging
+import os
+import tempfile
 import unittest
 import weakref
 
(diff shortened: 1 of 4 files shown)


gosselind1:
The current set of tests are failing for those awkward hard to reproduce issues I've been getting intermittently, not for the related file changes

nion/swift/test/DocumentController_test.py
```diff
@@ -1177,6 +1181,51 @@ def test_get_3_data_sources_with_two_selected_in_data_panel(self):
             self.assertEqual(2, len(document_controller._get_n_data_sources(2)))
             self.assertEqual(3, len(document_controller._get_n_data_sources(3)))
 
+    def test_filenames_in_exports(self):
+        # The bmp writer is a dumb writer, and used to avoid complex setup code
+        FILE_NAMES = ["test.bmp", "/././////.test.bmp", "$$%/%%_test.bmp", "\\aaaaa//test.bmp"]
```
cmeyer, line 1186: There are still quite a few failing cases that aren't covered. Some examples on macOS are:

`FILE_NAMES = ["a\N{WARNING SIGN}b", "\N{WARNING SIGN}", ".", "abc\n\rdef", "a\tb", "µscope"]`

Also, valid files should not start with a `.` (period or dash).

cmeyer:
After looking through this issue and bringing it up to date with current main branch, I've decided to close this. See notes in #591.

PR #894, opened 2022-10-18
Add support for help files to ScriptsDialog.

Brow71189:
This adds a long-awaited feature: Help files.

I chose html as the internal format, because it gives you maximum flexibility for the content and they can be viewed on any computer on all platforms.

To facilitate writing help files I chose to include an automatic conversion feature using pandoc if no html help file is found but other formats are present. Without pandoc installed we only show html files directly.

Pandoc is available on conda without any dependencies, so it is an easy extra package to install if we choose to do so.

### DIFF
--- a/nion/swift/ScriptsDialog.py
+++ b/nion/swift/ScriptsDialog.py
@@ -252,14 +252,11 @@ def open_location(location: str) -> None:
 
 
 def convert_help_file(path: str) -> str:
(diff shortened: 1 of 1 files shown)


Brow71189:
I also looked into just linking our documentation in the missing help file message, but this does not look nice with the current UI, because you cannot copy text from notifications. Ideally you would even be able to just click on a link, but at the very minimum users need to be able to copy text before we can include links in notifications, otherwise it is just too awkward.

PR #917, opened 2022-12-14
Add 1d rebin computation and make menu action set up the correct one depending on data dims.

Brow71189:
This change passes in the correct number of arguments depending on whether the input data is 1d or 2d.

This is a suggestion for how I would do this. If we are happy with it we can also adjust the other computations (resize, resample).

### DIFF
--- a/nion/swift/model/DocumentModel.py
+++ b/nion/swift/model/DocumentModel.py
@@ -2471,16 +2471,18 @@ class DocumentModel(Observable.Observable, ReferenceCounting.ReferenceCounted, D
                 "sources": [{"name": "src", "label": _("Source"), "croppable": True}], "parameters": [do_transpose_param, do_flip_v_param, do_flip_h_param]}
             width_param = {"name": "width", "label": _("Width"), "type": "integral", "value": 256, "value_default": 256, "value_min": 1}
             height_param = {"name": "height", "label": _("Height"), "type": "integral", "value": 256, "value_default": 256, "value_min": 1}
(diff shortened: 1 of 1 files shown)


nion/swift/model/DocumentModel.py
```diff
@@ -2471,8 +2471,12 @@ def _get_builtin_processing_descriptions(cls) -> typing.Dict[str, typing.Any]:
                 "sources": [{"name": "src", "label": _("Source"), "croppable": True}], "parameters": [do_transpose_param, do_flip_v_param, do_flip_h_param]}
             width_param = {"name": "width", "label": _("Width"), "type": "integral", "value": 256, "value_default": 256, "value_min": 1}
             height_param = {"name": "height", "label": _("Height"), "type": "integral", "value": 256, "value_default": 256, "value_min": 1}
-            vs["rebin"] = {"title": _("Rebin"), "expression": "xd.rebin_image({src}.cropped_display_xdata, (height, width))",
```
cmeyer, line 2474: Similar to my comment in `niondata`, can you _add_ a new function `rebin_factor` and leave the old one in place for backwards compatibility?

cmeyer (changes requested):

cmeyer:
Thanks! Merged e735d32370421aaee071f296d2ab27e1bc9df2e1.

One note: rebin will not be resilient to changes in the dimensions of the source data, i.e. if source data changes from 1D to 2D or back, since the choice is made at the time the menu item is chosen. It may be possible for this to happen when the user uses the redimension menu items. It's not a case we can handle currently anyway; I'm just making a note of it here for future reference.

PR #1050, opened 2024-05-15
Refactor export svg to isolate actual file writing code.

### LINKED ISSUE #591: Exporting data items with forward slashes results in wrong filename

cmeyer:
This is minor refactoring to isolate the actual code which exports the SVG file. Refactoring like this allows for tests to run SVG exports easily. This was part of my attempt to evaluate #874 which attempts to fix #591. However, I've decided to abandon #874 for now - and I'm planning to add more notes to #591.

References:
- #591 
- #874

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -53,6 +53,7 @@ from nion.swift.model import Processing
 from nion.swift.model import Profile
 from nion.swift.model import Project
 from nion.swift.model import Symbolic
+from nion.swift.model import UISettings
 from nion.swift.model import Utility
 from nion.swift.model import WorkspaceLayout
 from nion.ui import CanvasItem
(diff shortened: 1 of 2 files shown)


cmeyer:
@lisham2000 since this affects code you may be working on, please let me know if you want me to merge this. Alternatively you can use "git cherry-pick" to add this commit to your own changes. If you think this will merge cleanly with your changes (or if you haven't started on your changes), please let me know and I[ll merge this right away.

Issue #586, opened 2020-12-15
Export batch with no boxes checked overwrites same file repeatedly


cmeyer:
When the user is exporting a set of files by selecting multiple display items and then choosing File > Export, the user is presented with a dialog that allows them to set the file naming template.

<img width="313" alt="image" src="https://github.com/nion-software/nionswift/assets/34664/43e0faa2-4c13-4cee-8c6c-aed9941eb571">

If they deselect all items, or if the combination of check boxes results in duplicate filenames, the export procedure simply overwrites the earlier export with the same filename.
```[tasklist]
### Tasks
- [ ] Rewrite `do_export` in `ExportDialog.py` to its own function external to the `ExportDialog` class
- [ ] In `do_export`, generate a map of display items to filenames, ensuring a unique filename for each display item
- [ ] Extra credit: Rewrite `ExportDialog` to use declarative (see how `ExportSVGHandler` does it)
- [ ] Extra credit: Move the functionality of `do_export` to a `Action` in `DocumentController.py` for scriptability, testing
```

cmeyer:
XY reports "If I export multiple data with the same name (e.g. NionCCD-1010), only one file can be saved. I can check "Include Sequence Number" to allow them all to be saved. However, could you make this batch export smarter to automatically add sequence numbers if several items use the same file name."

cmeyer:
The UI for the batch export resides in [`ExportDialog.py`](https://github.com/nion-software/nionswift/blob/master/nion/swift/ExportDialog.py).

Once the user has configured the options, the [`do_export`](https://github.com/nion-software/nionswift/blob/34bc02b87347cb0f939a546f9d42679ae2c549bd/nion/swift/ExportDialog.py#L159) method loops through the selected display items and exports them to a file using [write_display_item_with_writer](https://github.com/nion-software/nionswift/blob/34bc02b87347cb0f939a546f9d42679ae2c549bd/nion/swift/ExportDialog.py#L185).

cmeyer:
Fixed 907bf563f6798a01d31c5848861dbded08428c9d

Ion-e:
Issue verified as working correctly.

Thanks!

PR #1055, opened 2024-05-23
Refactored Export Dialog in Declarative Mode

Tiomat85:
#586   Fixed the bug with no checkboxes led to duplicate filenames, stopping all duplicate names happening. Reimplemented in Declarative UI model.
Extracted the settings to a dataclass that can be re-used for scripting. Extracted the export functionality to a static function that can be re-used for scripting.

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -888,9 +888,9 @@ class DocumentController(Window.Window):
 
     def export_files(self, display_items: typing.Sequence[DisplayItem.DisplayItem]) -> None:
         if len(display_items) > 1:
-            export_dialog = ExportDialog.ExportDialog(self.ui, self)
-            export_dialog.on_accept = functools.partial(export_dialog.do_export, display_items)
-            export_dialog.show()
+            export_dialog = ExportDialog.ExportDialog(self.ui, self, self, display_items)
+            # export_dialog.on_accept = functools.partial(export_dialog.do_export, display_items)
+            # export_dialog.show()
         elif len(display_items) == 1:
             self.export_file(display_items[0])
 
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -889,8 +889,6 @@ class DocumentController(Window.Window):
     def export_files(self, display_items: typing.Sequence[DisplayItem.DisplayItem]) -> None:
         if len(display_items) > 1:
             export_dialog = ExportDialog.ExportDialog(self.ui, self, self, display_items)
-            # export_dialog.on_accept = functools.partial(export_dialog.do_export, display_items)
-            # export_dialog.show()
         elif len(display_items) == 1:
             self.export_file(display_items[0])
 
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -46,7 +46,7 @@ class ExportDialogViewModel:
     directory: Model.PropertyModel[str]
     writer: Model.PropertyModel[ImportExportManager.ImportExportHandler]
 
-    def __init__(self, title: bool, date: bool, dimensions: bool, sequence: bool, writer: ImportExportManager.ImportExportHandler, prefix: str = "", directory: str = ""):
+    def __init__(self, title: bool, date: bool, dimensions: bool, sequence: bool, writer: typing.Optional[ImportExportManager.ImportExportHandler], prefix: str = "", directory: str = ""):
         self.include_title = Model.PropertyModel(title)
         self.include_date = Model.PropertyModel(date)
         self.include_dimensions = Model.PropertyModel(dimensions)
@@ -58,11 +58,12 @@ class ExportDialogViewModel:
             self.include_prefix = Model.PropertyModel(False)
             self.prefix = Model.PropertyModel(None)
         self.directory = Model.PropertyModel(directory)
-        self.writer = Model.PropertyModel(writer)
+        if writer:
+            self.writer = Model.PropertyModel(writer)
 
 
 class ExportDialog(Declarative.Handler):
-    writer: ImportExportManager|None = None
+    writer: typing.Optional[ImportExportManager.ImportExportHandler] = None
     def __init__(self, ui: UserInterface.UserInterface, parent_window: Window.Window, document_controller: DocumentController.DocumentController, display_items: typing.Sequence[DisplayItem.DisplayItem]):
         super().__init__()
 
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -46,7 +46,7 @@ class ExportDialogViewModel:
     directory: Model.PropertyModel[str]
     writer: Model.PropertyModel[ImportExportManager.ImportExportHandler]
 
-    def __init__(self, title: bool, date: bool, dimensions: bool, sequence: bool, prefix: str = "", directory: str = "", writer: ImportExportManager.ImportExportHandler = None):
+    def __init__(self, title: bool, date: bool, dimensions: bool, sequence: bool, writer: ImportExportManager.ImportExportHandler, prefix: str = "", directory: str = ""):
         self.include_title = Model.PropertyModel(title)
         self.include_date = Model.PropertyModel(date)
         self.include_dimensions = Model.PropertyModel(dimensions)
@@ -71,7 +71,7 @@ class ExportDialog(Declarative.Handler):
         self.directory = self.ui.get_persistent_string("export_directory", self.ui.get_document_location())
         self.writer = ImportExportManager.ImportExportManager().get_writer_by_id(io_handler_id)
 
-        self.viewmodel = ExportDialogViewModel(True, True, True, True, "", self.directory, self.writer)
+        self.viewmodel = ExportDialogViewModel(True, True, True, True, self.writer, "", self.directory)
         u = Declarative.DeclarativeUI()
         self._build_ui(u)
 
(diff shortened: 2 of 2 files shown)


nion/swift/DocumentController.py
```diff
@@ -888,9 +888,9 @@ def export_file(self, display_item: DisplayItem.DisplayItem) -> None:
 
     def export_files(self, display_items: typing.Sequence[DisplayItem.DisplayItem]) -> None:
         if len(display_items) > 1:
-            export_dialog = ExportDialog.ExportDialog(self.ui, self)
-            export_dialog.on_accept = functools.partial(export_dialog.do_export, display_items)
-            export_dialog.show()
+            export_dialog = ExportDialog.ExportDialog(self.ui, self, self, display_items)
+            # export_dialog.on_accept = functools.partial(export_dialog.do_export, display_items)
+            # export_dialog.show()
```
cmeyer, line 893: As a general note, can you remove unused code unless there is a compelling reason to leave it around in comments?

nion/swift/ExportDialog.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 # standard libraries
+from dataclasses import dataclass
```
cmeyer, line 4: Can you follow the import/qualified-name convention used through the software? i.e. `import dataclasses`; then use `@dataclasses.dataclass` (i.e. fully qualified name).

In this case, `dataclass` seems unused, so can be removed entirely.

nion/swift/ExportDialog.py
```diff
@@ -17,6 +18,7 @@
 
 # local libraries
 from nion.swift.model import ImportExportManager
+from nion.swift.model import DisplayItem
```
cmeyer, line 21: We usually put imports within each section in alphabetical order.

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
-        super().__init__(ui, ok_title=_("Export"), parent_window=parent_window)
+class ExportDialogViewModel:
+    include_title: Model.PropertyModel[bool]
+    include_date: Model.PropertyModel[bool]
+    include_dimensions: Model.PropertyModel[bool]
+    include_sequence: Model.PropertyModel[bool]
+    include_prefix: Model.PropertyModel[bool]
+    prefix: Model.PropertyModel[str]
+    directory: Model.PropertyModel[str]
+    writer: Model.PropertyModel[ImportExportManager.ImportExportHandler]
```
cmeyer, line 47: These are not necessary to declare here since they are initialized in the `__init__` method. I'm guessing you were messing around with `dataclass` which requires you to put them here.

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
+        self.include_title = Model.PropertyModel(title)
+        self.include_date = Model.PropertyModel(date)
+        self.include_dimensions = Model.PropertyModel(dimensions)
+        self.include_sequence = Model.PropertyModel(sequence)
+        if prefix:
+            self.include_prefix = Model.PropertyModel(True)
+            self.prefix = Model.PropertyModel(prefix)
+        else:
+            self.include_prefix = Model.PropertyModel(False)
+            self.prefix = Model.PropertyModel(None)
```
cmeyer, line 59:
```
self.include_prefix = Model.PropertyModel(prefix is not None)
self.prefix = Model.PropertyModel(prefix)
```

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
+        self.directory = Model.PropertyModel(directory)
+        if writer:
+            self.writer = Model.PropertyModel(writer)
 
-        io_handler_id = self.ui.get_persistent_string("export_io_handler_id", "png-io-handler")
 
+class ExportDialog(Declarative.Handler):
+    writer: typing.Optional[ImportExportManager.ImportExportHandler] = None
```
cmeyer, line 66: This line is redundant and unnecessary.

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
+        writer = self.writers[current_index]
+        self.viewmodel.writer.value = writer
+
+    def _build_ui(self, u: Declarative.DeclarativeUI) -> None:
+        self.file_type_index = 0
+        self.writers = ImportExportManager.ImportExportManager().get_writers()
+        writers_names = [getattr(writer, "name") for writer in self.writers]
+        self.file_types = writers_names
+
+        #  self.viewmodel = viewmodel
```
cmeyer, line 106: Remove

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
+
+        # Export Folder
+        directory_label = u.create_row(u.create_label(text="Location:"))
+        directory_text = u.create_row(u.create_label(text=f'@binding(viewmodel.directory.value)'))
```
cmeyer, line 110: Can you use " instead of ' here? It's a minor thing, but I'd like to be consistent.

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
+        self.directory_text_label = directory_text
+        directory_button = u.create_row(u.create_push_button(text="Select Path..", on_clicked='choose_directory'))
```
cmeyer, line 112: `_("Select Path...")` (i.e. make it translatable; and use 3 dots)

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
+
+        # Filename
+        filename_label = u.create_row(u.create_label(text="Filename:"))
+        ## Title
```
cmeyer, line 116: Nit: needs space before this line.

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
+                                 dimension_checkbox,
+                                 sequence_checkbox,
+                                 prefix_checkbox,
+                                 prefix_textbox,
+                                 file_type_combobox,
+                                 spacing=12, margin=12)
+        self.ui_view = column
+
+    @staticmethod
+    def build_filename(components: typing.List[str], extension: str, suffix: str = "($)", path: str = "") -> str:
```
cmeyer, line 151: The `suffix` argument is unused. This looks partially implemented. So... when splitting into the two PR's, you can probably just drop it. Then let's discuss the unique filename argument. I'd rather see a slightly more holistic approach where we keep track of what filenames have been used and only then make the subsequent matching filenames unique. I think it's difficult to do it with a single `build_filename` method unless a `dict` of used base filenames and their count is passed in.

nion/swift/ExportDialog.py
```diff
@@ -34,167 +36,191 @@
...
+            next_index = 1
+            max_index = 999
+            while not found_available and next_index <= max_index:
+                test_suffix = suffix.replace("$", str(next_index))
+                test_filename = filename + test_suffix + extension
+                if path:
+                    test_filename = path + test_filename
+                if not os.path.exists(test_filename):
+                    return test_filename
+                next_index = next_index + 1
```
cmeyer, line 184: This seems weird. But see comment above.

cmeyer (changes requested):
Thanks for the PR - this is a good start and heading in the right direction. It will be nice to have the export dialog be declarative once we get this merged.

I'm adding some individual comments - but would it be possible to split this PR into two: (1) switching the dialog to declarative; and (2) the fix for avoiding duplicate file names?

I've added comments on both parts here, but splitting it into two PR's will make the submit-review-edit cycle easier. I'll have additional comments on the declarative part once it is split into its own PR.

nion/swift/ExportDialog.py
Tiomat85: Ok, gotcha if that is the convention you go with. Noted for future. As for it being unused, yeah I was testing out dataclasses to stretch my knowledge a bit but it ended up being a bit too isolated that way if we wanted to extend in future.  From what I could glean there didn't seem to be any major benefits for the dataclass other than some extra automatic boilerplate.

nion/swift/ExportDialog.py
Tiomat85: I can't remember if this was one I did when battling with the type-checker on pull or not, but is there a downside to doing this (aside from more words). It seems throughout the rest you are trying to be as type specific as possible, so trying to find where that line is.

cmeyer: I'll refer you to some sections of Python (first link is general section about classes; 2nd link is specific answer to your question above):

https://docs.python.org/3/tutorial/classes.html#class-objects
https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables

In a nutshell, instance variables should be declared in `__init__` and in doing so, they will have enough typing information to fully specify the type. So no need for the redundant typing information, i.e. line 40 is redundant due to line 50 (the correct line).

nion/swift/ExportDialog.py
cmeyer: We can discuss here; but at this point, it's unused code, so it can be removed from this PR, right?

Tiomat85: Ok so the suffix part has been removed as you are correct, it is unused.  Onto the actual unique filename bit.

So the requirement as I understood it was to make sure that when exporting files to not automatically overwrite so the user did not lose data, especially in the case where the parameters selected meant that multiple exported images would have the same filename.  The simplest solution to this is to follow the Windows paradigm where if attempting to create a file with the same name to simply add an incrementing suffix in brackets after the filename.

You have mentioned some dictionary of data items to filenames, but I have some questions about that.  Does it actually provide any extra benefit over the standard Windows method? How should it handle already existing files in that directory that are not in that dictionary?

cmeyer:
@Tiomat85 Thanks for the PR - looks overall good - although I ended up making more changes. Since the changes were more than seemed sensible in the comments, I made a new PR with my changes, and added detailed explanations here:

https://github.com/nion-software/nionswift/pull/1065

If you're happy with the additional changes on the new PR, please close this PR and we'll work from the other one instead.

PR #1056, opened 2024-05-24
Changing the SVG export to only require height an accept multiple units

lisham2000:
Changing SVG export to only consider height and maintain aspect ratio

Change SVG export to have multiple units


cmeyer:
This doesn't look like it's ready for review, so this comment might be preliminary - but can you submit a PR for the "only require user to enter height" task separately from a PR for "allow user to specify units when exporting"? Thanks.

lisham2000:
Closing this due to the revaluation of pull request process

PR #1061, opened 2024-05-31
Change to only consider height in SVG

lisham2000:
Only considering height in SVG Export

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -206,9 +206,9 @@ class ExportSVGHandler:
         self.int_converter = Converter.IntegerToStringConverter()
 
         u = Declarative.DeclarativeUI()
-        width_row = u.create_row(u.create_label(text=_("Width (in)"), width=80), u.create_line_edit(text="@binding(width_model.value, converter=int_converter)"), spacing=12)
+
         height_row = u.create_row(u.create_label(text=_("Height (in)"), width=80), u.create_line_edit(text="@binding(height_model.value, converter=int_converter)"), spacing=12)
-        main_page = u.create_column(width_row, height_row, spacing=12, margin=12)
+        main_page = u.create_column(height_row, spacing=12, margin=12)
 
         self.ui_view = main_page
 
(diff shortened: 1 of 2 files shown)


nion/swift/ExportDialog.py
```diff
@@ -241,9 +241,14 @@ def __init__(self, document_controller: DocumentController.DocumentController, d
 
         def ok_clicked() -> bool:
             dpi = 96
-            width_px = (handler.width_model.value or display_size.width) * dpi
-            height_px = (handler.height_model.value or display_size.height) * dpi
+            if display_item.display_data_shape is not None:
+                display_item_ratio = display_item.display_data_shape[1] / display_item.display_data_shape[0]
```
cmeyer, line 245: This line (`display_item_ratio = ...`) makes the assumption that the display data has two dimensions, resulting in failure if you try to export a 1D line plot to SVG. The logic for the `display_size` is already available on lines 235-238.

I suggest the following:
- update the logic on lines 235-238 to default to a height of 4 for both 1D and 2D cases
- in the 2D case, use the the data aspect ratio to determine the width
- in the 2D case, limit aspect ratio to 32:1 so that an image that is 2048x1 doesn't crash SVG (this may need more thinking)
- in the 1D case, use the golden ratio as the aspect ratio, width = 4 * 1.61803, height = 4
- calculate the display_item_ratio immediately after line 238
- recalculate the height_px, width_px size where you do now lines 250-251
- this is another potential spot to limit the SVG export size, maybe no larger than 16384 pixels in either dimension?

nion/swift/ExportDialog.py
```diff
@@ -241,9 +241,14 @@ def __init__(self, document_controller: DocumentController.DocumentController, d
...
+            else:
+                display_item_ratio = 1
+
```
cmeyer, line 248: This is an extra blank line - to keep the source code easy to read, only one blank line needs to appear between code sections. This code should go away, but it's good to have clean white space changes, if any.

nion/swift/ExportDialog.py
```diff
@@ -254,6 +259,7 @@ def ok_clicked() -> bool:
             if path:
                 ui.set_persistent_string("export_directory", selected_directory)
                 display_shape = Geometry.IntSize(height=height_px, width=width_px)
+
```
cmeyer, line 262: Unless there is a reason for it, remove this white space change. IMHO adding a blank line here doesn't add to readability. If anything, add a comment.

cmeyer (changes requested):
Needs changes (fails to export 1D line plot).

Issue #1051, opened 2024-05-20
When exporting to SVG, improve options for sizing data items


cmeyer:
User may want to export multiple data items (2D or 1D) such that they all have matching height.

User may want to export 2D multiple data items such that they have match resolution.

2D items should be exported at their aspect ratio.

1D items should be exported with a nice ratio (5:3, approximately the golden ratio).

cmeyer:
If we only ask for height, it is extremely difficult to get multiple 2D image SVG exports with different aspect ratios to match in resolution and I think having a matching resolution between exports is a desired use case in addition to the use case of having the same height for multiple exports.

I think we should go back to something where the user can specify width, height, or resolution, with the other two being dependent on whichever one the user chooses to specify. Something like this, where two of the values are dimmed (auto calculated) depending on the last value the user entered.

<img width="315" alt="image" src="https://github.com/nion-software/nionswift/assets/34664/4062185e-4f53-468b-9dc3-f99cf0e4630d">

bentps:
I'm a little unclear about what resolution means in the context of vector format like SVG. My understanding (happy to be corrected) is that the practical implication of the resolution parameter in this context is determining the relative size of the font with respect to the image dimensions. Is that correct?

cmeyer:
For 2D images, when drawing to SVG, we draw all of the pixels. As @ndellby pointed out, one option might be to just leave it at that. A typical SVG viewer draws pixels at 96dpi, so a 1024x1024 would display as 10.66in x 10.66in. A 256x256 would display as 2.66in x 2.66in. And a 96x96 image would display as 1in x 1in.

However, we can into problems when we add graphics, text, and the scale marker.

Lines will typically draw at 1 pixel on the screen and text at something like 14pt in the UI of Swift. So a rectangle on either a 256x256 or 1024x1024 image (or any other image) will draw at 1 pixel in the UI. Likewise, text will always draw as 14pt.

So when we export to SVG, if we draw the image at its pixel size, then the lines and text will draw relative to their pixel size. A 256x256 image will have text that is "14pt" or takes up 96dpi x 14px / 256px = 5.25 "dots per inch of the screen" or (5.5%) of the height of the image. A 1024x1024 image will have text that is "14pt" or takes up 96x14/1024 = 1.31 "dots per inch of the screen" (1.4%) of the height of the image.

[...]

PR #1062, opened 2024-05-31
Units considered SVG export

lisham2000:
Allows selection of centimeters inches or pixels within SVG Exports


ndellby:
Since our images have true size, typically in nm, and drawing programs are unlikely to play well with that.
Why do we need anything but pixels?

cmeyer:
The SVG standard includes a viewBox attribute that allows us to indicate how large the SVG graphic should appear in a document. So this is a convenience for the user.

Note: this is a pull request, so the discussion here will only apply to this PR, which is likely to be abandoned or modified - we can discuss the specification here, though: https://github.com/nion-software/nionswift/issues/1051

PR #1065, opened 2024-06-11
ExportDialogDeclarative2

cmeyer:
- **Refactored Export Dialog in Declarative Mode**
- **Do not pass redundant/unused 'parent_window'. Do not assign export dialog to unused variable.**
- **Clean up instance variables in export dialog and its model.**
- **Ensure that correct writer index is initialized.**
- **Change export filename building to use pathlib.**
- **Improve dialog box with default width and stretches.**
- **Ensure selected writer is saved in preferences when export is pressed.**
- **Fix issue where filename has a '.' (isodate on mac/linux).**

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -888,7 +888,7 @@ class DocumentController(Window.Window):
 
     def export_files(self, display_items: typing.Sequence[DisplayItem.DisplayItem]) -> None:
         if len(display_items) > 1:
-            export_dialog = ExportDialog.ExportDialog(self.ui, self, self, display_items)
+            ExportDialog.ExportDialog(self.ui, self, display_items)
         elif len(display_items) == 1:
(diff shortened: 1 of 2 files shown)


Tiomat85:
Sure, I had a quick look through some of it and it looked to be tidying up
variables that weren't needed that hung around from the previous version
and my lack of understanding of some of the structures.  The only
outstanding question would be the extra little issue-ettes you mentioned
that are worth doing post-merge, are they going to be created as their own
individual Issues? That would make the most sense to me as they are not
directly blocking the original issue from being completed, so better to get
that *done* and any follow up work that we discover makes sense as brand
new work items to be done at some point in the great future.
p.s. Trying the email reply option from github so this may not come through
clearly.

cmeyer:
Put smaller issues here:
- https://github.com/nion-software/nionswift/issues/586

Issue #932, opened 2023-01-18
Make export dialog behave similarly for a single window vs. multiple


tylerharvey:
This dialog window is only available for multiple-window export.
![image](https://user-images.githubusercontent.com/13757333/213307544-9887b746-86fa-47c7-ad34-059f3d71b2e5.png)
I can see design reasons for this, but for consistency of use it would be really nice to also have the same export options for single-window export. At the moment, for consistent naming, a user must make sure to always export multiple windows.

bentps:
I want to second this request. In addition to the dialog window options being available, the multi-data-item export puts dashes between the words of the data item title, while exporting a single data item uses spaces.

jamesrussell216:
@KRLango is going to start working on this one now. Just defining a few bits for prosperity...

- In moving the single item export code over to work off the multi item export, as @bentps says that will lead to single items exporting with the multi item export naming convention. Which as far as I can see is probably a good thing but I thought I'd explicitly state it unless anyone can see a problem!
- @KRLango is going to "tidy" the UI up a little as well.

jamesrussell216:
Fixed by #1070

Ion-e:
Blocked by https://github.com/nion-software/nion-internal/issues/136

jamesrussell216:
So I'd argue @Ion-e that #136 is an extension on the original spec for this - a worth while one, but an extension none the less, so I'd further argue if the feature is otherwise working as intended this is completed and we'll add #136 to the to do pile?

cmeyer:
This issue is closed. Can you open a discussion for further discussion of unrelated items?

jamesrussell216:
It is not an unrelated item and it's a simple question - I don't think it needs a discussion item.

Ion-e:
Verified as implemented, thank you!

PR #1070, opened 2024-06-14
Export dialog

KRLango:
This is to solve issue: https://github.com/nion-software/nionswift/issues/932

I have given a single export the same workflow as multiple exports and tidied up the UI for the multiple exports as well:

- Emboldened sections titles for clearer separation
- Allowed the folder path to wrap to a second line to stop the dialog looking unbalanced
- Reordered the different options so the order matched the order in the output filename
- Removed the redundant prefix checkbox


cmeyer:
Thanks! Merged and added an additional commit to make directory text expand to width of dialog if resized:
a399b5999ae3bb5e3816a6d15bf5d83c453b683b

jamesrussell216:
Fixes #932

Issue #749, opened 2021-10-21
Exporting multiple data items should report failures if any occur


cmeyer:
Exporting a batch of files is complicated and may result in failures of individual items due to combinations of data types and file formats. While this might be reported by individual notifications, it would be nice to have a consolidated window showing each exported item in a table, each item's progress, each item's final status (succeeded, failed, maybe specific notes), and each item's final name and location with a button to open its file folder.

Try to export an image and a composite line plot to ndata. The image will export. The composite line plot will silently fail.
```[tasklist]
### Tasks
- [ ] https://github.com/nion-software/nionswift/issues/630
- [ ] https://github.com/nion-software/nionswift/issues/245
```

jamesrussell216:
@cmeyer is this specific to an image and composite line plot or more generally? We've noticed that if you do some combination of exports and file format sometimes you just don't get a file for particular combinations - is that by design if you have invalid combination?

Ion-e:
Seeing that it's not addressed in #1072, Should unvalid formats be disabled/removed in the dropdown selection when there is not an expected valid export to happen? (i.e: PNG export on a 2D ramp).

Ion-e:
#630 hasn't been addressed.

cmeyer:
I removed it from the task list. This issue is closed - file new bugs if you encounter one. Thanks!

Ion-e:
Issue verified as implemented.

Thank you!

PR #1072, opened 2024-06-18
Export results tabulated in popup

KRLango:
This is for issue https://github.com/nion-software/nionswift/issues/749. It presents the results of exports in a popup showing which succeeded and which failed. For failed exports any error message is shown.

An option to open the folder where the exports are saved is also included for quickly getting to the outputs.

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -214,14 +214,21 @@ class ExportDialog(Declarative.Handler):
                         if viewmodel.include_sequence.value:
                             components.append(str(index))
                         filepath = ExportDialog.build_filepath(components, writer.value.extensions[0], directory_path=directory_path)
+
                         file_name = filepath.name
-                        ImportExportManager.ImportExportManager().write_display_item_with_writer(writer.value, display_item, filepath)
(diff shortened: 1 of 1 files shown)


Ion-e:
Bug - Attempting to export an invalid export is reported as "success" in the dialog, however the export doesn't complete, please also note that there are no error messages thrown on the Outputs when the export fails.

Please also note that there no error messages thrown on the Outputs when the export succeeds, is this something we would want? @cmeyer

cmeyer:
Merged with changes.json, squashed, minor edit to commit message 899098e4456dd27aac4692b184274c06b433ba5e

Notes and future work:
- You can get the `UserInterface` `ui` object using `document_controller.ui`. This fact isn't always used consistently in existing code...
- Another note: `document_controller` is basically the window. It is named "controller" due to historical reasons.
- The results dialog should probably be cleaned up to be centered after display (not sure how to do this, may need to measure outputs before-hand and estimate width).
- The results dialog should also probably include a scroll area in case there are 100 output items to report.
- An additional task is to move the "open file in explorer" code to `Utility.py` so it is common. It occurs in 3+ places right now.
- I didn't add new issues for any of the "to-do" style items above.

PR #1077, opened 2024-06-20
File name sanitation

### LINKED ISSUE #591: Exporting data items with forward slashes results in wrong filename

KRLango:
Added a sanitation of filenames to the export.

Fixes #591

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -30,7 +30,6 @@ from nion.utils import Geometry
 from nion.utils import Model
 
 
-
 if typing.TYPE_CHECKING:
     from nion.swift.model import DisplayItem
 
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -19,6 +19,7 @@ import unicodedata
 
 # local libraries
 from nion.swift.model import ImportExportManager
+from nion.swift.model import Utility
 from nion.swift import DocumentController
 from nion.swift import DisplayPanel
 from nion.ui import Declarative
@@ -28,7 +29,7 @@ from nion.utils import Converter
 from nion.utils import Geometry
 from nion.utils import Model
 
-from nionswift.nion.swift.model import Utility
+
 
 if typing.TYPE_CHECKING:
     from nion.swift.model import DisplayItem
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -28,6 +28,8 @@ from nion.utils import Converter
 from nion.utils import Geometry
 from nion.utils import Model
 
+from nionswift.nion.swift.model import Utility
+
 if typing.TYPE_CHECKING:
     from nion.swift.model import DisplayItem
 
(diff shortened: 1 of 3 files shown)


nion/swift/ExportDialog.py
```diff
@@ -28,6 +29,8 @@
 from nion.utils import Geometry
 from nion.utils import Model
 
+
```
cmeyer, line 32: Can you remove this white space change?

nion/swift/model/Utility.py
```diff
@@ -432,6 +433,50 @@ def do_sample() -> None:
     threading.Thread(target=do_sample).start()
 
 
+def simplify_filename(filename: str, replacement_char: str = '_', maximum_length: int = 64) -> str:
```
cmeyer, line 436: I'm questioning the limit of 64 - this seems too short and maybe unnecessary. Is there a reason for this? macOS limit is 255 utf-8 characters. Windows has different restrictions, seemingly 32K when using Unicode calls (see https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry). Linux is more flexible.

nion/swift/model/Utility.py
```diff
@@ -432,6 +433,50 @@ def do_sample() -> None:
...
+    # Windows illegal characters = < > : / \ | ? * " ASCII values 0-31 (non-printable chars)
+    # Windows illegal filenames = CON PRN AUX NUL COM1 COM2 COM3 COM4 COM5 COM6 COM7 COM8 COM9
+    #                             LPT1 LPT2 LPT3 LPT4 LPT5 LPT6 LPT7 LPT8 LPT9
+    # those files names are illegal with or without a suffix, upper and lower case
+    # Windows file names cannot start or end with periods or spaces
+
+    assert maximum_length > 5
+    assert len(replacement_char) == 1
+
+    # since windows is the most restrictive and covers the others, just focus on that
```
cmeyer, line 449: I'm not sure whether I like this idea -- but we could make the restrictions platform specific. The downside is one could save a quirky file on macOS and not be able to load it on Windows - but that may be preferable to strange limitations on macOS for macOS-only users. Any thoughts?

cmeyer (changes requested):
Overall this looks good - my main concern is the 64 char limit. Also a minor whitespace issue.

nion/swift/model/Utility.py
KRLango: That would be the third task in #591

KRLango: Also this is only the file name - it does not include the full path so we need to consider what a sensible maximum might be. That said, it is easy to change the default value once we make a decision.

PR #1078, opened 2024-06-24
2024 06 18 computation processor

### LINKED ISSUE #147: Display changes (slice) should not trigger dependent data computations (pick)
Changing the display slice of should not trigger picks to recompute.

To fix this, computations are going to need fine grained control over their triggers. The slice can legitimately affect the display data; but not the data itself. The pick computation depends on data but since changes to data and display data are grouped together, it has no way to limit itself from recomputing when only the display data changes.

cmeyer:
- **Refactor processors from dict to objects for typing and programmatic access.**
- **Improve typing of processor variables.**
- **Add more typing to processors, and clean init methods.**
- **Store built-in processor instead of descriptions.**
- **Move _processors to computation module (Symbolic).**
- **Add 'type' to computation processor source objects.**
- **Ensure add/remove graphic does not trigger pick recompute.**
- **Remove unused object-specifier function.**
- **Fix #147. Introduce fine grained dependency check to computations.**

### DIFF
--- a/nion/swift/ComputationPanel.py
+++ b/nion/swift/ComputationPanel.py
@@ -55,7 +55,7 @@ _ = gettext.gettext
 class AddVariableCommand(Undo.UndoableCommand):
 
     def __init__(self, document_model: DocumentModel.DocumentModel, computation: Symbolic.Computation,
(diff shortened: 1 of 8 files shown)


cmeyer:
Merged c4a2a3cd15ee184b9d72423cb3d7f3695af29bd6

PR #1082, opened 2024-06-26
Also use Registry for default console startup lines.

Brow71189:
Use the same system that 3rd parties can use to register console startup code also for the default code. This will facilitate alternative console implemetations, because they can then get the same startup code as the builtin one from the registry.


cmeyer:
Merged with minor edits 10498cf033d958ac4153188b5917244aa4d466c1

PR #1083, opened 2024-06-26
Updates to the export results dialog

KRLango:
Following the discussions we had about the export results dialog, I have made a few changes:

- Added a scrollable area for when there are more exports than can fit on the screen

- Only display the dialog if there are some failed exports


cmeyer:
Did somewhat extensive cleanup 140337565295151f4bcbd993e89e1638f52f2329

Issue #1080, opened 2024-06-25
Remove, disable, or fix unmaintained display script feature


cmeyer:
The display script was an experimental feature that allowed a user to use a script to generate a display, particularly for matplotlib. The future feature can be discussed here:
- https://github.com/nion-software/nionswift/issues/140

Ion-e:
@cmeyer is there any other instances or locations where this stool can be activated?

Ion-e:
Awaiting for response.

cmeyer:
FYI - there are any other user accessible places - but I think it's possible via scripting.

Ion-e:
Issue verified as fixed, settings as Dondone

PR #1090, opened 2024-07-08
Remove display script

KRLango:
Fixes https://github.com/nion-software/nionswift/issues/1080

Removed the option from the drop-down in the inspector and ensured if it has previously been selected that the default will be automatically selected


cmeyer:
Thanks! Merged bcbdc6e2a0d49e72227e2574921b06afc4920379

PR #1092, opened 2024-07-10
Fix #1071. Make thumbnails dependent on display item graphic changes.

### LINKED ISSUE #1071: Thumbnails do not update when graphics added/removed
On any 2D image, add a rectangle by using the Graphics menu. Thumbnail doesn't update. Now adjust rectangle - thumbnail updates. Now delete rectangle. Thumbnail doesn't update. Also occurs on line plots with intervals.

See also:
- https://github.com/nion-software/nionswift/issues/251

cmeyer:
See explanation in https://github.com/nion-software/nionswift/pull/1091

### DIFF
--- a/nion/swift/Thumbnails.py
+++ b/nion/swift/Thumbnails.py
@@ -123,7 +123,11 @@ class ThumbnailSource(ReferenceCounting.ReferenceCounted):
                 thumbnail_processor.mark_data_dirty()
                 thumbnail_processor.recompute(ui)
 
+        def graphics_changed(graphic_selection: DisplayItem.GraphicSelection) -> None:
+            thumbnail_changed()
+
         self.__display_changed_event_listener = display_item.display_changed_event.listen(thumbnail_changed)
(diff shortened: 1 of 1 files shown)


KRLango (approved):
This is a much cleaner solution than I was proposing.

My only concern is that the event is fired twice on adding/deleting the graphic: once for adding/removing the graphic in the list of graphics, and once for the selected graphic changing. In this case, I think it is fine, but I always prefer to minimise the number of times an event is fired due to a single action by the user.

cmeyer:
The double notification is also something I'd like to avoid. I have several less-then-optimal solutions for that - for example look at `DataItem.__begin_changes` and `DataItem.__end_changes`. In this case, thumbnails hold off updating immediately so it isn't a big deal; but generally I'd like to have a better solution for this general problem of batching changes. In the thumbnail case, I'd also consider switching from an event based mechanism to a polling mechanism since the updates are rate limited; but that's a different project, too.

PR #1094, opened 2024-07-11
Make inspector sections expanded state persistent.

cmeyer:


Tiomat85:
Just a small thing.
The Bool-To-String converter, it would probably be worth doing a case insensitive convert_back check instead of a case sensitive one. Might especially be useful if the string part is user-facing editable at any point to stop frustration if they type in "True" instead.

cmeyer:
Merged 5db5c1182bbb2cd85bc5d48b57cbbc28cc39e90f

Issue #1093, opened 2024-07-11
Rewrite inspector sections to use declarative


cmeyer:
The existing inspector is difficult to maintain. One way to make it easier is to use the declarative UI instead of the widgets directly. Doing this would allow testing of the model objects. Some of the rewrites will be easier than others.

Ion-e:
task: https://github.com/nion-software/nionswift/pull/1201

cmeyer:
Closing this as complete; computation inspector is separate.

Ion-e:
Verified as implemented.

Great work!

PR #1103, opened 2024-07-19
Switching image display inspector section to declarative UI

KRLango:
Part of https://github.com/nion-software/nionswift/issues/1093

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -1839,6 +1839,31 @@ class ChangeDisplayTypeCommand(Undo.UndoableCommand):
         return isinstance(command, self.__class__) and bool(self.command_id) and self.command_id == command.command_id and self.__display_item_uuid == command.__display_item_uuid
 
 
+class DisplayTypeChooserHandler(Declarative.Handler):
+    def __init__(self, display_item: DisplayItem.DisplayItem, document_controller: DocumentController.DocumentController) -> None:
+        super().__init__()
+        self._document_controller = document_controller
+        self._display_item = display_item
+        self._display_type_items = (_("Default"), _("Line Plot"), _("Image"))
+        self._display_type_flags = (None, "line_plot", "image")
+        self._display_type_reverse_map = {None: 0, "line_plot": 1, "image": 2}
+        self._current_index = self._display_type_reverse_map[self._display_item.display_type]
+
+        u = Declarative.DeclarativeUI()
+
+        self.ui_view = u.create_row(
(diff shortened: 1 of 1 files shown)


nion/swift/Inspector.py
```diff
@@ -1865,7 +1889,6 @@ def change_display_type(item: typing.Tuple[str, typing.Optional[str]]) -> None:
     display_type_row.add_stretch()
     return display_type_row, listener
 
-
```
cmeyer, line 1868: There should be two blank lines before top-level functions and classes.

nion/swift/Inspector.py
```diff
@@ -1839,6 +1839,30 @@ def can_merge(self, command: Undo.UndoableCommand) -> bool:
...
+        )
+
+    def change_display_type(self,  widget: Declarative.UIWidget, current_index: int) -> None:
+        current_display_type = self._display_type_flags[current_index]
+        if self._display_item.display_type != current_display_type:
+            command = ChangeDisplayTypeCommand(self._document_controller.document_model, self._display_item, display_type=current_display_type)
+            command.perform()
+            self._document_controller.push_undo_command(command)
+
 def make_display_type_chooser(document_controller: DocumentController.DocumentController, display_item: DisplayItem.DisplayItem) -> typing.Tuple[UserInterface.BoxWidget, Event.EventListener]:
```
cmeyer, line 1866: There should be two blank lines before top-level functions and classes.

cmeyer (changes requested):
This looks ok - but I'm wondering whether you could replace the other use of `make_display_type_chooser` in `LinePlotDisplayInspectorSection` at the same time? You don't have to fully convert `LinePlotDisplayInspectorSection` - just that individual item. I didn't look closely - is there a reason not to do that right now?

Also made a few whitespace comments - there should be two blank lines before top level classes and functions.

nion/swift/Inspector.py
KRLango: This is done now - I also rebased to avoid having "Whitespace changes" commit messages

nion/swift/Inspector.py
KRLango: This is done now - I also rebased to avoid having "Whitespace changes" commit messages

KRLango:
The reason I didn't change the `LinePlotDisplayInspectorSection` to use the `make_display_type_chooser` was to keep this PR focused on just the 1 section of the inspector. My plan was to do the `LinePlotDisplayInspectorSection` after this PR was accepted using that function and clean up the (then unused) old method at that point.

PR #1105, opened 2024-07-19
Add control to computation edit dialog inputs to control crop rectangles

cmeyer:
Proposed control to allow user to control a computation input source crop rectangle.
Click a cropped input source will uncrop it. Selecting a crop region and clicking
will make that crop region active. Clicking the control when uncropped will look for
the first suitable rectangle crop region or create one. The control is presented as a
small rectangle at the bottom right of the source thumbnail. This is a minimal interface
and we might want to have something more obvious, but this is a first proposal and allows
us to experiment with the functionality.

This is part of a project to provide more computation editor functionality in order to eliminate the custom computation UI in nionswift-experimental.

- **Added DisplayItem accessor for single selected_graphic.**
- **Refactor DataItemThumbnailWidget to use canvas items to draw overlays.**
- **Add cropping control to source thumbnail in computation editor.**

image

### DIFF
--- a/nion/swift/DataItemThumbnailWidget.py
+++ b/nion/swift/DataItemThumbnailWidget.py
@@ -36,7 +36,7 @@ class AbstractThumbnailSource:
     def __init__(self) -> None:
         self.on_thumbnail_data_changed: typing.Optional[typing.Callable[[typing.Optional[_NDArray]], None]] = None
         self.__thumbnail_data: typing.Optional[_NDArray] = None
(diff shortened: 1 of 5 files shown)


jamesrussell216:
Yes I'd echo your comment on perhaps making it a bit more obvious how it works, but none-the-less it works for me under the guise of experimenting with the functionality.

jamesrussell216 (approved):

Tiomat85 (approved):
Nothing jumps out from the code side, so assuming the requirements are as wanted then it looks fine.

cmeyer:
Merged de9941f3a423bc36bc5f0be256dae43c77ea85ee

PR #1100, opened 2024-07-16
changed formatting changes

lisham2000:

PR #1106, opened 2024-07-23
Changed SVG exports to a model architecture with handler and dialog

lisham2000:


cmeyer:
This looks like it supersedes #1100 - if that is the case, can you close that one?

lisham2000:
> This looks like it supersedes #1100 - if that is the case, can you close that one?

Yes happy to. We originally spoke about merging that one first? If that is no longer the case i will go and close it

cmeyer:
I can review each commit separately within this PR if that helps, but I think those commits are pretty close anyway, and they're included here, too.

nion/swift/ExportDialog.py
```diff
@@ -246,66 +250,176 @@ def cancel(self) -> bool:
...
+    UnitType.CENTIMETERS: 37,
+    UnitType.INCHES: 96
+}
 
-        self.int_converter = Converter.IntegerToStringConverter()
 
+class ExportSizeModel(Observable.Observable):
+    def __init__(self, display_item: DisplayItem.DisplayItem) -> None:
+        super().__init__()
+        display_size = self.__determine_display_size(display_item)
```
cmeyer, line 269: Please add units to the function name - it will help prevent errors like the one that is trivially occurring.

`__determine_display_size` -> `__calculate_display_size_in_pixels`

nion/swift/ExportDialog.py
```diff
@@ -246,66 +250,176 @@ def cancel(self) -> bool:
...
+        self.__width: int = display_size.width
+        self.__height: int = display_size.height
+        self.__aspect_ratio = self.__width / self.__height
+        self.__units = UnitType.PIXELS
+
+    def __determine_display_size(self, display_item: DisplayItem.DisplayItem) -> Geometry.IntSize:
+        if display_item.display_data_shape and len(display_item.display_data_shape) == 2:
+            return Geometry.IntSize(height=display_item.display_data_shape[0], width=display_item.display_data_shape[1])
+        return Geometry.IntSize(height=3, width=5)
```
cmeyer, line 278: This is wrong. This function (which should be called `__calculate_display_size_in_pixels`) is returning a _pixel_size_ of 5x3 for line plots.

nion/swift/ExportDialog.py
```diff
@@ -246,66 +250,176 @@ def cancel(self) -> bool:
...
-        main_page = u.create_column(width_row, height_row, spacing=12, margin=12)
+        self._float_to_string_converter = Converter.FloatToStringConverter()
+
+        self.ui_view = u.create_column(
+            u.create_row(
+                u.create_label(text=_("Width:"), width=80),
+                u.create_line_edit(
+                    text="@binding(model.width, converter=_float_to_string_converter)",
+                    on_text_edited="value_updated",
+                    background_color="white",
```
cmeyer, line 337: Comments: 
- Let's not change the background color. That's non-standard UI.
- If we were to change the background color, we could bind it to a value in the model.
- Instead, the model could have two values for each dimension: `width_text` and `width_placeholder_text`. Those would be updated automatically (i.e. trigger notifications) when the user set either the width or height. You might also need to include which value the user has last modified in the model.
- Then you can bind the `text` to `@binding(model.width_text)` and the `placeholder_text` to `@binding(model.placeholder_text)` and not use a converter here (the converter would be used in the model, though).

nion/swift/ExportDialog.py
```diff
@@ -246,66 +250,176 @@ def cancel(self) -> bool:
...
-            return True
-
-        def cancel_clicked() -> bool:
-            return True
-
-        dialog = typing.cast(Dialog.ActionDialog, Declarative.construct(document_controller.ui, document_controller, u.create_modeless_dialog(handler.ui_view, title=_("Export SVG")), handler))
-        dialog.add_button(_("Cancel"), cancel_clicked)
-        dialog.add_button(_("Export"), ok_clicked)
+    def __ok_clicked(self) -> bool:
+        display_shape = Geometry.IntSize(width=int(self.__model.width), height=int(self.__model.height))
```
cmeyer, line 401: This is wrong. You should add `display_size_in_pixels` to the model and use that property here. Instead, you're just grabbing whatever the last field values are, ignoring the units. I would also suggest renaming this local variable `display_shape` to `display_shape_in_pixels` so that it is clear.

nion/swift/ExportDialog.py
```diff
@@ -246,66 +250,176 @@ def cancel(self) -> bool:
...
-        self.height_model = Model.PropertyModel(display_size.height)
+class UnitType(enum.Enum):
+    PIXELS = 0
+    INCHES = 1
+    CENTIMETERS = 2
+
+
+ConversionUnits = {
+    UnitType.PIXELS: 1,
+    UnitType.CENTIMETERS: 37,
```
cmeyer, line 261: These should be floating point values. The conversion to pixels for centimeters results in a ~2% error in the size when using "37" for the pixels per centimeter (which is literally defined several lines above) instead of the more accurate `PIXELS_PER_CM = 37.795275591` value.

nion/swift/ExportDialog.py
```diff
@@ -28,13 +29,16 @@
...
+from nion.utils import Observable
 
 if typing.TYPE_CHECKING:
     from nion.swift.model import DisplayItem
 
 _ = gettext.gettext
 
+PIXELS_PER_INCH = 96.0
+PIXELS_PER_CM = 37.795275591
+
```
cmeyer, line 41: These don't seem to be used anymore. These values should go into the `ConversionUnits` enum.

nion/swift/ExportDialog.py
lisham2000: Changed to in model public function

cmeyer:
I merged this with some minor changes here a8bf1254b8df599ccd35e83aa295ca7c8ec34afd

To fix other issues, I also added these two commits c81ae5808cd1094a46787ae9379006965f3e7634 and b48dda650b2ccb5bb2cc13643b9468c7d2c10a8a

PR #1107, opened 2024-07-24
Moving Collection index inspector section to declarative UI

KRLango:
As part of 1093, moving the collection index inspector section to declarative UI

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -49,7 +49,6 @@ from nion.utils import Observable
 from nion.utils import ReferenceCounting
 from nion.utils import Registry
 from nion.utils import Validator
-from nion.utils.Model import T
 
 if typing.TYPE_CHECKING:
     from nion.swift import Application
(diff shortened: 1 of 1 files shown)


KRLango:
Waiting on https://github.com/nion-software/nionutils/pull/26

cmeyer:
I don't think this use of converters is good - the side effects of accessing a second object are difficult to reconcile.

For this PR, can you move the proposed `nionutils` classes directly into `Inspector.py`? Then we have a single place to replace them with a better solution and don't risk anyone else using that technique.

I'll propose an alternate solution, but it will take a few days.

cmeyer:
Ok - not a few days - we can use a model instead of a converter. It's a very clean solution that doesn't use tuple converters. I put the commit in my repo, look at the latest commit on the link below.

https://github.com/cmeyer/nionswift/commits/review-1107/

KRLango:
@cmeyer my concern with using the model rather than a converter is reusability. I agree that the converter is not the cleanest, but the code you put in the model would have to be rewritten almost verbatim every time we come across a similar situation.

cmeyer:
I haven't had a reason to use a tuple converter yet - so it's not a frequent situation. There might be a few other cases, but it won't be many.

Can you rebase this PR on the latest nionswift and force push to address the mypy issues?

cmeyer:
Follow-up comment: if you're concerned about reusability, we can add a "tuple value model" or something similar. However, before putting it in nionutils, we can try it in nionswift first. And before trying it in nionswift, we can wait for a 2nd occurrence of it to be needed.

KRLango:
You are far more familiar with the codebase than I am so I am happy to concede to your experience that there are no other places you have needed to use this and put my reusability concerns aside unless we do find other places we want to do this again.

cmeyer:
Merged afa45e31b551c1652fa1204399d95e14155f9e39 Thanks!

PR #1108, opened 2024-07-24
1093 sequence inspector section

KRLango:
Moving the sequence inspector section over to declarative UI


cmeyer:
LGTM merged. I forgot to squash, so with extra commit.

Issue #1104, opened 2024-07-19
Splitting/closing panels in workspaces work inconsistently, drawing anomalies


cmeyer:
The drawing anomalies are likely the result of the recent layout algorithm changes.

From @bentps:

> Example 1:
> 
> Start with a blank workspace. Use the toolbar button to divide the display into 2x2 panels. Select one of the smaller panels, and again divide that 2x2. The current behavior is that the sizes of the original 2x2 division get radically changed by subdividing one of the panels. The behavior I'd prefer, and what mostly used to happen, is that a panel that gets divided does not change the size of any surrounding panel.
> 
> Example 2:
> 
> In some cases, adjusting the size of a display panel by clicking and dragging on the panel edge will also affect other nearby panels. My preferred behavior would be that if I am dragging a panel edge, only that panel edge is affected.

### Steps to reproduce layout issue:

Make a 2x2 workspace using toolbar, select bottom right panel, split into 4x4 using toolbar.

<img width="1027" alt="image" src="https://github.com/user-attachments/assets/42754891-23fa-41f3-a10c-edff416bea0e">

### Steps to reproduce drawing anomaly:

Make a 3x3 workspace using toolbar, select bottom right panel, split into 4x4 using toolbar. This will lead to a wacky workspace, but this drawing problem occurs with or without the fix for the wacky workspace. Now start closing the new 4x4 using top-right of bottom-right 3x3 panels. Drawing problem occurs quickly.

<img width="1029" alt="image" src="https://github.com/user-attachments/assets/08451fab-a31d-4f3c-87ba-699487448416">

cmeyer:
Fixed b17966a766aa883ea1944069217d443c98bc8b9b 6b6e892c2e336eb6ede852581379a517590d44fb

Ion-e:
Issue verified as fixed.

Thanks!

PR #1109, opened 2024-07-26
Fix 1104 and only split panels two

cmeyer:
- **Fix split issue by keeping splits when inserting new panel.**
- **Always split the target display panel rather than adding to row/column.**

This addresses #1104.

It also addresses a longstanding request to not make closing a display panel affect other display panels in same column/row, which was the case if you dragged a new item into a display panel such that it extended a row/column and then immediately closed it, in which case it would expand the remaining panels in a non-intuitive way. For instance, before this fix, make a 2x2, then drag an image to the left side of the top-left panel, then close the panel with the image. Resulting layout is not the original 2x2. In new version with this PR, it is. I think this is much more intuitive for arranging layouts, but may have unintended behavior that I haven't thought of yet.

This PR also fixes an issue with the workspace not updating properly when closing panels in some cases (also described in #1104). However, I encountered at least one case where the layout got completely messed up - maybe as the result of undo.

NOTE: requires the latest `nionui` with commits nion-software/nionui@61d936f0c5ecdc448bb44496cb1424b154a16a3f and nion-software/nionui@9ed8105e7d9691e4ec1ba7dd4332b61b7b0a79cf

### DIFF
--- a/nion/swift/Workspace.py
+++ b/nion/swift/Workspace.py
@@ -934,6 +934,7 @@ class Workspace:
         index = container.canvas_items.index(display_panel)
         if isinstance(container, CanvasItem.SplitterCanvasItem):
             # modify the existing splitter
+            splits = list(container.splits)
             old_split = container.splits[index] if not new_splits else 0.0
             new_index_adj = 1 if region == "right" or region == "bottom" else 0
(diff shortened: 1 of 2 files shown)


Ion-e:
Issue verified as fixed, thanks!

PR #1114, opened 2024-07-31
1093 - Moving session inspector section to declarative UI

KRLango:
Moving the session inspector section over to use declarative UI as part of https://github.com/nion-software/nionswift/issues/1093

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -1185,12 +1185,27 @@ class SessionInspectorModel(Observable.Observable):
         super().__init__()
         self.__document_controller = document_controller
         self.__data_item = data_item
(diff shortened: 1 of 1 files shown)


cmeyer:
Before merging this, can you look at `SessionHandler` and `SessionModel` in `DisplayEditPopup.py` and see if you can design this implementation similarly (i.e. without a hard-coded list of fields/properties)? We may be able to have a single common class that is used in both places. The two uses are slightly different in that the inspector updates the values immediately whereas the dialog waits until the end.

cmeyer:
Merged c7545b1b0faca3b1fd491896ff53d7f5fe54449d 9b91eccbe567402d96a42dc4ccc5cfa44031b510

PR #1118, opened 2024-08-05
Switching Info Inspector section to Declarative UI

KRLango:
Switching the Info section of the inspector to declarative UI as part of https://github.com/nion-software/nionswift/issues/1093

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -540,6 +540,23 @@ class DisplayDataChannelPropertyCommandModel(Model.PropertyChangedPropertyModel[
 
 
 class DisplayItemPropertyCommandModel(Model.PropertyChangedPropertyModel[typing.Any]):
+    def __init__(self, document_controller: DocumentController.DocumentController,
+                 display_item: DisplayItem.DisplayItem, property_name: str) -> None:
+        super().__init__(display_item, property_name)
+        self.__display_item = display_item
+        self.__document_controller = document_controller
+        self.__property_name = property_name
+
+    def _set_property_value(self, value: typing.Optional[typing.Any]) -> None:
+        command = ChangeDisplayItemPropertyCommand(self.__document_controller.document_model, self.__display_item, self.__property_name, value)
+        command.perform()
+        self.__document_controller.push_undo_command(command)
+
+    def _get_property_value(self) -> typing.Optional[typing.Any]:
+        return getattr(self.__display_item, self.__property_name)
+
+
+class DisplayItemDisplayPropertyCommandModel(Model.PropertyChangedPropertyModel[typing.Any]):
     """Display item channel property command model.
 
     This model makes undoable changes to a display item property.
@@ -592,13 +609,13 @@ class InfoInspectorHandler(Declarative.Handler):
     def __init__(self, document_controller: DocumentController.DocumentController, display_item: DisplayItem.DisplayItem):
         super().__init__()
         self._display_item = display_item
-        self._title_model = Model.PropertyChangedPropertyModel[str](display_item, "title")
-        self._placeholder_title_model = Model.PropertyChangedPropertyModel[str](display_item, "placeholder_title")
+        self._title_model = DisplayItemPropertyCommandModel(document_controller, display_item, "title")
+        self._placeholder_title_model = DisplayItemPropertyCommandModel(document_controller, display_item, "placeholder_title")
         self._caption_model = DisplayItemPropertyCommandModel(document_controller, display_item, "caption")
         self._editable_caption_model = Model.PropertyModel[str](display_item.caption)
         self._caption_current_index = Model.PropertyModel[int](0)
         self._session_id_model = Model.PropertyChangedPropertyModel[str](display_item, "session_id")
-        self._created_local_as_string_model = Model.PropertyChangedPropertyModel[str](display_item, "created_local_as_string")
+        self._created_local_as_string_model = DisplayItemPropertyCommandModel(document_controller, display_item, "created_local_as_string")
         self.info_title_label: typing.Optional[UserInterface.Widget] = None
 
         u = Declarative.DeclarativeUI()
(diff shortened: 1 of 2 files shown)


nion/swift/Inspector.py
```diff
@@ -588,109 +588,97 @@ def __property_changed_from_graphic(self, name: str) -> None:
...
 
 
+class InfoInspectorHandler(Declarative.Handler):
+    def __init__(self, document_controller: DocumentController.DocumentController, display_item: DisplayItem.DisplayItem):
+        super().__init__()
+        self._display_item = display_item
+        self._title_model = Model.PropertyChangedPropertyModel[str](display_item, "title")
+        self._placeholder_title_model = Model.PropertyChangedPropertyModel[str](display_item, "placeholder_title")
+        self._caption_model = Model.PropertyChangedPropertyModel[str](display_item, "caption")
+        self._editable_caption_model = Model.PropertyModel[str](display_item.caption)
```
cmeyer, line 598: To make editing undo-able, `_title_model` and either `_caption_model` or `_editable_caption_model` can use `DisplayItemPropertyCommandModel`. Also see note below.

nion/swift/Inspector.py
```diff
@@ -588,109 +588,97 @@ def __property_changed_from_graphic(self, name: str) -> None:
...
+            spacing=4
+        )
+
+    def _begin_caption_edit(self, widget: UserInterface.Widget) -> None:
+        self._editable_caption_model.value = self._display_item.caption
+        self._caption_current_index.value = 1
+
+    def _save_caption_edit(self, widget: UserInterface.Widget) -> None:
+        new_caption = self._editable_caption_model.value
+        self._display_item.caption = new_caption if new_caption is not None else str()
```
cmeyer, line 657: `self._display_item.caption = ...`: should go through `DisplayItemPropertyCommandModel` to make it undoable.

Additional comment: with these changes, I'm ok with this code; but there might be a more streamlined way to do the editing and updating rather than using two text area widgets. After this gets merged, we might want to revisit it.

cmeyer (changes requested):

nion/swift/Inspector.py
KRLango: I have done the _caption_model, but the other won't work:
_title_model - I get a conflict making a DisplayPanel.ChangeDisplayCommand because it already has a (different) title input giving me the error "nion.swift.DisplayPanel.ChangeDisplayCommand() got multiple values for keyword argument 'title'"
_editable_caption_model - I was trying to maintain the existing behaviour where it does not get pushed to the display model until the save button is pressed - this would change that.

nion/swift/Inspector.py
KRLango: Changing the caption directly was an oversight on my part and is not fixed.

I had a similar thought about streamlining it while I was working on it, but decided to maintain the current process for now.

nion/swift/Inspector.py
```diff
@@ -594,7 +594,7 @@ def __init__(self, document_controller: DocumentController.DocumentController, d
         self._display_item = display_item
         self._title_model = Model.PropertyChangedPropertyModel[str](display_item, "title")
         self._placeholder_title_model = Model.PropertyChangedPropertyModel[str](display_item, "placeholder_title")
```
cmeyer, line 596: `self._title_model` needs to be changed too.

cmeyer (changes requested):
`self._title_model` should be a `DisplayItemPropertyCommandModel` too.

nion/swift/Inspector.py
KRLango: @cmeyer I am guessing you missed my earlier comment: I tried doing this, but it failed: I got a conflict making a DisplayPanel.ChangeDisplayCommand because it already has a (different) title input giving me the error "nion.swift.DisplayPanel.ChangeDisplayCommand() got multiple values for keyword argument 'title'" Since the property/value pairs are passed into kwargs, it detected this as a second attempt to set the title, and I am not sure of the best way to resolve this.

nion/swift/Inspector.py
cmeyer: Yes I missed this comment... looking at it now.

cmeyer: Ok here's what's going on... it not pretty. There are two types of "properties" on the display item. One is the direct properties - things like the `title` and `caption`. The other properties are display type specific and include things like `y_min` and `left_channel` for line plots, which are stored within a sub-structure of the display item named "display properties".  Setting aside the design of this (needs work!), within this context, the `DisplayItemPropertyCommandModel` is mis-named and should be renamed to`DisplayItemDisplayPropertyCommandModel` (also update in `LinePlotDisplaySectionHandler`). Crucially, that class uses `DisplayPanel.ChangeDisplayCommand` to call `set_display_property` on the display item in an undoable way. Then you'll need to add a new class named `DisplayItemCommandModel` which uses `ChangeDisplayItemPropertyCommand` to set a property directly on the display item. `title` and `caption` should use the _new_ `DisplayItemCommandModel`.

nion/swift/Inspector.py
cmeyer: Yes, missed it! - see my explanation above.

nion/swift/Inspector.py
```diff
@@ -540,6 +540,23 @@ def _set_property_value(self, value: typing.Optional[typing.Any]) -> None:
...
+                 display_item: DisplayItem.DisplayItem, property_name: str) -> None:
+        super().__init__(display_item, property_name)
+        self.__display_item = display_item
+        self.__document_controller = document_controller
+        self.__property_name = property_name
+
+    def _set_property_value(self, value: typing.Optional[typing.Any]) -> None:
+        command = ChangeDisplayItemPropertyCommand(self.__document_controller.document_model, self.__display_item, self.__property_name, value)
+        command.perform()
+        self.__document_controller.push_undo_command(command)
```
cmeyer, line 553: I think this needs `if value != self._get_property_value():` to check if the value changed.

First, to describe how I discovered this: editing the caption isn't undo-able with a single Undo without it. It takes two undo's. With this change, it is undo/redo-able.

Putting a `import traceback; traceback.print_stack()` here showed that the value is being set after editing, but then the 'text edited' is received from Qt and then it gets set again. This is a common problem with UI bindings, and my typical workaround is to check if the value changed. I'm not sure whether the equality check should go into `_set_property_value` or done by its caller (in `PropertyChangedPropertyModel`) - let's leave that as an open question for now.

I think something similar needs to go wherever `_set_property_value` is overridden in this file, btw.

cmeyer (changes requested):
Noticed another subtle undo bug...

cmeyer:
Merged here 84ccf5f24ede847b8f2d357400ecd4a4378a5256

Note: removed spurious white space change in DisplayItem.py

Issue #1028, opened 2024-04-15
Improve display panel title bar


cmeyer:
- [ ] make it match panel platform style
- [ ] macos: filled 'x' button on the left
- [ ] windows: title aligned left
- [x] add a distinct 'edit' or equivalent icon (pencil?) button to help users find title/info editing
- [x] consider adding size/storage info when available layout space allows it
- [x] "(4)(512,512)(1024,) [1024MB]"

jamesrussell216:
@cmeyer quick question on this one - on your last two points why is it important to have that info in the title? Every piece of additional info that goes in there dilutes what I'd argue is the important information - the actual title. Surely having that info in the Inspector is sufficient?

cmeyer:
It already shows up as a tool-tip - but I think having the "space used" would be helpful.

cmeyer:
Something like "45MB" meaning 45MB used on disk.

If we add compression storage, we might want something like "45MB disk/90MB memory" meaning "45MB storage on disk" but "90MB when loaded in memory". I know that could be confusing, so I'm open to other ideas, but both are valid/useful pieces of information. Note: the memory usage is difficult to quantify since HDF5 files are memory mapped and don't necessarily use RAM.

An alternative to the memory memory is to have something like Firefox and Chrome have, which is a window that shows resources being used, i.e. if memory is being used.

In any case, this sounds like a different issue that shouldn't go into the title bar. I've edited the top comment to remove that part of the request.

cmeyer:
I had experimented with this, but seemingly didn't keep anything other than this minimal code snippet to experiment with:

```
diff --cc nion/swift/Panel.py
index a8aa52c3,a8aa52c3..9c50c3b3
--- a/nion/swift/Panel.py
+++ b/nion/swift/Panel.py
@@@ -440,7 -440,7 +440,7 @@@ class HeaderCanvasItem(CanvasItem.Canva
                  drawing_context.text_align = 'center'
                  drawing_context.text_baseline = 'bottom'
                  drawing_context.fill_style = '#000'
--                drawing_context.fill_text(self.title, canvas_size.width // 2, canvas_size.height - self.__text_offset)
++                drawing_context.fill_text(self.title + " \N{LOWER RIGHT PENCIL}", canvas_size.width // 2, canvas_size.height - self.__text_offset)
```

jamesrussell216:
Notes:

@Tiomat85 has put in a PR for the edit functionality so hopefully we can tick that off.

[...]

PR #1119, opened 2024-08-05
Refactor Control Rendering Code in HeaderCanvasItem

Tiomat85:
re Issue #1028
Added in a edit control, just to the right of the rendered title. Created local rectangles saving the location of the close and edit titles, and refactored the click handler to check the rectangles rather than absolute values. Broke out the rendering code into sections to allow easier refactoring for different display modes for platforms. Currently all platforms still use the windows rendering mode.

### DIFF
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -1901,9 +1901,10 @@ class DisplayPanel(CanvasItem.LayerCanvasItem):
         self.__content_canvas_item.on_context_menu_event = self.__handle_context_menu_event
 
         self.__header_canvas_item = Panel.HeaderCanvasItem(DisplayPanelUISettings(document_controller.ui),
-            display_close_control=True,
-            display_edit_control=True,
-            get_font_metrics_fn=typing.cast(typing.Callable[[str, str], UISettings.FontMetrics], self.ui.get_font_metrics))
+                                        display_close_control=True,
+                                        display_edit_control=True,
+                                        get_font_metrics_fn=typing.cast(typing.Callable[[str, str],
+                                        UISettings.FontMetrics], self.ui.get_font_metrics))
 
         def header_double_clicked(x: int, y: int, modifiers: UserInterface.KeyboardModifiers) -> bool:
             action_context = document_controller._get_action_context()
@@ -2009,7 +2010,6 @@ class DisplayPanel(CanvasItem.LayerCanvasItem):
             action_context.display_panel = self
             action_context.display_item = self.display_item
             document_controller.perform_action_in_context("window.open_title_edit", action_context)
-
 
         self.__header_canvas_item.on_select_pressed = self._select
         self.__header_canvas_item.on_drag_pressed = self.__handle_begin_drag
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -1900,11 +1900,15 @@ class DisplayPanel(CanvasItem.LayerCanvasItem):
         self.__content_canvas_item.on_focus_changed = self.set_focused
         self.__content_canvas_item.on_context_menu_event = self.__handle_context_menu_event
 
-        self.__header_canvas_item = Panel.HeaderCanvasItem(DisplayPanelUISettings(document_controller.ui), display_close_control=True)
+        self.__header_canvas_item = Panel.HeaderCanvasItem(DisplayPanelUISettings(document_controller.ui),
+            display_close_control=True,
+            display_edit_control=True,
+            get_font_metrics_fn=typing.cast(typing.Callable[[str, str], UISettings.FontMetrics], self.ui.get_font_metrics))
 
         def header_double_clicked(x: int, y: int, modifiers: UserInterface.KeyboardModifiers) -> bool:
             action_context = document_controller._get_action_context()
             action_context.display_panel = self
+            action_context.display_item = self.display_item
             document_controller.perform_action_in_context("window.open_title_edit", action_context)
             return True
 
@@ -2000,9 +2004,17 @@ class DisplayPanel(CanvasItem.LayerCanvasItem):
                 command = workspace_controller.remove_display_panel(self)
                 document_controller.push_undo_command(command)
 
+        def edit() -> None:
+            action_context = document_controller._get_action_context()
+            action_context.display_panel = self
+            action_context.display_item = self.display_item
+            document_controller.perform_action_in_context("window.open_title_edit", action_context)
+
+
         self.__header_canvas_item.on_select_pressed = self._select
         self.__header_canvas_item.on_drag_pressed = self.__handle_begin_drag
         self.__header_canvas_item.on_close_clicked = close
+        self.__header_canvas_item.on_edit_clicked = edit
 
         ui = document_controller.ui
 
--- a/nion/swift/Panel.py
+++ b/nion/swift/Panel.py
@@ -6,6 +6,7 @@ import dataclasses
 import functools
 import gettext
 import logging
+import platform
 import sys
 import threading
 import typing
@@ -338,11 +339,14 @@ class HeaderCanvasItem(CanvasItem.AbstractCanvasItem):
     # header_height = 20 if sys.platform == "win32" else 22
 
     def __init__(self, ui_settings: UISettings.UISettings, title: typing.Optional[str] = None,
-                 label: typing.Optional[str] = None, display_close_control: bool = False) -> None:
+                 label: typing.Optional[str] = None, display_close_control: bool = False,
+                 display_edit_control: bool = False,
+                 get_font_metrics_fn: typing.Optional[typing.Callable[[str, str], UISettings.FontMetrics]] = None) -> None:
         super().__init__()
         self.wants_mouse_events = True
         self.__title = title if title else ""
         self.__display_close_control = display_close_control
+        self.__display_edit_control = display_edit_control
         self.__ui_settings = ui_settings
         self.__set_default_style()
         self.update_sizing(self.sizing.with_fixed_height(self.header_height))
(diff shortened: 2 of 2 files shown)


nion/swift/DisplayPanel.py
```diff
@@ -1868,9 +1872,17 @@ def close() -> None:
                 command = workspace_controller.remove_display_panel(self)
                 document_controller.push_undo_command(command)
 
+        def edit() -> None:
+            action_context = document_controller._get_action_context()
+            action_context.display_panel = self
+            action_context.display_item = self.display_item
+            document_controller.perform_action_in_context("window.open_title_edit", action_context)
+
+
```
cmeyer, line 1881: Should only be a single blank line here. Please consider using `flake8 --ignore=E501 nion/swift/Panel.py nion/swift/DisplayPanel.py` and addressing any issues specific to this PR. You may need to install `conda install flake8` in your environment.

nion/swift/Panel.py
```diff
@@ -376,6 +395,108 @@ def context_menu_event(self, x: int, y: int, gx: int, gy: int) -> bool:
             return self.on_context_menu_clicked(x, y, gx, gy)
         return False
 
+
```
cmeyer, line 398: Should only be a single blank line here. Please consider using `flake8 --ignore=E501 nion/swift/Panel.py nion/swift/DisplayPanel.py` and addressing any issues specific to this PR. You may need to install `conda install flake8` in your environment.

nion/swift/Panel.py
```diff
@@ -376,6 +395,108 @@ def context_menu_event(self, x: int, y: int, gx: int, gy: int) -> bool:
...
+        self.__close_control_rect = self._draw_close_control_windows(drawing_context)
+        title_rect = self._draw_title_text(drawing_context)
+        if title_rect and title_rect.width > 0:
+            self.__edit_control_rect = self._draw_edit_button(drawing_context, title_rect)
+        else:
+            self.__edit_control_rect = Geometry.IntRect.from_tlhw(0, 0, 0, 0)
+
+    def _draw_close_control_mac(self, drawing_context: DrawingContext.DrawingContext) -> None:
+        self._draw_close_control_windows(drawing_context)
+    def _draw_close_control_linux(self, drawing_context: DrawingContext.DrawingContext) -> None:
```
cmeyer, line 409: Should be a single blank line here. Please consider using `flake8 --ignore=E501 nion/swift/Panel.py nion/swift/DisplayPanel.py` and addressing any issues specific to this PR. You may need to install `conda install flake8` in your environment.

nion/swift/Panel.py
```diff
@@ -376,6 +395,108 @@ def context_menu_event(self, x: int, y: int, gx: int, gy: int) -> bool:
...
+                    (canvas_size.width // 2) - (font_metrics.width // 2),
+                    font_metrics.height,
+                    font_metrics.width)
+            else:
+                 # Without knowing how big the text is, need to do a clunky fallback
+                 # Close button and pencil is roughly 40 pixels wide
+                title_rect = Geometry.IntRect.from_tlhw(0, 40, canvas_size.height, canvas_size.width - 80)
+        return title_rect
+
+
```
cmeyer, line 471: Should only be a single blank line here. Please consider using `flake8 --ignore=E501 nion/swift/Panel.py nion/swift/DisplayPanel.py` and addressing any issues specific to this PR. You may need to install `conda install flake8` in your environment.

nion/swift/Panel.py
```diff
@@ -376,6 +395,108 @@ def context_menu_event(self, x: int, y: int, gx: int, gy: int) -> bool:
...
+                drawing_context.move_to(close_box_left + 4, close_box_bottom - 1)
+                drawing_context.line_to(close_box_left, close_box_bottom - 3)
+                drawing_context.line_width = 1
+                drawing_context.line_cap = "round"
+                drawing_context.stroke_style = self.__control_style
+                drawing_context.stroke()
+            control_rect = Geometry.IntRect.from_tlhw(close_box_top - control_margin, close_box_left - control_margin, control_width + 2 * control_margin, control_width + 2 * control_margin)
+            self.__edit_control_rect = control_rect
+        return control_rect
     def _repaint(self, drawing_context: DrawingContext.DrawingContext) -> None:
```
cmeyer, line 500: Should be a single blank line here. Please consider using `flake8 --ignore=E501 nion/swift/Panel.py nion/swift/DisplayPanel.py` and addressing any issues specific to this PR. You may need to install `conda install flake8` in your environment.

nion/swift/Panel.py
```diff
@@ -420,27 +541,20 @@ def _repaint(self, drawing_context: DrawingContext.DrawingContext) -> None:
...
+                #        self._draw_close_control_mac(drawing_context)
+                #    case "Linux":
+                #        self._draw_close_control_linux(drawing_context)
+                #    case _: #  Default to Windows
+                #        self._draw_controls_windows(drawing_context)
+
+
+
+
+
```
cmeyer, line 556: The `match` is only available in Python 3.10, which is why `mypy` is failing with the cryptic error message. Just use an `if...elif...` construct until we officially deprecate Python 3.9 in the CI.

cmeyer (changes requested):
Overall looks good - some formatting changes (and there might be more, I didn't look carefully). Consider using `flake8 --ignore=E501 nion/swift/Panel.py nion/swift/DisplayPanel.py` and addressing any issues specific to this PR. You may need to install `conda install flake8` in your environment.

The `match` statement in Python 3.10 or later and was causing the `mypy` issue. Can you rewrite using `if...elif...` for now?

nion/swift/Panel.py
Tiomat85: Ah so it was that causing the weird mypy error, and not the package.  I didn't realise that match was added in recently, but sure I can fix that to prep for adding in per platform styling.  It also explains some of the other areas of code that had massive nested if/else statements instead of a single match.
As you say, maybe this is a good reason to consider deprecating the 3.9 build to tidy up those.

nion/swift/Panel.py
```diff
@@ -420,27 +541,12 @@ def _repaint(self, drawing_context: DrawingContext.DrawingContext) -> None:
...
-                drawing_context.text_align = 'center'
-                drawing_context.text_baseline = 'bottom'
-                drawing_context.fill_style = '#000'
-                drawing_context.fill_text(self.title, canvas_size.width // 2, canvas_size.height - self.__text_offset)
+                if platform.system() == "Darwin":  # Mac
+                    self._draw_controls_mac(drawing_context)
+                elif platform.system() == "Linux":
+                    self._draw_controls_linux(drawing_context)
+                else:
+                    self._draw_controls_windows(drawing_context)
```
cmeyer, line 549: These all fall back to drawing `_draw_controls_windows` - is there any reason to make them different for this particular PR? We could add platform specific code in the next PR. I'm ok leaving this, but just trying to clean up unnecessary changes. What do you think?

Tiomat85: You're correct they all default currently to the same (windows) rendering mode. I unrolled this to do the If/else bit as part of the tidying up bit.  I don't mind removing it, but the plan was to try and work towards the other half of the Issue of having the more platform based titles.  So it is a structural part of that, with each platform going down its own rendering path of placing the controls but re-using the grouping code (edit to the right of the title etc).

cmeyer: Do you plan to do that for this PR or do you want me to merge already?

I'm not too picky on this, but generally I think it's better practice to make each commit stand on its own without anything extraneous. This 'commit' (after we squash the 3 commits into a single commit) adds the pencil icon. The additional platform stuff is extraneous with respect to that change, so I would generally leave it out. The follow-up commit/PR will be something like "add platform specific variants for headers" and would include the platform specific changes.

Tiomat85: I've pulled the 'styling' parts out as I think you are right about not leaving the placeholders in there if we aren't working on them right now, and the more I have thought about it the more there are probably a lot more questionmarks and open ended questions on how styling wants to be done that are more far reaching than this relatively small Issue/PR.

nion/swift/Panel.py
```diff
@@ -248,13 +251,26 @@ def __init__(self, ui_settings: UISettings.UISettings, title: typing.Optional[st
...
+        self.__close_control_rect: typing.Optional[Geometry.IntRect] = None
+        self.__get_font_metrics: typing.Optional[typing.Callable[[str, str], UISettings.FontMetrics]] = None
+
+        if get_font_metrics_fn:
+            self.__get_font_metrics = get_font_metrics_fn
+        elif self.__ui_settings:
+            if hasattr(self.__ui_settings, "ui"):
+                local_ui = self.__ui_settings.ui
+                if hasattr(local_ui, "get_font_metrics"):
+                    self.__get_font_metrics = local_ui.get_font_metrics
```
cmeyer, line 267: I didn't see this on first pass of review - but this is again breaking encapsulation. In this case, let's just change the HeaderCanvasItem to _not_ take the `get_font_metrics_fn` argument. It's redundant since `UISettings` is required and supplies the same info. Also, we can eliminate the `label` argument too. Unused. Also, the `HeaderCanvasItem` is only used in two places and only one of those two places `FilterPanel.py` requires eliminating the `get_font_metrics_fn` argumnet. Neither of those places passes the `label` argument, so that can just be removed I think.

cmeyer (changes requested):
Sorry, one more change request that I missed upon the first pass of the review...

jamesrussell216:
Note (more for my own reference!): This is paused pending drawing code changes.

nion/swift/Panel.py
cmeyer: You can use `self.__ui_settings.get_font_metrics(...)`. No need to pass in `get_font_metrics_fn` anymore.

cmeyer (changes requested):
The new canvas item rendering code will never call `_repaint` on the canvas item - so if we keep this approach (see below) it would need its own composer similar to the other canvas items. All of the methods from `_repaint` like `_draw_controls` will need to go into a composer.

However, I think a better approach here is to use existing canvas items and canvas item layout to construct the header. Using this approach the header canvas item would be a `CanvasItemComposition` containing text, buttons, etc. The header canvas item would be responsible for responding to changes by setting the appropriate values on its constituent items. This will also minimize updates as the new drawing code is very efficient.

If you're amenable to that approach, you can take a pass at it; or if you want me to do it I can do it in about 30 minutes since I've been working heavily on the canvas item code over the last few months. In that case, close this PR and let me know.

If you work on it, you might find the [canvas item example code](https://github.com/nion-software/nionui/blob/4cab31a9cc23d1349b1bd96c805dd5cc2e3ee4c6/nionui_app/nionui_examples/canvas_demo/main.py#L153-L167) helpful to see how layout works. Then you'll need to have each canvas item properly handle its state, update itself, and provide a immutable composer for drawing that state.

cmeyer:
I'd like to close this in favor of #1328, which I think addresses most of the concerns of #1028 and #1280.
- #1328 
- #1028 
- #1280

Tiomat85:
Closing as superceded by #1328

PR #1130, opened 2024-08-09
Fix #983. Bypass panel multi-select when only shift and tool utilizes shift.

### LINKED ISSUE #983: Using shift for multi-select panels conflicts with rectangle graphic square constraint key
From email

> If I select window A, and then want to take an FFT of a region in window B.
> I choose rectangle tool, draw a rectangle in window B-holding down shift to make the FFT have isotropic pixels. Then do ctrl-F. Nothing will happen because when I drew the square I needed the hotkey shift, but the shift hotkey is also overloaded as multiselect.
> Note that the click in B is not shift click but shift unclick.
> 
> Correct behavior should be that multiselect of windows should happen when the second window is chosen by shift click, but NOT if a tool is used on the window that takes a shift modifier and the activation of the window is a secondary effect of using the tool.
>

cmeyer:


Tiomat85 (commented):
It appears to fix the initial problem with the shift key and Rectangle triggering both actions.  It certainly seems better, but there look to have been a few extra niggles which could either end up with other Issues or just a weakness going forwards.
You can hold Ctrl with creating the rectangle, and it still selects after drawing the rectangle.  That is less of a problem as it doesn't have a direct use, but it can still lead to strange UX behaviour where using a control can gain side-effects.

cmeyer:
If there are other problems, let's file issues. I like this PR solution because it is isolated to nionswift and fixes the problem. But I'd also like to either approve this PR and move on or reject it with a valid reason and a better solution.

PR #1138, opened 2024-08-16
Size restraints svg

lisham2000:
Height and width restraints between 3/12 inches for SVG


lisham2000:
To solve issue #1124

cmeyer (changes requested):
Thanks for this PR - it's close, but I think we don't want to limit the user if they explicitly change it - we only want to limit the initial values that the dialog shows.

As per original issue text:

> Algorithm for ***initial*** size should be something like:

Also, if reworking the commit, can you use [imperative mood](https://en.wikipedia.org/wiki/Imperative_mood) for the commit message?

"Add initial height/width to SVG dialog."

Issue #111, opened 2018-02-28
Provide easier way to discover keyboard shortcuts


cmeyer:
Provide an easier way to discover keyboard shortcuts, including the "hidden" shortcuts not visible in menus.

Many standard menu item shortcuts are discoverable by clicking through the menus.

The tool tips on the tool bar provide more discoverable shortcuts.

Other shortcuts are embedded in code, e.g., option key while dragging graphics will adjust from center (missing on line graphic, btw). More shortcuts appear in the instrumentation kit. Search code for these key overloads  by searching for "modifiers.*".

However, it would be nice to have a dialog which lists all of the shortcuts, grouped into categories. See [comment below](https://github.com/nion-software/nionswift/issues/111#issuecomment-2293945010) for some ideas about the categories and how to present them).

We will eventually want to provide an ability to edit keyboard shortcuts and perhaps execute commands from this window, so design accordingly. See
- https://github.com/nion-software/nionswift/issues/298

The code for the dialog could be based loosely on the "Recent Changes" tab in the About Box dialog code.

Remember that "Ctrl+" is "Cmd+" on macOS; and "Alt+" is usually "Option+" on macOS. Consider using Unicode symbols ⌥ and ⌘ symbols on macOS in dialog.

### Tasks
- [ ] Propose list of shortcut groups as a comment in this issue
- [ ] Propose a design for the dialog. Dialog title should be "Keyboard Shortcuts".
- [ ] Implement proposed and accepted dialog
- [ ] Add "Keyboard Shortcuts" menu item (after the "Preferences..." (macOS) or "Settings..." (Windows) menu item).

cmeyer:
See this link for organization categories; consider making dialog like this in anticipation of eventual shortcut editor:

Organizing into groups https://helpx.adobe.com/illustrator/using/default-keyboard-shortcuts.html

Possible dialog ideas in anticipation of eventual editing capabilities (takeaway: organize into groups with combo box selecting which group is displayed):

<img width="453" alt="image" src="https://github.com/user-attachments/assets/0462e621-aa3f-4e9f-8e56-b854e8be6ca7">

cmeyer:
From Niklas December 21 2017:

> 2 keyboard shortcuts are great, if you can find them. I knew that the computation window has moved to a keyboard shortcut, it took me 15 attempts to come up with cntl-r for computation. Right clicking on a window should give access to a list of options.

lisham2000:
[...]

PR #1139, opened 2024-08-16
Changed Menu Json to include keybind menu item

lisham2000:
The Json changes to add keybind menu item

### DIFF
--- a/nion/swift/resources/menu_config.json
+++ b/nion/swift/resources/menu_config.json
@@ -759,7 +759,11 @@
             {
                 "type": "item",
                 "action_id": "application.about"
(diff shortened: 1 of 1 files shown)


lisham2000:
Linked directly with PR #80 in nionui but is a solution to #111

cmeyer:
Rejecting this, too. See comments in nion-software/nionui#80

In addition to those comments, even if this were to be accepted, I would request that the tabs be replaced with spaces. A diff in GitHub shows that this change isn't following existing formatting:

https://github.com/nion-software/nionswift/pull/1139/files

PR #1141, opened 2024-08-19
1093 - graphics inspector section

KRLango:
Change the graphics section of the inspector to use the declarative UI


cmeyer:
I changed this to draft - it looks like it needs to be cleaned up (merge conflicts) + code like this:

```
if isinstance(value, str):
    print("blah!")
```

Is it possible to split this commit into multiple PR's that are easier to review, like you did for the the point inspector? f3e67d85c153931c8801c2873df0d470d32145ba

cmeyer:
Sorry this is taking a while - lots of code to review.

While I'm doing that, this seems like a good one for @Ion-e to look through carefully. There are some obvious issues... e.g. line profile doesn't show the correct inspector and several text mis-alignments which were there prior to this PR - but it would be nice to (a) not make them worse; (b) fix the old problems here too where possible.  There are probably other issues to uncover in testing.

My suggested approach to testing is to make an image with non-mask graphics, take screenshot of old/new inspector with each graphic type to compare visual display, then use functionality of each to make sure it works. If a bug is found in the old version but not in new version, file an issue. If a bug is found in the new version but not old, fix in PR (and make a comment here). Do the same for each mask type and for 1D line plots with interval and channel graphics. Be aware of #134 during testing (no lattice inspector was ever made).

**Example alignment issue** (old on top, new on bottom). See "Rotation" is cut off in new. Also alignment on lock/position/shape/center-button is not great in either old/new, but worse in new due to the "center graphic" button style change. I think you might be able to use the (recent) text alignment capabilities to improve the alignment, although your mileage may vary! Search for "text_alignment_vertical" for some example code. Maybe align everything in row to top/bottom/center (not sure which will work the best). I'm aware that to truly fix things may require updates to `Declarative.py` and maybe even the `nionui-tool` / Qt code. Let's do the best we can on the Python side before skipping to the Qt changes.

image

vs

image

cmeyer:
Is this ready for review?

cmeyer (changes requested):
Sorry this is going to need another merge commit after I merged the color picker and text button.

cmeyer:
Merged here: e94c8b8d43bd77a99feba95c3f3504677dc69aee

Also see these two follow-up commits that I merged without review
- 7a9b518585c49d9fb3251e0d8ca7d9f28c673c00 (remove unused functions)
- a64adc80c114b656eac60f4e9cf4e419c77d3d0d (fix graphic label undo)

PR #1144, opened 2024-08-20
Add persistence for unit values in SVG

lisham2000:


cmeyer (changes requested):

nion/swift/ExportDialog.py
```diff
@@ -261,18 +261,18 @@ class UnitType(enum.Enum):
 
 
 class ExportSizeModel(Observable.Observable):
-    def __init__(self, display_item: DisplayItem.DisplayItem, ui: UserInterface.UserInterface) -> None:
+    def __init__(self, display_item: DisplayItem.DisplayItem, units_persistent_model: UserInterface.StringPersistentModel) -> None:
```
cmeyer, line 264: In this case, you can just call it `units_model: Model.PropertyModel[str]` instead of the more specific `units_persistent_model: UserInterface.StringPersistentModel`. When receiving parameters, it's better to take the more general case instead of requiring the more specific. It makes testing easier. You could also consider adding a test here to avoid the bug that you'll discover when merging the latest changes on `master` to this branch.

nion/swift/ExportDialog.py
```diff
@@ -261,18 +261,18 @@ class UnitType(enum.Enum):
...
         super().__init__()
-        self.ui = ui
+        self.__units_persistent_model = units_persistent_model
         display_size = self.__calculate_display_size_in_pixels(display_item)
         self.__width = display_size.width
         self.__height = display_size.height
         self.__aspect_ratio = self.__width / self.__height
         self.__float_to_string_converter = Converter.FloatToStringConverter()
         self.__primary_field = 'width'  # Primary field to determine which text is calculater
         # Load the persisted units value, default to PIXELS if not found
```
cmeyer, line 273: `# Load the persisted units value, default to PIXELS if not found`

This comment is outdated and can be removed. The default value is set by the caller.

nion/swift/ExportDialog.py
```diff
@@ -261,18 +261,18 @@ class UnitType(enum.Enum):
...
-        persisted_units = self.ui.get_persistent_string("export_units", UnitType.PIXELS.name)
-        self.__units = UnitType[persisted_units]
+        persisted_units = self.__units_persistent_model.value
+        self.__units = UnitType[str(persisted_units)]
```
cmeyer, line 275: A couple of notes here (minor points, depending on how you look at it):
- if you were to keep this name, can you use `persistent_units` rather than the past tense `persisted`
- instead of making the assumption that these are persisted, it would be better to just assume the value is coming from the model, which is a string, and so should be named `units_str`

A more difficult problem that is slightly trickier: we may want to expand the list of units in the future. If we do that and then try to load a file in this version, it may have persistent units that aren't recognized in this version and will throw an exception on line 275 above. 

The way I usually like to deal with this is by creating a separate map and reverse map for converting between the enum and the persistent value. Something along these lines (I didn't run this, but it's close):

```
unit_to_str_map = { UnitType.PIXELS: "pixels", UnitType.INCHES: "inches", UnitType.CENTIMETERS: "centimeters" }
str_to_unit_map = { "pixels": UnitType.PIXELS, "inches": UnitType.INCHES, "centimeters": UnitType.CENTIMETERS 

self.__units = str_to_unit_map.get(units_str, UnitType.PIXELS)
self.__units_model.value = unit_to_str_map[self.__units]
```

Also, this avoids tying the file format to Python code (the enum.name).

cmeyer (changes requested):
Good start - I think this needs a merge commit from the main branch and then that will highlight a bug in the lines like `min_size_in_current_units = min_size_in_inches * ConversionUnits[UnitType.INCHES] / ConversionUnits[self.__units]` in `__enforce_width_height_constraints`.

lisham2000:
Couldnt get the merge commits to work so created new branch at #1171

PR #1147, opened 2024-08-27
Add information line on SVG export to give information on image size

lisham2000:


nion/swift/ExportDialog.py
```diff
@@ -263,20 +263,38 @@ class UnitType(enum.Enum):
...
         super().__init__()
-        display_size = self.__calculate_display_size_in_pixels(display_item)
-        self.__width = display_size.width
-        self.__height = display_size.height
-        self.__aspect_ratio = self.__width / self.__height
-        self.__units = UnitType.PIXELS
-        self.__float_to_string_converter = Converter.FloatToStringConverter()
-        self.__primary_field = 'width'  # Primary field to determine which text is calculated
-        self.__enforce_width_height_constraints()
+        if display_item.data_item is not None:
```
cmeyer, line 266: Generally, putting initializers in `__init__` inside an `if` statement is prone to error since it is impossible for other methods to know whether the class has initialized fully or not. If you were to do something like this, you could just do the check and assign `self.__image_2d` either a value or `None` if it wasn't available. However, there is no need to store a field `self.__image_2d` - it can just be removed and calculated on the fly.

nion/swift/ExportDialog.py
```diff
@@ -263,20 +263,38 @@ class UnitType(enum.Enum):
...
+            self.__width = display_size.width
+            self.__height = display_size.height
+            self.__aspect_ratio = self.__width / self.__height
+            self.__units = UnitType.PIXELS
+            self.__float_to_string_converter = Converter.FloatToStringConverter()
+            self.__primary_field = 'width'  # Primary field to determine which text is calculated
 
     def __calculate_display_size_in_pixels(self, display_item: DisplayItem.DisplayItem) -> Geometry.IntSize:
-        if display_item.display_data_shape and len(display_item.display_data_shape) == 2:
+        if display_item.display_data_shape and self.__image_2d:
```
cmeyer, line 278: Use `display_item.used_display_type` here - it will return "line_plot", "image", or something else. You can just check for "line_plot" and assume "image" if it's not "line_plot" for now. i.e.

`if dipslay_item.used_display_type != "line_plot":`

nion/swift/ExportDialog.py
```diff
@@ -263,20 +263,38 @@ class UnitType(enum.Enum):
...
             return Geometry.IntSize(height=display_item.display_data_shape[0], width=display_item.display_data_shape[1])
         return Geometry.IntSize(height=288, width=480)
 
+    @property
+    def image_info(self) -> typing.Optional[str]:
+        assert self.__display_item.data_item and self.__display_item.dimensional_shape
```
cmeyer, line 284: This assert prevents exporting a composite line plot (a line plot with multiple layers).

nion/swift/ExportDialog.py
```diff
@@ -263,20 +263,38 @@ class UnitType(enum.Enum):
...
+        data_information = self.__display_item.displayed_title + "\n"
+        if self.__image_2d:
```
cmeyer, line 286: See comment above about `used_display_type`

nion/swift/ExportDialog.py
```diff
@@ -263,20 +263,38 @@ class UnitType(enum.Enum):
...
+            data_information += "Image Size " + str(self.__display_item.dimensional_shape) + "\n"
+            calibrated_image_size = (self.__display_item.dimensional_shape[0] * self.__display_item.data_item.dimensional_calibrations[0].scale,
+                                     self.__display_item.dimensional_shape[1] * self.__display_item.data_item.dimensional_calibrations[1].scale)
+            data_information += ("(" + str(calibrated_image_size[0]) + " " + self.__display_item.data_item.dimensional_calibrations[0].units + "," +
+                                 str(calibrated_image_size[1]) + " " + self.__display_item.data_item.dimensional_calibrations[1].units + ")")
+        else:
+            data_information += "Line Plot Size " + str(self.__display_item.dimensional_shape[0])
+            calibrated_line_plot_size = self.__display_item.dimensional_shape[0] * self.__display_item.data_item.dimensional_calibrations[0].scale
+            data_information += ("(" + str(calibrated_line_plot_size) + " " + self.__display_item.data_item.dimensional_calibrations[0].units + ")")
```
cmeyer, line 295: To avoid having to look into the `display_item.data_item`, which is `None` for composite displays, you can use `display_item.displayed_dimensional_calibrations` which will give the dimensional calibrations for images, line plots, and composites.

cmeyer (changes requested):
I made code comments - but overall, I think the look and feel of this could be improved. The three lines, without spacing, don't look that great. What about expanding the width of the dialog? Maybe bounce this around to other people and see if they have suggestions.

nion/swift/ExportDialog.py
```diff
@@ -273,10 +274,26 @@ def __init__(self, display_item: DisplayItem.DisplayItem) -> None:
...
+                                     self.__display_item.dimensional_shape[-1] * self.__display_item.data_item.dimensional_calibrations[-1].scale)
+            data_information += ("(" + str(calibrated_image_size[-2]) + " " + self.__display_item.data_item.dimensional_calibrations[-2].units + "," +
+                                 str(calibrated_image_size[-1]) + " " + self.__display_item.data_item.dimensional_calibrations[-1].units + ")")
+        else:
+            assert self.__display_item.display_data_shape
+            data_information += "Line Plot Size " + str(self.__display_item.display_data_shape[0])
+            calibrated_line_plot_size = self.__display_item.display_data_shape[0] * self.__display_item.displayed_dimensional_calibrations[0].scale
+            data_information += ("(" + str(calibrated_line_plot_size) + " " + self.__display_item.displayed_dimensional_calibrations[0].units + ")")
+        return data_information
+
```
KRLango, line 296: Do we need to worry about this info changing once the window is open? If so, this will need a listener on the `property_changed_event` of `self.__display_item` and `self.__display_item.data_item` to catch any relevant changes and trigger a `property_changed_event` for `image_info`

KRLango (changes requested):
The excess row should be removed. Otherwise I am happy with this now.

nion/swift/ExportDialog.py
cmeyer: I suggest not worrying about this for now. It's definitely a concern but we should address it separately.

nion/swift/ExportDialog.py
```diff
@@ -263,6 +264,9 @@ class UnitType(enum.Enum):
 class ExportSizeModel(Observable.Observable):
     def __init__(self, display_item: DisplayItem.DisplayItem) -> None:
         super().__init__()
+        self.__display_item = display_item
+        self.__display_item_listener = self.__display_item.display_property_changed_event.listen(
+            ReferenceCounting.weak_partial(ExportSizeModel.__update_dimensional_calibrations, self))
```
cmeyer, line 269: Let's leave this out for now.

nion/swift/ExportDialog.py
```diff
@@ -273,10 +277,31 @@ def __init__(self, display_item: DisplayItem.DisplayItem) -> None:
...
-        if display_item.display_data_shape and len(display_item.display_data_shape) == 2:
-            return Geometry.IntSize(height=display_item.display_data_shape[0], width=display_item.display_data_shape[1])
+        if display_item.display_data_shape and self.__display_item.used_display_type != "line_plot":
+            return Geometry.IntSize(height=display_item.display_data_shape[-2], width=display_item.display_data_shape[-1])
         return Geometry.IntSize(height=288, width=480)
 
+    def __update_dimensional_calibrations(self, property: str) -> None:
+        """Recalculate the image information string based on updated scales"""
+        if property == "dimensional_calibrations" or property == "displayed_dimensional_calibrations":
+            self.notify_property_changed("image_info")
```
cmeyer, line 287: Leave out for now. This is complicated and I think it's going to need a lot more work. Not only can dimensions change, but the title can change, etc. Let's leave this to another PR.

cmeyer (changes requested):
First note is that this PR doesn't rebase cleanly, so you might want to consider closing and opening a new PR with just the one file. Alternatively you can rebase it yourself and force push to your PR branch.

Next, I think you need to have a uniform way of displaying calibrations with the correct precision. The inspector has a way to do this based on the data dimensions.

Review the code in the `CalibratedSizeFloatToStringConverter` class in `Inspector.py`, particularly the `convert_calibrated_value_to_str` method of that class. I think you can use that class directly.

Without that fix, I get results like this (from exporting the Ronchigram simulator image):

image

nion/swift/ExportDialog.py
```diff
@@ -273,10 +274,26 @@ def __init__(self, display_item: DisplayItem.DisplayItem) -> None:
...
+    @property
+    def image_info(self) -> typing.Optional[str]:
+        data_information = self.__display_item.displayed_title + "\n"
+        if self.__display_item.used_display_type != "line_plot":
+            assert self.__display_item.data_item and self.__display_item.dimensional_shape
+            calibrated_image_size = (self.__display_item.dimensional_shape[-2] * self.__display_item.data_item.dimensional_calibrations[-2].scale,
+                                     self.__display_item.dimensional_shape[-1] * self.__display_item.data_item.dimensional_calibrations[-1].scale)
+            data_information += ("(" + str(calibrated_image_size[-2]) + " " + self.__display_item.data_item.dimensional_calibrations[-2].units + "," +
+                                 str(calibrated_image_size[-1]) + " " + self.__display_item.data_item.dimensional_calibrations[-1].units + ")")
+        else:
```
cmeyer, line 290: Can you change these lines to use `display_item.displayed_dimensional_calibrations` like the other clause?

cmeyer (changes requested):
Overall this looks very close (see my comments about using `displayed_dimensional_calibrations` from today; and my comments about using `CalibratedSizeFloatToStringConverter` from last week), but also needs a clean commit. Easiest way from your branch:

```
git reset --soft master
git commit -m "Add data info line to SVG export dialog."
git show  # to ensure it's as expected
git push --force
```

lisham2000:
> Overall this looks very close (see my comments about using `displayed_dimensional_calibrations` from today; and my comments about using `CalibratedSizeFloatToStringConverter` from last week), but also needs a clean commit. Easiest way from your branch:
> 
> ```
> git reset --soft master
> git commit -m "Add data info line to SVG export dialog."
> git show  # to ensure it's as expected
> git push --force
> ```
I have run these git commands and it has put it to one commit but it now says it has changes 22 files just checking this is right?

PR #1148, opened 2024-08-27
Fix Issue with precision not being maintained when swapping units in SVG or inputting units on SVG dialog

lisham2000:


cmeyer (changes requested):
This PR looks like a combination of other PR's - can you make a clean PR from the master branch and/or delay this until other PR's are merged?

Didn't review for now.

PR #1152, opened 2024-09-10
Add declarative text push button

KRLango:
The line plot display layer inspector section currently uses a text push button to create a cleaner UI. This will likely look far worse with normal pushbuttons here so I added a declarative wrapper to the text push button. This could also be used on the graphics "move to center" button.

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -976,21 +976,11 @@ class LinePlotDisplayLayersInspectorSection(InspectorSection):
                 label_row.add_spacing(12)
                 label_edit_widget.on_editing_finished = functools.partial(change_label, label_edit_widget, display_layer)
                 # move up, move down, add layer, remove layer
-                move_forward_button_widget = Widgets.TextPushButtonWidget(ui, "\N{UPWARDS WHITE ARROW}")
-                move_backward_button_widget = Widgets.TextPushButtonWidget(ui, "\N{DOWNWARDS WHITE ARROW}")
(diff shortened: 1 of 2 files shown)


nion/swift/Inspector.py
```diff
@@ -4985,3 +4975,60 @@ def data_item_delete() -> None:
             return widget
 
         return None
+
+
+def make_line_plot_display_layer_button_row(document_controller: DocumentController.DocumentController,
```
Tiomat85, line 4980: Double blank line before method.

Tiomat85 (changes requested):
Generally fine, just one extra whitespace line.

lisham2000 (approved):
All looks good to me

nion/swift/Inspector.py
KRLango: This is not a class method, so I think 2 blank lines is correct.

Ion-e (approved):
Draft looks good to me

Ion-e (approved):
Looks good, thanks!

cmeyer (changes requested):
This looks good and I'm almost ready to merge it, but I think we should use a more qualified name "nionswift.text_push_button" instead of just "text_push_button" because I think we'll eventually move this code to `nionui` and I want to avoid the hassle of having two items registered with the same name while we migrate.

As long as you're making a change, adding tool tips to the buttons would be a nice addition.

cmeyer:
Merged def9ad917301e71789a7d9e67faf37ee01da5394 333124a12a515c5f10bf93bec2149f4efe27d791

PR #1170, opened 2024-09-19
Change centre graphic button to work without the graphic being selected

KRLango:
Change the callback on the button to setup the action directly to use the related graphic rather than using the selected graphic only

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -3547,8 +3547,16 @@ class GraphicsInspectorSection(InspectorSection):
             locked_row.add_stretch()
 
             def move_to_center_clicked() -> None:
-                action_context = self.__document_controller._get_action_context()
-                self.__document_controller.perform_action_in_context("display_panel.center_graphics", action_context)
+                def reset_position(graphic: Graphics.Graphic) -> None:
+                    graphic.reset_position()
+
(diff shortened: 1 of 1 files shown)


KRLango:
This will likely affect the graphics inspector PR

Tiomat85 (approved):

Ion-e (approved):
Draft looks good to me.

cmeyer:
Sorry for the delay reviewing this.

I propose an alternate solution - and it was easier to just make changes rather than type out the description of the changes. So I filed another PR:

https://github.com/nion-software/nionswift/pull/1185

To explain: I'm hoping to consolidate most user actions into `Window.Action`s. They're a work in progress, but I'd prefer to move them forward rather than eliminate their use. In this case, I expanded the `DisplayPanelCenterGraphicsAction` to use the context for selected graphics, then expanded `_get_action_context_for_display_items` to allow the caller to setup `selected_graphics` too. Then we can keep the `Window.Action` mostly the same in the inspector by just ensuring that (a) create an action context that describes the specific graphic; and (b) uses the `DisplayPanelCenterGraphicsAction`.

Bigger picture on the actions is that they are script-able and will provide a long term solution for scripting complex user actions.

cmeyer (commented):
See my last comment and alternate proposal:
https://github.com/nion-software/nionswift/pull/1185

PR #1171, opened 2024-09-19
Add persistence to Units value in export SVG dialog

lisham2000:


nion/swift/ExportDialog.py
```diff
@@ -261,15 +261,29 @@ class UnitType(enum.Enum):
...
+        UnitType.CENTIMETERS: "centimeters"
+    }
+
+    str_to_unit_map = {
+        "pixels": UnitType.PIXELS,
+        "inches": UnitType.INCHES,
+        "centimeters": UnitType.CENTIMETERS
+    }
+
+    def __init__(self, display_item: DisplayItem.DisplayItem, units_model: UserInterface.StringPersistentModel) -> None:
```
KRLango, line 276: unit_model should be a more general type here, e.g. Model.PropertyModel[str] - that would allow us to change the model used in the future without having to alter this function

KRLango (changes requested):

nion/swift/ExportDialog.py
```diff
@@ -423,7 +438,12 @@ def __init__(self, document_controller: DocumentController.DocumentController,
         super().__init__()
         self.__document_controller = document_controller
         self.__display_item = display_item
-        self.__model = ExportSizeModel(display_item)
+        self.__units_model = UserInterface.StringPersistentModel(
```
cmeyer, line 441: Can you create the persistent model using `ui.create_persistent_string_model` instead of instantiating it directly to be consistent with other similar usages throughout the code. Also can you qualify the storage_key further, something like `export_svg_dialog_units` instead of `export_units`?

nion/swift/ExportDialog.py
```diff
@@ -261,15 +261,29 @@ class UnitType(enum.Enum):
...
+    def __init__(self, display_item: DisplayItem.DisplayItem, units_model: Model.PropertyModel[str]) -> None:
         super().__init__()
+        self.__units_model = units_model
         display_size = self.__calculate_display_size_in_pixels(display_item)
-        self.__last_input_value: float = display_size.width
         self.__aspect_ratio = display_size.width / display_size.height
-        self.__units = UnitType.PIXELS
         self.__float_to_string_converter = Converter.FloatToStringConverter()
         self.__primary_field = 'width'
+        units_str: str = str(self.__units_model.value)
```
cmeyer, line 283: This should be `units_str = self.__units_model.value or str()` since `value` may be `None`. Also, you don't need the type specifier in the original code (or the new code) since the type checker can determine the type from context.

nion/swift/ExportDialog.py
```diff
@@ -261,15 +261,29 @@ class UnitType(enum.Enum):
...
+        self.__units = self.str_to_unit_map.get(units_str, UnitType.PIXELS)
         self.__last_input_units = self.__units
+        self.__last_input_value: float = display_size.width/ConversionUnits[self.__last_input_units]
```
cmeyer, line 286: There should be spaces around the `/` -> ` / ` to make it readable.

nion/swift/ExportDialog.py
```diff
@@ -366,6 +380,7 @@ def units(self, new_units: int) -> None:
         new_enum = UnitType(new_units)
         if self.__units != new_enum:
             self.__units = new_enum
+            self.__units_model.value = self.unit_to_str_map[self.__units]
```
cmeyer, line 383: The `self.__units` could just be eliminated. The value can be computed on the fly using `UnitType(self.__units_model.value or 0)` and set directly. Watch out for None values when making that change, though. You could add a new property named `_units` for use within this class.

I think as part of that change, there might need to be another property `units_index` and that new property should be the one that is used for binding in `ExportSVGHandler`. It's confusing as it is now to have `units` property referring to an `int` and `self.__units` referring to an enum.

There are many ways to handle this, but it needs to be cleaned up and clear to someone reading the code.

nion/swift/ExportDialog.py
```diff
@@ -261,15 +261,29 @@ class UnitType(enum.Enum):
 
 
 class ExportSizeModel(Observable.Observable):
-    def __init__(self, display_item: DisplayItem.DisplayItem) -> None:
+    unit_to_str_map = {
```
cmeyer, line 264: I know this same thing is done in other parts of the code, but I think we should use better naming for these variables - `unit_to_unit_identifier_map` and `unit_identifier_to_unit_map` are verbose, but explain better what they are. `str` is too general and is confusing with the titles used in `ExportSVGHandler`.

cmeyer (changes requested):
This code is improving - but still needs work. There is some confusing naming of `__units` and the property `units` having different types: an enum and int respectively.

cmeyer:
I'd like to propose an alternate easier to understand PR for this issue:
- https://github.com/nion-software/nionswift/pull/1239

The new PR has far less duplicated code (all of the conversion to units is gone), applies cleanly to the trunk, and has comments. In addition, I think it provides a option for reusing the 'Quantity' class introduced in the PR.

If you're happy with the other PR, let's close this one and move on.

cmeyer:
Close in favor of #1239.

PR #1173, opened 2024-09-20
Restore ability to do quick single exports

lisham2000:


nion/swift/resources/menu_config.json
```diff
@@ -41,12 +41,23 @@
...
             {
-                "type": "item",
-                "action_id": "file.export"
-            },
-            {
-                "type": "item",
-                "action_id": "file.export_svg"
+                "type": "sub_menu",
+                "menu_id": "file_export",
+                "title": "Export...",
```
cmeyer, line 46: Sub-menus should not have "..." on them.  But I don't think these should go into a sub-menu, at least for now.

As part of this change, though, we could update the context menu to supply "single export" and "export svg". That's done here: https://github.com/nion-software/nionswift/blob/6f71ebc70e61ea209ba17f7a7e34939cd8f0aeed/nion/swift/DocumentController.py#L2641

nion/swift/DocumentController.py
```diff
@@ -3012,9 +3012,9 @@ def get_action_name(self, context: Window.ActionContext) -> str:
         return self.action_name
 
 
-class ExportAction(Window.Action):
-    action_id = "file.export"
-    action_name = _("Export...")
+class SingleExportAction(Window.Action):
+    action_id = "export.single_export"
```
cmeyer, line 3016: The old `action_id` should not change. It's unlikely that it is used, but the action id's are meant to be stable over long periods of time. Likewise, I'd like to use consistent identifiers, even though your identifier names are better. That means using `file.batch_export` or maybe `file.export_batch` to make it easily distinguishable in an alphabetized list.

Along those same lines, the classes should be in alphabetical order in the file. This is to make it easier to locate them, which is a problem in the huge file. To that end, I'd keep the original name along with the original identifier (for the reasons above): `ExportAction` or perhaps `ExportSingleAction`.

nion/swift/DocumentController.py
```diff
@@ -3080,13 +3099,13 @@ def invoke(self, context: Window.ActionContext) -> Window.ActionResult:
 
 Window.register_action(DeleteItemAction())
 Window.register_action(DeleteDataItemAction())
-Window.register_action(ExportAction())
+Window.register_action(SingleExportAction())
```
cmeyer, line 3102: These should appear in alphabetical order within their group.

nion/swift/DocumentController.py
```diff
@@ -3023,6 +3023,25 @@ def invoke(self, context: Window.ActionContext) -> Window.ActionResult:
...
+        if selected_display_item:
+            window.export_file(selected_display_item)
+        return Window.ActionResult(Window.ActionStatus.FINISHED)
+
+    def is_enabled(self, context: Window.ActionContext) -> bool:
+        context = typing.cast(DocumentController.ActionContext, context)
+        return len(context.display_items) == 1 and context.display_item is not None
+
+
+class BatchExportAction(Window.Action):
```
cmeyer, line 3035: These should appear in alphabetical order, and I'd rename it to `ExportBatchAction` to make sure it's located in the same area of the code as `ExportAction` when alphabetized.

cmeyer (commented):
I know this is still draft, but I happened to be looking at it so thought I throw in some comments now.

nion/swift/resources/menu_config.json
lisham2000: Hi Chris, Can I ask why we wouldn't want to group these in a sub menu?

nion/swift/DocumentController.py
Tiomat85: Can I ask the silly question of why do classes need to be alphabetical in a file?  Most modern IDEs (including PyCharm) can list classes/functions in a file in alphabetical order anyway.  If a file has enough classes in it that that list isn't enough surely that is indicative that the file could do with being broken down into smaller chunks rather than trying to use the first few letters of a class name to create pseudo-structure - potentially leading to names that are less descriptive of the actual function and instead are named just to keep them together.

Tiomat85 (approved):

nion/swift/DocumentController.py
cmeyer: They're not really totally alphabetical - only within each major section. It's simply something to make it easier when searching for commands. This file should definitely be broken down into more manageable pieces and now might be a sensible time to do that (just before a release). Or maybe just after. Either way, when a file gets split, it's more difficult to get a history of that file for regressions, so I tend not to do it for that reason.

nion/swift/resources/menu_config.json
cmeyer: I'm not 100% opposed to it, but the File menu is relatively short, so adding an item isn't a big change. The downside of the sub menu is it is (a) less visible to users exploring the menus (b) one extra click/mouse-motion to select.

I actually think it might be nice to have them all visible with a separator between the 'Import' items and the 'Export' items. This would be the addition of two lines in the menu.

If your group has a strong opinion, it's fine. But if not, let's go with no sub-menu with the option of putting them in a sub-menu in another PR.

nion/swift/DocumentController.py
cmeyer: NOTE: the important part of my comment is to not rename the existing command.

cmeyer:
Merged with minor edits 3e084a57873f79184354498fbf111c1d04f7c9fd

PR #1185, opened 2024-10-03
Fix #1158. Center correct graphic when using inspector.

### LINKED ISSUE #1158: Bug - Inspector: graphic inspector center button has broken functionality
The button to centre graphiis in the display items is broken when attempted to be used from the *display item* inspector.
The button works correctly from the graphic's own inspector.

Please take in consideration that because the item selected is the "display item", what would we expect about pressing the centre button here? To affect all graphics?

cmeyer:


KRLango (approved):
That is a much cleaner way to achieve this. I was not aware it was possible to pass the graphics via the action context before. I will close the other PR for this.

cmeyer:
Great - I'll merge this one. It might be helpful to review the action context idea. The context itself contains the "selected graphics" (in this case, not selected, but specified), and I had to add the ability to construct an action context with the "selected graphics." The action context idea (loosely based on how Blender does things) is a work in progress, and we should look at whether naming is consistent, etc.

PR #1188, opened 2024-10-03
Use batch updating for updating display item display properties.

cmeyer:


cmeyer:
Merged aa55ed526106cfb71f7f7e3608b3867fbb77e970

PR #1191, opened 2024-10-04
Fix #1176. Make Collections panel filter/data-group persistent.

### LINKED ISSUE #1176: Collections panel should store its state persistently and return to same state on relaunch
The collection panel selected filter or group should be persistent (per project) between launches or when switching projects. There should be integration tests.

Be aware of #1189 when testing (contents of data group is not persistent).

cmeyer:


cmeyer:
Merged df528abb14b7be7cfb7135168e06451756b6fff7

PR #1201, opened 2024-10-10
Move line plot display layers inspector section to declarative UI

KRLango:
Switching the line plot display layers section of the inspector to use the declarative UI

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -49,6 +49,7 @@ from nion.utils import Model
 from nion.utils import Observable
 from nion.utils import ReferenceCounting
 from nion.utils import Registry
+from nion.utils import Stream
 from nion.utils import Validator
 
 if typing.TYPE_CHECKING:
@@ -925,16 +926,19 @@ class DisplayLayerPropertyCommandModel(Model.PropertyChangedPropertyModel[typing
             self.__document_controller.push_undo_command(command)
 
 
+
 class LinePlotDisplayLayerHandler(Declarative.Handler):
-    def __init__(self, line_plot_display_layer_model: LinePlotDisplayLayerModel, update_layer_list: typing.Callable[[], None]):
+    def __init__(self, line_plot_display_layer_model: LinePlotDisplayLayerModel):
         super().__init__()
         self._line_plot_display_layer_model = line_plot_display_layer_model
-        self.__update_layer_list = update_layer_list
+        self.integer_to_string_converter = Converter.IntegerToStringConverter()
 
         u = Declarative.DeclarativeUI()
         self.ui_view = u.create_column(
             u.create_row(
-                u.create_label(text=_("Layer") + f" {self._line_plot_display_layer_model.index}:"),
+                u.create_label(text=_("Layer ")),
+                u.create_label(text="@binding(_line_plot_display_layer_model.index)"),
+                u.create_label(text=":"),
                 u.create_line_edit(text="@binding(_line_plot_display_layer_model.label_model.value)"),
                 u.create_spacing(12),
                 spacing=12
@@ -961,9 +965,9 @@ class LinePlotDisplayLayerHandler(Declarative.Handler):
             ),
             u.create_row(
                 u.create_label(text=_("Data Index")),
-                u.create_line_edit(text="@binding(_line_plot_display_layer_model.data_index_model.value)"),
+                u.create_line_edit(text="@binding(_line_plot_display_layer_model.data_index_model.value, converter=integer_to_string_converter)"),
                 u.create_label(text=_("Row")),
-                u.create_line_edit(text="@binding(_line_plot_display_layer_model.data_row_model.value)"),
+                u.create_line_edit(text="@binding(_line_plot_display_layer_model.data_row_model.value, converter=integer_to_string_converter)"),
                 u.create_stretch(),
                 spacing=12
             ),
@@ -984,9 +988,16 @@ class LinePlotDisplayLayerHandler(Declarative.Handler):
             u.create_row(
                 u.create_label(text=_("Stroke Width"), width=80, height=30),
                 u.create_spacing(44 + 8),  # color push button width + spacing to avoid collapse
-                u.create_line_edit(text="@binding(_line_plot_display_layer_model.stroke_width_model.value)", width=36),
+                u.create_line_edit(text="@binding(_line_plot_display_layer_model.stroke_width_model.value, converter=integer_to_string_converter)", width=36),
                 u.create_stretch(),
                 spacing=8
+            ),
+            u.create_row(
+                u.create_label(text=_("Complex Display Type:"), width=120),
+                u.create_combo_box(items=self._line_plot_display_layer_model.display_type_items,
+                                   current_index="@binding(_line_plot_display_layer_model.current_display_type_index_model.value)"),
+                u.create_stretch(),
+                visible="@binding(_line_plot_display_layer_model.is_data_complex)"
             )
         )
 
@@ -1000,7 +1011,6 @@ class LinePlotDisplayLayerHandler(Declarative.Handler):
                                                            index - 1)
             command.perform()
             self._line_plot_display_layer_model.document_controller.push_undo_command(command)
-            self.__update_layer_list()
 
     def _move_layer_backward(self) -> None:
         index = self._line_plot_display_layer_model.index
@@ -1012,7 +1022,6 @@ class LinePlotDisplayLayerHandler(Declarative.Handler):
                                                            index + 1)
             command.perform()
             self._line_plot_display_layer_model.document_controller.push_undo_command(command)
-            self.__update_layer_list()
 
     def _add_layer(self) -> None:
         index = self._line_plot_display_layer_model.index + 1  # add new layer after current layer
@@ -1021,7 +1030,6 @@ class LinePlotDisplayLayerHandler(Declarative.Handler):
                                                       index)
         command.perform()
         self._line_plot_display_layer_model.document_controller.push_undo_command(command)
-        self.__update_layer_list()
 
     def _remove_layer(self) -> None:
         index = self._line_plot_display_layer_model.index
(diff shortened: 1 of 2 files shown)


nion/swift/Inspector.py
```diff
@@ -292,6 +292,13 @@ def finish_widget_content(self) -> None:
         """Subclasses should all this after calls to add_widget_content."""
         pass
 
+    def remove_widget_from_content(self, widget: UserInterface.Widget) -> None:
+        # need to remove widget and spacing from before the widget
+        index = self.__section_content_column.children.index(widget)
+        spacing = self.__section_content_column.children[index - 1]
```
Tiomat85, line 298: Is this guaranteed to have a spacing before the widget?

KRLango: Yes based on the function `add_widget_to_content` - line 286. If I didn't do that, the space there grows every time you interact with any of the controls.

Ion-e:
When adding composite line plots, the second line plot fill colour appears as Blank.

lisham2000 (approved):

nion/swift/test/Inspector_test.py
```diff
@@ -1440,12 +1440,13 @@ def test_setting_display_layer_to_no_display_data_channel_and_back(self):
...
-            display_data_channel_index_widget.on_editing_finished("")
-            display_data_channel_index_widget.on_editing_finished("A")
-            display_data_channel_index_widget.on_editing_finished(0)
+            line_plot_display_layers_section = inspector_panel.widget.find_widget_by_id("line_plot_display_layers_inspector_section")
+            line_plot_display_layer_model = line_plot_display_layers_section._handler._line_plot_display_layer_handlers[0]._line_plot_display_layer_model
+            line_plot_display_layer_model.data_index_model.Value = None
+            line_plot_display_layer_model.data_index_model.Value = 0
+            line_plot_display_layer_model.data_index_model.Value = ""
+            line_plot_display_layer_model.data_index_model.Value = "A"
+            line_plot_display_layer_model.data_index_model.Value = 0
```
lisham2000, line 1449: Why do you test =0 twice?

Ion-e:
When pressing the "+" button to add a layer > Current behaviour is opposite to Master > A new layer is inserted in the list (as opposed to appended in master)

nion/swift/test/Inspector_test.py
KRLango: Yes, the particular test is `test_setting_display_layer_to_no_display_data_channel_and_back`, the last set to 0 is the `back` of that.

cmeyer (changes requested):
There are a few issues with this PR - but the main thing is that the list structure is being done manually rather than using a list model with a declarative column.

I'm attaching a new version of `Inspector.py` which switches to a cleaner structure where the column is managed by the list model; and the list model is observed from the display item. This avoids having to explicitly manage the items in the list (i.e. `remove_widget_from_container` and its ilk are no longer needed). This was easier to show rather than explain.

In addition, I've made a number of other minor changes to fix some, but not all, issues that came up in testing. There were things like changing the 'data index' threw a conversion error that prevented it from working. I did NOT fix the problem of the complex display type chooser not appearing in the inspector.

I'm also attaching the test procedure I was testing against.

The new version of `Inspector.py` requires this PR: https://github.com/nion-software/nionutils/pull/30

[Inspector.py.zip](https://github.com/user-attachments/files/17388578/Inspector.py.zip)

[test-procedure.zip](https://github.com/user-attachments/files/17388597/test-procedure.zip)

nion/swift/Inspector.py
cmeyer: This is unused. It can be removed.

nion/swift/Inspector.py
```diff
@@ -918,245 +926,246 @@ def _set_property_value(self, property_value: typing.Any) -> None:
...
+                u.create_label(text=_("Stroke Color"), width=80),
+                {"type": "nionswift.color_chooser", "color": "@binding(_line_plot_display_layer_model.stroke_color_model.value)"},
+                u.create_line_edit(text="@binding(_line_plot_display_layer_model.stroke_color_model.value)", placeholder_text="None", width=80),
+                u.create_stretch(),
+                spacing=8
+            ),
+            u.create_row(
+                u.create_label(text=_("Stroke Width"), width=80, height=30),
+                u.create_spacing(44 + 8),  # color push button width + spacing to avoid collapse
+                u.create_line_edit(text="@binding(_line_plot_display_layer_model.stroke_width_model.value, converter=integer_to_string_converter)", width=36),
```
cmeyer, line 991: Stroke width is a float, not an integer.

nion/swift/Inspector.py
```diff
@@ -918,245 +926,246 @@ def _set_property_value(self, property_value: typing.Any) -> None:
...
 
-class LinePlotDisplayLayersInspectorSection(InspectorSection):
-    def __init__(self, document_controller: DocumentController.DocumentController, display_item: DisplayItem.DisplayItem) -> None:
-        super().__init__(document_controller.ui, "line-plot-display-layer", _("Line Plot Display Layers"))
-        ui = self.ui
 
-        def change_label(label_edit_widget: UserInterface.LineEditWidget, display_layer: DisplayItem.DisplayLayer, label: str) -> None:
-            index = display_item.display_layers.index(display_layer)
-            command = ChangeDisplayLayerPropertyCommand(document_controller.document_model, display_item, index, "label", label)
+class LinePlotDisplayLayerHandler(Declarative.Handler):
```
cmeyer, line 930: There is now an extra blank line before this class.

nion/swift/Inspector.py
```diff
@@ -918,245 +926,246 @@ def _set_property_value(self, property_value: typing.Any) -> None:
...
-        def change_data_index(data_index_widget: UserInterface.LineEditWidget, display_layer: DisplayItem.DisplayLayer, data_index_str: str) -> None:
-            data_index = Converter.IntegerToStringConverter(pass_none=True).convert_back(data_index_str)
-            display_data_channel = display_item.display_data_channels[data_index] if data_index is not None else None
+    @property
+    def index(self) -> int:
+        if self.display_layer in self.display_item.display_layers:
+            return self.display_item.display_layers.index(self.display_layer)
+        else:
+            # we occasionally get an old LinePlotDisplayLayerModel that is not in the list
+            return 0
```
cmeyer, line 1088: This should never happen and indicates a bug somewhere else. How do you reproduce this?

nion/swift/Inspector.py
```diff
@@ -918,245 +926,246 @@ def _set_property_value(self, property_value: typing.Any) -> None:
...
+    def __init__(self, document_controller: DocumentController.DocumentController, display_item: DisplayItem.DisplayItem) -> None:
+        super().__init__(document_controller.ui, "line-plot-display-layer", _("Line Plot Display Layers"))
+        self.widget_id = "line_plot_display_layers_inspector_section"
+        self.__document_controller = document_controller
+        self.__display_item = display_item
+        display_layer_list_model = ListModel.ObservedListModel[DisplayItem.DisplayLayer](self.__display_item, "display_layers")
+        self._handler = LinePlotDisplayLayersSectionHandler(self.__document_controller, self.__display_item, display_layer_list_model)
+        self.__widget = Declarative.DeclarativeWidget(self.__document_controller.ui, self.__document_controller.event_loop, self._handler)
+        self.add_widget_to_content(self.__widget)
+
```
cmeyer, line 1168: This adds an extra blank line. There should be two, not three.

nion/swift/Inspector.py
```diff
@@ -918,245 +926,246 @@ def _set_property_value(self, property_value: typing.Any) -> None:
...
+        self.current_display_type_index_model = Model.PropertyModel[int](self._display_type_reverse_map.get(current_flag, 0))
+
+        self.__data_index_listener = self.data_index_model.property_changed_event.listen(
+            ReferenceCounting.weak_partial(LinePlotDisplayLayerModel.__on_data_index_changed, self))
+        self.__display_type_index_listener_model = self.current_display_type_index_model.property_changed_event.listen(
+            ReferenceCounting.weak_partial(LinePlotDisplayLayerModel.__on_display_type_index_changed, self))
+        self.__display_layer_list_added_listener = display_layers_list_model.item_added_event.listen(
+            ReferenceCounting.weak_partial(LinePlotDisplayLayerModel.__on_display_layer_list_changed, self))
+        self.__display_layer_list_removed_listener = display_layers_list_model.item_removed_event.listen(
+            ReferenceCounting.weak_partial(LinePlotDisplayLayerModel.__on_display_layer_list_changed, self))
```
cmeyer, line 1080: I'm not sure this is the best way to do things. I suspect this leads to the errors you mention below (old LinePlotDisplayLayerModel in list). This might take some more rework before we can handle the index properly. I think that we're going to have to refactor `ChangeDisplayLayerDisplayDataChannelCommand` to take a `DisplayLayer` directly and track it via uuid rather than index. That would be a separate PR that goes in _before_ this PR. Doing that would eliminate most uses of `index` except for the display in the text item - but that could be accomplished using the straightforward implementation of `index` (`return self.display_item.display_layers.index(self.display_layer)`).

cmeyer (changes requested):
Some minor changes + major change for handling the index.

nion/swift/Inspector.py
cmeyer: Ok I have a plan - let's eliminate the 'index' property from `LinePlotDisplayLayerModel`. To do this, `ChangeDisplayLayerDisplayDataChannelCommand` the move methods in `LinePlotDisplayLayerHandler` can calculate their indexes on the fly. That leaves only the static text item - and let's just not put the index there for now. That eliminates the `index` property and issue for now.

nion/swift/Inspector.py
KRLango: I added a listener to detect display layers being added/removed to trigger a property changed event for the index so the "Layer x" text could update. When I tested that with moving layers up/down the line `return self.display_item.display_layers.index(self.display_layer)` was throwing an exception saying the display_layer was not in the list. This didn't happen every time so I assumed it was a race condition where the old model was still attached to the event some of the time.

cmeyer:
Merged here c446d54d9423a9bd4af6b60314de8260a9718e79 with minor edits

Ion-e (commented):
Looks good to me. :)

Ion-e (commented):
Seems good.

PR #1202, opened 2024-10-10
Fix #707. Add ability to register color tables from packages.

### LINKED ISSUE #707: Add mechanism for packages to include additional color maps

cmeyer:
I just need one reviewer for this...


KRLango (approved):
Looks good to me.

Not sure if this belongs here or not, but is it worth having an example of how to register a color table somewhere?

PR #1213, opened 2024-10-15
Fix inspector typing issues (mypy 112).

cmeyer:


KRLango (approved):
All seems good to me.

PR #1215, opened 2024-10-16
Relax undo check to ignore changes outside display item (computations) after adding graphics

### LINKED ISSUE #1159: Control + Z doesn't remove the Lattice tool
As title.

cmeyer:
- **Add a few mask graphic undo tests to confirm nominal behavior.**
- **Fix #1159. Relax undo check to ignore changes outside display item (computations) after adding graphics.**

I only need one reviewer.

Also, I decided to delve into this because I thought it might be a canvas drawing issue. It wasn't, but I uncovered a slew of issues. This also introduces a new type of test that uses Window.Action and requires keyboard focus in the test. So if any reviewers want to look at the test, it's worthwhile for future reference.


cmeyer:
Merged 9c435970617642fa224f15a0beea9a6aaaf1cb77

Ion-e (commented):
Tested this and looks good, thanks!

PR #1216, opened 2024-10-16
Fix missing color map inspector issue. Improve color map registry.

cmeyer:
- **Fix inspector issue where color table missing.**
- **Revise color table to include identifier. Registry now takes object and registers from data, json, or xml.**


cmeyer:
Merged around here 1d21feda6baeeb9c15cc9e7b6ca77ab6ab08c789

Issue #96, opened 2018-01-10
Make it easier to center graphics


cmeyer:
The typical way this is done is that if grid snapping is enabled, the graphic should snap to the center. We might consider an indicator to indicate that it is centered.

Reference: A35. I often re-center circle annotations, and check whether ones I have are centered. Annotations that are already aligned on the 0,0 center pixel should be give a special helpful cue that they are already centered. Perhaps the center crosshair could turn green, or the double in size, or something like that.

cmeyer:
Add right click to center, align graphics

cmeyer:
Edit > Align Graphic to Center

Issue #84, opened 2017-09-26
Selecting last member of Library/Collections in data panel causes problems


cmeyer:
MH: Selecting the last member of Library or Collections removes the first member from the displayed list.  Restarting swift makes the hidden member reappear.

Issue #58, opened 2017-09-05
Provide ability to select multiple graphics by dragging rubber band


cmeyer:
### Tasks
- [ ] Implement a tool in the raster display
- [ ] Add drawing code to show the rubber band while dragging
- [ ] Add tests for various functionality
- [ ] Dragging with shift key should select additional items

Ion-e:
What is the requirement we are trying to solve by this feature? @cmeyer

PR #1221, opened 2024-10-22
Fix #651. Add title manager to ensure unique displayed titles.

### LINKED ISSUE #651: Ensure that all display items show a unique name
See #32 snapshot title based on template.

From BPS 2020-10-28:

2) Is there an open issue for coming up with a good long term solution for how files are automatically named? Especially in combination with no timestamps, having a bunch of identically named files is a recipe for confusion and loss of useful data.

From AL 2020-10-21:

Actually looking at this, I noticed 2 things that could be improved:
1. Numbering of images. It seems those reset each session, so I have three images called, "SuperScan (MAADF) 3" in the same library. Hmmm. I don't know if that's important or not, as they have timestamps and so on.
2. Numbering of sequences. - Can we number sequences the same way we number images? I have about 8 movies all called "Recording of SuperScan (MAADF)"
Actually this second point is more important. I suspect we are moving to a lot more movies now...

### See Also
- https://github.com/nion-software/nionswift/issues/32
- https://github.com/nion-software/nionswift-instrumentation-kit/issues/184

cmeyer:
How I tested:

Use Processing > Generate to create 2+ data items with same name. Take FFT of results to generate new data items. Run acquisition and take multiple snapshots and see that they have unique titles.

### DIFF
--- a/nion/swift/model/DisplayItem.py
+++ b/nion/swift/model/DisplayItem.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 # standard libraries
 import asyncio
+import collections
 import concurrent.futures
 import contextlib
 import copy
@@ -29,7 +30,6 @@ from nion.swift.model import Cache
 from nion.swift.model import Changes
 from nion.swift.model import ColorMaps
 from nion.swift.model import DataItem
-from nion.swift.model import DynamicString
 from nion.swift.model import Graphics
 from nion.swift.model import Model
 from nion.swift.model import Persistence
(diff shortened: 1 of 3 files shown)


KRLango (commented):
I have concerns about this PR. 

What is the problem it is trying to solve?
Having unique names - but why do users want unique names, and does this solve their actual need?

During my testing, I got 2 identical suffixes `#96` with 4 snapshots. While I accept this will be rare - how often do we want to accept collisions happening?

Will the random nature of the suffixes confuse users? For example, the first few I got all appeared to be just numbers `#58`, `#96`, '#84`. Are users going to wonder what is going on with those numbers.

I believe the implementation done for the Acquisitions is a much cleaner user experience, and I think something similar would work better here.

cmeyer:
This may not be needed in light of the acquisition index change and it may indeed be confusing to users. We can discuss that before merging (or not merging).

However, I'm concerned that even if we decided it was a useful feature, it shouldn't ever have duplicate suffixes. So I'm wondering how that occurred. I tested by changing the minimum length to 1 and then creating multiple snapshots and when it got to a duplicated suffix out of the 0-9A-F set, it increased the length to 2. Did this happen when the '96' appeared for you, e.g. increase to length 3 to distinguish between the two '96's?

KRLango:
I have gone back to the project I used for this. I did some snapshots using ctrl + s and some using the menu. When I found the 2 96's I think it was one with each since the names are slightly different (Display Snapshot vs Snapshot).

{4BBBCB56-BE3B-48B9-8B3E-935582C87E9D}

cmeyer:
Ok that makes sense - but it does highlight a problem with this scheme. I'm going to change this to draft while I think it through.

PR #1243, opened 2024-11-04
Introduce nhdf file format based on HDF5 for import/export any display/data item

cmeyer:
This adds a new import/export file format which can handle any internal data type and also includes the associated display. It is based on HDF5 and I chose the file extension "nhdf". The commits up until the last one are preparatory and can be reviewed in order. The last commit (bottom of the list below) is the actual introduction of the file format. This PR is not optimized or tested for performance. The writer implementation currently only exports a single display/data item; but the format itself supports multiple data and display items. The reader implements multiples already.

Testing this PR involves exporting any display item to a file and re-importing it into the same or another project. During import, all internal objects are assigned new uuid's so there is no conflict.

- **Change io handler to determine writeability based on display item rather than data only.**
- **Restructure receive_files to be not-threaded and not-undoable.**
- **Change _make_storage_handler to take metadata rather than the data item itself.**
- **Narrow typing of storage_system passed to Project to ProjectStorageSystem.**
- **Refactor project loading into smaller methods. Add function to register storage handler.**
- **Use numpy.copy to be more compatible with h5py arrays.**
- **Add support for updating all display item uuid's before added to document.**
- **Refactor import to return a more generalized import data structure.**
- **Restructure importer to faciliate low level importer/exporter via registration.**
- **Add ndata HDF file format via low level import-export driver.**

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -45,6 +45,7 @@ from nion.swift.model import DataGroup
 from nion.swift.model import DataItem
 from nion.swift.model import DisplayItem
 from nion.swift.model import DocumentModel
+from nion.swift.model import FileStorageSystem
 from nion.swift.model import Graphics
 from nion.swift.model import ImportExportManager
 from nion.swift.model import Notification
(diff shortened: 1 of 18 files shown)


jamesrussell216:
I don't have anything to comment on this one.

jamesrussell216 (approved):
I don't have anything to comment on this one.

KRLango (approved):
LGTM

PR #1248, opened 2024-11-07
Fix that information for 5d data doesn't show in info panel

lisham2000:


cmeyer:
Again, reviewing takes longer than fixing. The indices were out of order and the bounds check was missing index=4. Fixed and committed here a8e82e3a87fd4d3fcd58c291784de3fef3675690. See also the related data fix nion-software/niondata@6b5a7f665955df53367bf47628da8d3291801ddd

Issue #1210, opened 2024-10-14
Inspector: Graphics section is not being removed when removing graphics


Ion-e:
As title.

STR: Select a display item > Add a graphic (or drag some line plots if on line plot graph) > Select the same display item again > Observe the "Graphics" section gets added to the inspector > Control Z > Graphic should disappear < Graphic section remains unchanged. (To actually remove the graphics section you need to "manually refresh".)

Ion-e:
Update: the particular section of the deleted graphic gets removed, however the "graphics section" in the inspector remains.

cmeyer:
@Ion-e I can't reproduce this anymore. Recheck?

Ion-e:
Issue verified as failed.

![Image](https://github.com/user-attachments/assets/5c9aa01a-042b-4b10-99c9-2f5dac06439a)

Issue #1219, opened 2024-10-22
List of graphics in inspector doesn't update on a undo/redo command


KRLango:
To recreate:

- Create a display item.
- Add 3 graphics to the display item.
- Select the display item and check the graphics inspector section: there should be 3 graphics there.
- Delete 1 of the graphics.
- Select the display item and check the graphics inspector section: there should be 2 graphics there.
- Press Ctrl+Z (or equivalent)
- Check the graphics inspector section: there should be 3 graphics there now, but there are only 2.

Ion-e:
Issue verified as fixed, thanks!

PR #1253, opened 2024-11-18
Update graphics inspector section to use ObservedListModel to catch c…

KRLango:
…hanges to the graphics list

#1210

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -3199,6 +3199,7 @@ class GraphicsSectionHandler(Declarative.Handler):
     def __init__(self, document_controller: DocumentController.DocumentController,
                  display_item: DisplayItem.DisplayItem,
                  graphics_model: ListModel.ObservedListModel[DisplayItem.DisplayLayer]):
+        super().__init__()
         self.__document_controller = document_controller
         self.__display_item = display_item
         self._graphics_model = graphics_model
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -3195,12 +3195,14 @@ class GraphicsInspectorHandler(Declarative.Handler):
             return self.__shape_and_pos_func()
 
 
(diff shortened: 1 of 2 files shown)


KRLango:
I think #1219 should also be fixed by this

nion/swift/Inspector.py
```diff
@@ -3195,12 +3195,14 @@ def __create_position_and_shape_ui(self) -> Declarative.UIDescriptionResult:
...
 
 
-class GraphicsCalibrationHandler(Declarative.Handler):
-    def __init__(self, document_controller: DocumentController.DocumentController, display_item: DisplayItem.DisplayItem):
-        super().__init__()
-
+class GraphicsSectionHandler(Declarative.Handler):
+    def __init__(self, document_controller: DocumentController.DocumentController,
+                 display_item: DisplayItem.DisplayItem,
+                 graphics_model: ListModel.ObservedListModel[DisplayItem.DisplayLayer]):
```
cmeyer, line 3201: This needs a `super().__init__()` to properly initialize `Declarative.Handler`, which itself has to initialize `Observable`.

cmeyer (commented):
Other than the missing `super.__init__()`, this looks good. Any reason it is still a draft?

I'm not sure why mypy doesn't catch the missing super.init... but see [this issue](https://github.com/python/mypy/issues/11809).

Also, note that `flake8 --ignore E501 nion/swift/Inspector.py` does not catch this issue either.

However, `pylint nion/swift/Inspector.py` it shows up as "nion/swift/Inspector.py:3199:4: W0231: __init__ method from base class 'Handler' is not called (super-init-not-called)". However it is mixed into 100's of other lint errors. Oh well.

lisham2000 (approved):
All looks good

Ion-e (commented):
Looks good to me!

cmeyer:
Squashed/merged here 2d73324c190d6a05b097286cac3a98a816ab6b64

Issue #1140, opened 2024-08-16
Ring mask calculates filter incorrectly for low/high pass cases


cmeyer:
The graphic editor always uses "ring 1" for the graphical display of both low pass and high pass.

The filter uses "the innermost ring" for high pass; and "the outermost ring" for low pass.

Instead, I propose that the editor uses "ring 1" for high pass and "ring 2" for low pass. And the filter does the same.

This will lead to some imperfect situations if the user has made "ring 1" have a smaller radius than "ring 2" and switches from band-pass to high pass (and similar for low pass) but it will be easy to remedy.

We could possibly swap radius when switching from band pass to low/high if they're reversed.

KRLango:
An alternate proposal that I think will be clearer to the users:

- Always use ring 1 for high and low pass filters (on the UI and for the filter)
- Hide the ring 2 UI when high or low pass is selected so there is no confusion to the user which is being used

cmeyer:
I think neither solution is ideal - the use case I was considering is if the user is switching between band-pass and either low- or high- pass (back and forth), then either proposal can lead to frustrating behavior. Your proposal is easier to implement, so let's go with that for now.

See
- https://github.com/nion-software/nionswift/issues/1143

Ion-e:
Draft looks good to me!

cmeyer:
Closed with #1396

Ion-e:
Looks good!

PR #1254, opened 2024-11-19
Ensure graphic and calculated mask match

KRLango:
#1140 
After trying the approach I suggested, I saw the problems that @cmeyer referenced with it so ended up implementing his original solution as I believe that it is a better user experience.


Tiomat85 (approved):

Ion-e (commented):
Looks good!

jamesrussell216:
Can this one be put in now @KRLango?

cmeyer:
This still looks incorrect... in the first screenshot below, the mode is set to 'low pass' and the resulting masked image (lower left) is showing (i.e. "passing") the high frequency data (further from FFT center). In the second screenshot below, the opposite is occuring - the low frequency (closer to FFT center) is being "passed". I don't know how you want to fix it, but I think the 'low pass' and 'high pass' labels in the inspector are still reversed - but I think instead of changing the inspector, 'get_mask' and 'draw' (band pass) should be reversed. This might introduce backwards incompatibility in the file format, but I think it's acceptable to get it right.

image

image

cmeyer (changes requested):
See comment above.

jamesrussell216:
@cmeyer just looking at your screenshots there - is it my imagination or is there a weird ~few pixel shift of the top hemisphere to the bottom in the masked images?

cmeyer:
Yes that needs to be investigate, although this could be a consequence of the Hermitian data layout.

KRLango:
It has been a while since I last looked at this and the code has progressed to the point where it is easier to make a new branch and PR rather than try and merge this one. Specifically some of the functions this PR edits have moved into new classes. The new PR is #1396

PR #1260, opened 2024-11-22
Switch centre graphic button to clickable text style

KRLango:


cmeyer:
I think either button style is acceptable. Do you all of a consensus preference there?

cmeyer:
After nion-software/nionui#92 is merged (see PR nion-software/nionui#93) I propose using a push button with 'minimal' style and a width around 20. This will make it less visually intrusive and still make it look like a button.

cmeyer:
I went ahead with my proposed change:

cc9aa6e84b79f016828f3fac5d392b827e2fdd2a

Do you still want this PR on top of that?

KRLango:
I like the new button you have done. I don't think this PR is needed following that commit.

Issue #926, opened 2023-01-17
Make it possible to configure a run script to appear in the menu or a button


cmeyer:
Requested multiple times.

ND 2021-11-05:

2b) Computations are just snippets of code. It should become trivial to provide a file with the name and the snippet and get it installed in the menu.

---

Also: with an optional keyboard accelerator.

Also: perhaps there should be a dedicated panel for holding buttons. One could click '+' in that window to add run script buttons.

If the ability to group buttons/controls into separate panels, then this issue is dependent on:

See also:
- #508

PR #1263, opened 2024-11-25
improve master detail browser

cmeyer:
- **Change master detail handler to take sizing properties as parameter.**
- **Add support for add/remove buttons to master-detail browser.**
- **Refactor master detail browser to use simplified list based on property rather than mapping.**
- **Add tracking of parent entity in entities for compatibility with persistence framework.**
- **Expand entity custom class factory capability.**
- **Add ability to write entity to storage when inserted/removed.**

The master-detail browser is currently used in the project items dialog, which is a debugging dialog used to examine most objects in a project.

However, it is a generally useful UI element that can be used in future situations. Particularly, I'm working on a capability to add custom command buttons #926.

Additionally, the original main project items (data item, display item, computation, etc.) were implemented within a persistence framework which ensures that they get written to disk when changed. A newer cleaner entity system also exists which potentially could simplify the object and persistence framework. The entity system lacks some capabilities that the persistence system implements, such as writing out sub-items when inserted/removed from relationships. This PR also addresses those in conjunction with the master-detail handler.

All of the changes in this PR are required to implement the custom command buttons which will be submitted in an upcoming PR.

Also, these are all "internal" changes that won't have a visible effect on the existing project items dialog, although it should continue to work the same.

### DIFF
--- a/nion/swift/EntityBrowser.py
+++ b/nion/swift/EntityBrowser.py
@@ -12,6 +12,7 @@ import asyncio
 import dataclasses
 import gettext
 import operator
+import threading
 import typing
 
 # third party libraries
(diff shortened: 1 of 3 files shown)


Ion-e:
What is "entity system" and "persistence system"? :)

cmeyer:
> What is "entity system" and "persistence system"? :)

Internal capabilities.

PR #1265, opened 2024-11-26
Fix #1203. Eliminate use of data sampling to calculate complex display limits.

### LINKED ISSUE #1203: test_display_range_is_correct_on_complex_data_display_as_log_absolute is unstable
```
FAIL: test_display_range_is_correct_on_complex_data_display_as_log_absolute (test.Display_test.TestDisplayClass.test_display_range_is_correct_on_complex_data_display_as_log_absolute)
----------------------------------------------------------------------
  File "/home/runner/work/nionswift/nionswift/nion/swift/test/Display_test.py", line 589, in test_display_range_is_correct_on_complex_data_display_as_log_absolute
    self.assertAlmostEqual(display_data_channel.get_latest_computed_display_values().display_range[0], min_)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: np.float64(0.0) != np.float64(-744.4400719213812) within 7 places (np.float64(744.4400719213812) difference)

----------------------------------------------------------------------
```

cmeyer:
The sample size was 200 and in no way could be representative of an
arbitrary complex valued image displayed in log-absolute. This was probably
done for perceived performance; but modern processors make the performance
issue minimal. So do not sample; just use the full range until a good
general sampling solution is implemented.

### DIFF
--- a/nion/swift/model/DisplayItem.py
+++ b/nion/swift/model/DisplayItem.py
@@ -177,23 +177,12 @@ class GraphicSelection:
             self.changed_event.fire()
 
 
-def calculate_display_range(display_limits: DisplayLimitsType,
-                            data_range: typing.Optional[typing.Tuple[float, float]],
-                            data_sample: typing.Optional[_ImageDataType],
-                            xdata: typing.Optional[DataAndMetadata.DataAndMetadata],
-                            complex_display_type: typing.Optional[str]) -> typing.Optional[typing.Tuple[float, float]]:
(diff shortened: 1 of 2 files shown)


Brow71189 (commented):
Generally this looks fine to me. The only issue I can see is that removing access to `xdata` from `calculate_display_range` will make it harder to implement more sophisticated algorithms for auto display range that don't just look at the data min and max.
But maybe we would just implement that in the `DisplayRangeProcessor` class anyway, in which is case it is not an issue.

cmeyer:
Yes, we can add that back when needed.

PR #1267, opened 2024-11-26
Add graphics capabilities to handle new features in  Inspector.py

lisham2000:


lisham2000:
@cmeyer Agree in hindsight keeping the float value as a float and not an optional means less downstream changes and means no PR is needed for nionUI. This downstream problem was only discovered when testing backwards compatibility with Kev when it had already been implemented and we discussed whether it was sensible to change and were on the fence about it so if you think going for a simpler solution is better for now will update and put through.

PR #1268, opened 2024-11-26
Inspector graphics update

lisham2000:


lisham2000:
Features implemented @Ion-e :

- Stroke Style
1. Line stroke color changer pallette capability into the inspector panel
2. Line stroke width input box
3. Line stroke visibility toggle (for graphics where only fill is wanted)

- Fill style
1. Fill color changer pallette capability into the inspector panel
3. Fill visibility toggle

Ion-e:
The "Show?" option may be better if it appears as "Show".
The point,line tools have a Fill option however it doesn't have a Fill functionality.
Do we want transparency as default? do we want a "opaque" tool?
There seems to be a lag whist using some of the tools, however I just redid the installation and I'm unsure if that may be related.
The line profile tool has a Stroke width and Fill colour options that don't have a functionality attached to them.

nion/swift/Inspector.py
```diff
@@ -2838,12 +2838,17 @@ def __init__(self, document_controller: DocumentController.DocumentController,
...
+        self._stroke_color_model = GraphicPropertyCommandModel(document_controller, display_item, graphic,"stroke_color", title=_("Change Stroke Color"), command_id="change_stroke_color")
+        self._stroke_width_model = GraphicPropertyCommandModel(document_controller, display_item, graphic, "stroke_width", title=_("Change Stroke Width"),command_id="change_stroke_width")
+        self._stroke_enabled_model = GraphicPropertyCommandModel(self.__document_controller, self.__display_item, graphic, "is_stroke_enabled", title=_(f"Change {self._graphic_type_model.value} Stroke Enabled"), command_id=f"change_{self._graphic_type_model.value}_stroke_enabled")
+        self._fill_color_model = GraphicPropertyCommandModel(document_controller, display_item, graphic,"fill_color", title=_("Change Fill Color"),command_id="change_fill_color")
+        self._fill_enabled_model = GraphicPropertyCommandModel(self.__document_controller, self.__display_item, graphic, "is_fill_enabled", title=_(f"Change {self._graphic_type_model.value} Fill Enabled"), command_id=f"change_{self._graphic_type_model.value}_fill_enabled")
```
KRLango, line 2851: Mixing self.__document_controller/document_controller and self.__display_item/display_item: I would suggest picking one and being consistent

nion/swift/Inspector.py
```diff
@@ -3194,6 +3202,35 @@ def __create_position_and_shape_ui(self) -> Declarative.UIDescriptionResult:
...
+                                   u.create_label(text=_("Fill"), width=40, text_alignment_vertical="center",text_alignment_horisontal= "left"),
+                                   u.create_spacing(15),
+                                   u.create_check_box(text=_("Show?"),checked="@binding(_fill_enabled_model.value)",text_alignment_vertical="center"), spacing=1),
+                                   u.create_row(u.create_spacing(3),
+                                   u.create_label(text=_("Fill Color"), width=80, text_alignment_vertical="center",text_alignment_horisontal= "right"), {"type": "nionswift.color_chooser", "color": "@binding(_fill_color_model.value)"},
+                                   u.create_line_edit(text="@binding(_fill_color_model.value)", placeholder_text="None", width=80),
+                                   u.create_stretch(),spacing=8),
+                                   )
+        else:
+            return self.__stroke_style_func()
```
KRLango, line 3233: Do all graphics have stroke and fill -> could we add something to specify what the graphic needs?

nion/swift/model/Graphics.py
```diff
@@ -456,8 +457,8 @@ def draw_marker(ctx: DrawingContextLike, p: Geometry.FloatPoint, is_enabled: boo
             ctx.fill()
 
 
-def draw_circular_marker(ctx: DrawingContextLike, p: Geometry.FloatPoint, is_enabled: bool) -> None:
-    marker_color = '#00FF00'
+def draw_circular_marker(ctx: DrawingContextLike, p: Geometry.FloatPoint, is_enabled: bool, fill_style: typing.Optional[str] = None) -> None:
+    marker_color = fill_style if fill_style else '#00FF00'
```
KRLango, line 461: Why is the fill style affecting the markers?

KRLango (changes requested):
A couple of points in the code. But I would also make some suggestions for the UI:

Separate Show/Colour controls - in other parts of Swift (e.g. Line Plot Display Layers), this is done by removing the colour (making it transparent) being how it is hidden - I think we should be consistent with this.

Default stroke colour shows up as None, but graphics are orange/pink. The inspector should reflect the correct defaults.

Do we really want the fill controlling the markers? That seem strange to me.

nion/swift/model/Graphics.py
```diff
@@ -7,6 +7,7 @@
 import gettext
 import math
 import uuid
+from faulthandler import is_enabled
```
Tiomat85, line 10: What is this for?

nion/swift/model/Graphics.py
```diff
@@ -485,12 +485,12 @@ def draw_ellipse_graphic(ctx: DrawingContextLike, center: Geometry.FloatPoint, s
...
-        ctx.line_width = 1
+        ctx.line_width = stroke_width or 1
         draw_ellipse(ctx, origin + size / 2, size, stroke_style, fill_style)
         ctx.begin_path()
-        ctx.move_to(center.x, rect.top + 10)
-        ctx.line_to(center.x, rect.top + 2)
-        draw_arrow(ctx, Geometry.FloatPoint(y=rect.top + 10, x=center.x), Geometry.FloatPoint(y=rect.top + 2, x=center.x), arrow_size=4)
+        ctx.move_to(center.x, rect.top + 10 + (.5 * int(stroke_width or 1)))
+        ctx.line_to(center.x, rect.top + 2 + (.5 * int(stroke_width or 1)))
+        draw_arrow(ctx, Geometry.FloatPoint(y=rect.top + 10 + (.5 * int(stroke_width or 1)), x=center.x), Geometry.FloatPoint(y=rect.top + 2 + (.5 * int(stroke_width or 1)), x=center.x), arrow_size=4)
```
Tiomat85, line 493: Could those stroke_width or 1's be replaced with using ctx.line_width that you have set to that?

Ion-e:
For testing:
- Make sure that Default orange/pink colours on graphic creation match in the inspector.

nion/swift/Inspector.py
lisham2000: This is inline with the other GraphicsPropertyCommandModels

nion/swift/Inspector.py
lisham2000: Changed inline with @Ion-e comment about simmilar

nion/swift/Inspector.py
```diff
@@ -573,6 +573,21 @@ def __property_changed_from_graphic(self, name: str) -> None:
...
+class FillVisibilityModel(Model.PropertyModel[str]):
+
+    def __init__(self, graphic_type: Model.PropertyModel[str]) -> None:
+        super().__init__()
+        self._graphic_type = graphic_type
+        self._fill_visible = self._compute_fill_visibility()
+
+    def _compute_fill_visibility(self) -> bool:
+        """ Determines if fill should be visible based on the graphic type """
+        disallowed_types = {"Line", "Point"}
```
KRLango, line 585: Should line profiles and intervals also be in this list?

nion/swift/Inspector.py
```diff
@@ -3207,31 +3221,34 @@ def __create_stroke_style_ui(self) -> Declarative.UIDescriptionResult:
...
-                                   u.create_spacing(15),
-                                   u.create_check_box(text=_("Show?"),checked="@binding(_fill_enabled_model.value)",text_alignment_vertical="center"), spacing=1),
-                                   u.create_row(u.create_spacing(3),
-                                   u.create_label(text=_("Fill Color"), width=80, text_alignment_vertical="center",text_alignment_horisontal= "right"), {"type": "nionswift.color_chooser", "color": "@binding(_fill_color_model.value)"},
-                                   u.create_line_edit(text="@binding(_fill_color_model.value)", placeholder_text="None", width=80),
-                                   u.create_stretch(),spacing=8),
-                                   )
+u.create_spacing(10),
+                u.create_column(
+                u.create_label(text=_("Stroke"), width=40, text_alignment_vertical="left", text_alignment="left"),
```
KRLango, line 3226: text_alignment_vertical="left" - that doesn't make sense

nion/swift/Inspector.py
```diff
@@ -3207,31 +3221,34 @@ def __create_stroke_style_ui(self) -> Declarative.UIDescriptionResult:
...
+                u.create_stretch()),
+                u.create_row(u.create_spacing(15),
+                u.create_label(text=_("Stroke Width"), width=68, text_alignment_vertical="center"),
+                u.create_spacing(6),
+                u.create_line_edit(text="@binding(_stroke_width_model.value, converter=float_str_converter)", placeholder_text="1", width=80),
+                u.create_stretch())),
+                u.create_spacing(20),
+                u.create_column(
+        u.create_row(
+        u.create_label(text=_("Fill"), text_alignment_vertical="center",text_align="left"),
```
KRLango, line 3240: Stroke label has `text_alignment="left"`, Fill label has `text_align="left"` is there a reason these are different?

nion/swift/model/Graphics.py
```diff
@@ -770,6 +783,8 @@ def role(self) -> typing.Optional[str]:
     @role.setter
     def role(self, value: typing.Optional[str]) -> None:
         self._set_persistent_property_value("role", value)
+        self.stroke_color = self.get_default_stroke_color()
```
KRLango, line 770: I think a comment about why you are resetting to defaults would be useful here

KRLango (changes requested):
Looking good, a few little bits left, but this is almost there I think.

Ion-e:
- Graphics with a level of transparency revert to black:
image
- Consider whether we want to give transparency to the stroke, is this something we actually want to add?
- Creating new line profiles reverts the default to "F80" orange.

Ion-e:
Line profile shows Stroke width.

Ion-e (commented):
LGTM

nion/swift/model/Graphics.py
```diff
@@ -860,6 +882,13 @@ def used_fill_style(self) -> typing.Optional[str]:
             return "rgba(0, 0, 255, 0.1)"
         return None
 
+    @used_fill_style.setter
```
cmeyer, line 885: Same as above... `used_fill_style` is a computed value and should not have a setter.

nion/swift/model/Graphics.py
```diff
@@ -716,6 +717,15 @@ def stroke_color(self) -> typing.Optional[str]:
...
         self._set_persistent_property_value("stroke_color", value)
+        self.notify_property_changed("used_stroke_style")
+
+    @property
+    def stroke_width(self) -> typing.Optional[float]:
+        return typing.cast(typing.Optional[float], self._get_persistent_property_value("stroke_width"))
+
+    @stroke_width.setter
+    def stroke_width(self, value: typing.Optional[float]) -> None:
+        self._set_persistent_property_value("stroke_width", value)
```
cmeyer, line 728: For symmetry of code, here should probably be a `used_stroke_width` property, too. `used_stroke_width` would be a computed value and would always be `float`, never `None`.

Then all of the places where you do `self.stroke_width or 1` can be replaced with `self.used_stroke_style`.

Along with this, you can probably change the return values of `used_stroke_style` and `used_fill_style` to be `str` instead of `str | None`.

cmeyer (changes requested):
Getting closer! Main comment is to eliminate setters for `used_stroke_color` and `used_fill_color`. We should (1) remove the setters, read comments below; (2) add `used_stroke_width`; (3) make all three `used_xxx` functions return a non-None value, i.e. make return type `-> str`.

nion/swift/Inspector.py
```diff
@@ -2871,12 +2871,17 @@ def __init__(self, document_controller: DocumentController.DocumentController,
...
-
+        self.float_str_converter = Converter.FloatToStringConverter()
         self._graphic_type_model = Model.PropertyModel[str]()
         self.__set_type_specifics()
-
+        self.__stroke_style_func = None
         self._graphic_label_model = GraphicPropertyCommandModel(document_controller, display_item, graphic, "label", title=_("Change Label"), command_id="change_label")
         self._lock_position_model = GraphicPropertyCommandModel(self.__document_controller, self.__display_item, graphic, "is_position_locked", title=_(f"Change {self._graphic_type_model.value} Position Locked"), command_id=f"change_{self._graphic_type_model.value}_position_locked")
+        self._stroke_color_model = GraphicPropertyCommandModel(document_controller, display_item, graphic,"stroke_color", title=_("Change Stroke Color"), command_id="change_stroke_color")
+        self._used_stroke_color_model = GraphicPropertyCommandModel(document_controller, display_item, graphic,"used_stroke_style", title=_("Change Stroke Color"), command_id="change_stroke_color")
```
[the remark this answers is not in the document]
lisham2000: Hi Chris, I have implemented all but one of these changes with the last one being the removal of the optional on fill style. This one I questioned as I think a user would want the default to to be No fill hence the end case of used_fill_style being the return of None. We could do this with a pass of transparent or a colour with an alpha channel of 0 but that wouldn't appear as transparent in the inspector and in both cases if you want to change the color using the color picker you have to manually change the alpha channel up from 0 which seems very unintuitive from a UI standpoint. I wanted to get your opinion on whether the optional nature of this is still a problem if there is genuine use to it?

cmeyer:
Merged with additional changes:
- I separated the Graphics.py [changes](https://github.com/nion-software/nionswift/commit/7d1be9726fa77caf1304cc80dd2ab00058a18d5a) from the Inspector.py [changes](https://github.com/nion-software/nionswift/commit/ec7b55bf8128cf285635b22bc1eee5e2e6202829) so that there were two clean commits.
- I [updated the layout](https://github.com/nion-software/nionswift/commit/4846418c3468f319c5151db46213af03d15adf18) so it looks better. See screen shots below. Also update line plot inspector.
- I [fixed issue](https://github.com/nion-software/nionswift/commit/6392498fe2daaf1a43beba1cbf4e4e60253aaea0) when clearing stroke width field - it was setting it to 0 instead of default.

Note about layout: setting the label horizontal or vertical alignment to "center" will center BOTH and subsequently setting horizontal alignment to "left" will fail; instead vertical alignment and horizontal alignment should use "vcenter" and "hcenter" respectively. This is a historical quirk for compatibility with Qt.

New layout:

image

Old layout:

image

PR #1290, opened 2024-12-16
Add verify of Project File Name on text_changed in New Project window

Tiomat85:
Added explicit sanity check of Project Name when text changed on the field.  Verifies compatibility using Utility.simplify_filename which detects any non-compatible filenames and returns a sanitised version so if they are different it means it was not compatible. Currently the identification for the user is some orange background flair and a tooltip on the Create button.

### DIFF
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -45,6 +45,7 @@ from nion.swift.model import FileStorageSystem
 from nion.swift.model import PlugInManager
 from nion.swift.model import Profile
 from nion.swift.model import Symbolic
+from nion.swift.model import Utility
 from nion.ui import Application as UIApplication
 from nion.ui import CanvasItem
 from nion.ui import Declarative
@@ -899,6 +900,19 @@ class NewProjectAction(UIWindow.Action):
                     safe_request_close()
                     return True
 
+                def verify_project_name(text: str) -> None:
+                    valid = text == Utility.simplify_filename(text)
+                    if valid:
+                        create_project_button.enabled = True
+                        create_project_button.tool_tip = None
+                        create_project_button.background_color = None
+                        project_name_field.background_color = None
+                    else:
+                        create_project_button.enabled = False
+                        create_project_button.background_color = "orange"
+                        create_project_button.tool_tip = "Invalid Characters in Project Name"
+                        project_name_field.background_color = "orange"
+
                 column = self.ui.create_column_widget()
 
                 directory_header_row = self.ui.create_row_widget()
@@ -933,6 +947,7 @@ class NewProjectAction(UIWindow.Action):
                 project_name_field.text = self.project_name
                 project_name_field.on_return_pressed = handle_new_and_close
                 project_name_field.on_escape_pressed = safe_request_close
+                project_name_field.on_text_edited = verify_project_name
                 project_name_row.add(project_name_field)
                 project_name_row.add_stretch()
                 project_name_row.add_spacing(13)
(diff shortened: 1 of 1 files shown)


KRLango (approved):
Not sure about the orange - mainly because I don't think that pattern is anywhere else in Swift. But the code looks good to me.

Tiomat85:
> Not sure about the orange - mainly because I don't think that pattern is anywhere else in Swift. But the code looks good to me.

I didn't find any common way that those type of errors are displayed so picked something obvious to highlight something needs correcting.  If a new pattern is picked we can go with that.

cmeyer:
Setting background colors triggers Qt to drop back to their own rendering for controls - so I like to avoid it. I have an alternate suggestion: just a status line and disable the 'create' button.

Any objections?

```
diff --git a/nion/swift/Application.py b/nion/swift/Application.py
index 3cc595ab..98a869ba 100644
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -906,13 +906,12 @@ class NewProjectAction(UIWindow.Action):
                     if valid:
                         create_project_button.enabled = True
                         create_project_button.tool_tip = None
-                        create_project_button.background_color = None
-                        project_name_field.background_color = None
+                        project_name_status_label.text = None
                     else:
                         create_project_button.enabled = False
-                        create_project_button.background_color = "orange"
-                        create_project_button.tool_tip = "Invalid Characters in Project Name"
-                        project_name_field.background_color = "orange"
+                        create_project_button.tool_tip = _("Invalid Characters in Project Name")
+                        project_name_status_label.text = _("Invalid Characters in Project Name")
+                        project_name_status_label.text_color = "red"
 
                 column = self.ui.create_column_widget()
 
@@ -953,6 +952,13 @@ class NewProjectAction(UIWindow.Action):
                 project_name_row.add_stretch()
                 project_name_row.add_spacing(13)
 
+                project_name_status_label = self.ui.create_label_widget()
+                project_name_status_row = self.ui.create_row_widget()
+                project_name_status_row.add_spacing(26)
+                project_name_status_row.add(project_name_status_label)
+                project_name_status_row.add_stretch()
+                project_name_status_row.add_spacing(13)
+
                 column.add_spacing(12)
                 column.add(directory_header_row)
                 column.add_spacing(8)
@@ -963,6 +969,8 @@ class NewProjectAction(UIWindow.Action):
                 column.add(project_name_header_row)
                 column.add_spacing(8)
                 column.add(project_name_row)
+                column.add_spacing(4)
+                column.add(project_name_status_row)
                 column.add_stretch()
                 column.add_spacing(16)
```

New version:

image

Old version (on macOS):

image

Tiomat85:
I have no problem with removing the orange, as it very much didn't feel like a great solution anyway and was hoping for peoples opinions.

With your solution though there is a problem on Windows - The Create Project button, when set to disabled, doesn't look disabled.  It was that bit that led me to trying to find alternatives such as setting the background colour myself.

image

cmeyer:
For some reason this came up on my notifications today. Whatever the reason, let's wait until the upcoming Qt 6.9 and see how that affects things. This 'disabled not showing in Windows 11' issue in Qt is annoying. More investigation needed for something that should be a no-brainer! Maybe I'm missing something on the Qt side of the code.

cmeyer:
Merged version w/ button disable, which is now handled correctly because we told Qt not to use it's broken W11 style.

50258f07ffd8099f1ce86b9c125c6c40672f3f17

Issue #961, opened 2023-06-20
Line-type graphics should resize from center when position locked or option-key is down


cmeyer:
See also:
- #105

Here's a starter test for this issue:

LINE-GRAPHIC-1. Test line graphic dragging.
-------------------------------------------
1. Create a raster display.
2. Using the line tool, drag within the bounds of the display.
  2.1. The line is created and visible with endpoints at the start and end of the drag.
3. Drag one end of the line. Then drag the other end of the line.
  3.1. The line endpoint is updated to the drag position.
4. Drag the ends of the line and press and release the shift key while dragging.
  4.1. The line endpoint is limited to a multiple of 45° while shift key is depressed.
5. Drag the ends of the line and press and release the alt/option key while dragging.
  5.1. The line endpoint is adjusted from the center as it was at the start of the drag while alt/option key is depressed.
6. Drag the middle of the line.
  6.1. The line moves without changing shape.

Ion-e:
@cmeyer Sorry could you elaborate on which graphics do you consider "line-type", and what do you mean by "center"? :)

Ion-e:
Answer: normally graphics are resized from the opposite end of the drag

cmeyer:
For rectangle-likes:

On macOS it's just 'option' to keep center constant (or the reverse in case of ellipse) and the 'shift' key keeps it square.

On Linux/Windows it's 'alt' to keep center constant (or the reverse in case of ellipse) and 'shift' to keep it square.

I propose to make `.alt` keep the line centered.

The `.alt` modifier maps to 'option' on macOS and 'alt' on Windows. So it's the proper method to use.

The logic for the rectangle is here: https://github.com/nion-software/nionswift/blob/d5f747ba6382792ed5d7e4e2e435fdb69b7037fe/nion/swift/model/Graphics.py#L204

Ion-e:
Verified as implemented. Thanks!

Closed as per comment: https://github.com/nion-software/nionswift/issues/1311#issuecomment-2598752676

PR #1298, opened 2025-01-02
Added Alt/Option key binding to lines to extend/pivot from center

Tiomat85:
Issue https://github.com/nion-software/nionswift/issues/961

Alt/Option mode when dragging the ends of a line control. Added a defined pivot point in the code to use as the rotation centre and refactored accordingly.

### DIFF
--- a/nion/swift/model/Graphics.py
+++ b/nion/swift/model/Graphics.py
@@ -1248,6 +1248,7 @@ class LineTypeGraphic(Graphic):
     def __init__(self, type: str, title: typing.Optional[str]) -> None:
         super().__init__(type)
         self.title = title
+        self.__old_pivot: typing.Optional[Geometry.FloatPoint] = None
 
         def read_vector(persistent_property: Persistence.PersistentProperty, properties: Persistence.PersistentDictType) -> typing.Any:
             # read the vector defined by persistent_property from the properties dict.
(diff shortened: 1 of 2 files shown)


Ion-e:
https://github.com/nion-software/nionswift/issues/961

cmeyer:
Can you add a test case for this in `test_dragging_regions` in `Graphics_text.py`? This is a big messy method, but should provide a good starting point.

Tiomat85:
> Can you add a test case for this in `test_dragging_regions` in `Graphics_text.py`? This is a big messy method, but should provide a good starting point.

I was in the middle of writing those test cases while waiting for Kev/Luke to get a chance to check over the code.  I have done them in a new method instead, test_line_dragging. Currently it has 4 tests added, dragging both ends with no modifiers and then just dragging the start with the alt and shift modifiers.

KRLango (approved):
The changes look good to me, but as a more general comment, there is a lot of duplicated code between the `part[0] == "start"` and `part[0] == "end"` cases - primarily the shift modifier handling - that existed prior to this work. This could be pulled out into a function to remove this duplication.

cmeyer (changes requested):
This overall looks nice - thanks!

However, I'd request one change to make it behave more consistently: the end point / center point should not change after the first click. Right now if you start dragging with the alt-key, then let go of the alt-key, it uses the "start" point at the time you let go of the alt-key rather than the original start point. I think it should use the original. Right now, if you press/release the alt key while dragging, you can make the line walk around the image and it's impossible to know even where it started.

Making it remember the first start point and first center point will cause some jumping around when you press and release the alt key, but I think this is a better way to do it since the user may want to try the alt-key and see what happens.

Making this change would make it work like the rectangle resizing works, too. You can see how it works by experimenting with the rectangle tool.

Tiomat85:
Updated it based on that.  It feels a little weird behaviourally to have the line 'jump' when you release alt but it does match the rectangle and some other pieces of software so it seems that weirdness is the standard.  Also used this as an excuse to address Kev's point on the almost-identical sections for the different ends of the line.

cmeyer:
Squashed, added `changes.json` entry, and merged here 3f900f9309f051e34d7cc701088761683b843ad8

Additional modification here b74b80f19e538af6674f2e89ace75cf2147d0dbc

PR #1315, opened 2025-01-23
copy paste line profile

### LINKED ISSUE #1238: Add ability to copy and paste line profiles from one display to another
The user should be able to create a line profile or pick on one data item, copy the line profile or pick region, and paste it to another data item display (or the same one). Pasting should create a new line profile/pick computation on the new data item. Undo should work. Redo should work.

Essentially something like this code below.

```
diff --git a/nion/swift/DisplayPanel.py b/nion/swift/DisplayPanel.py
index 6534aa6b..28df3bdb 100644
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -1187,6 +1187,10 @@ class InsertGraphicsCommand(Undo.UndoableCommand):
                 display_item.add_graphic(graphic)
                 new_graphic = display_item.graphics[-1]
                 self.__graphic_proxies.append(new_graphic.create_proxy())
+                if isinstance(graphic, Graphics.LineProfileGraphic):
+                    data_item = display_item.data_item
+                    if data_item:
+                        self.__document_controller.document_model.get_line_profile_new(display_item, data_item, None, new_graphic)
             self.__graphics = typing.cast(typing.Any, None)
 
     def _get_modified_state(self) -> typing.Any:
```

cmeyer:
- **Add graphic attribute testing to replace isinstance use.**
- **Fix #1238. Add line profile automatically computation when pasting line profile graphic.**

Note: I inadvertently included a maintenance commit (graphic attribute testing). Both are reviewable. I plan to merge these Jan 24.

### DIFF
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -1157,7 +1157,8 @@ class InsertGraphicsCommand(Undo.UndoableCommand):
     def __init__(self, document_controller: DocumentController.DocumentController,
                  display_item: DisplayItem.DisplayItem, graphics: typing.Sequence[Graphics.Graphic], *,
                  existing_graphics: typing.Optional[typing.Sequence[Graphics.Graphic]] = None,
-                 display_item_modified_state: typing.Any = None) -> None:
+                 display_item_modified_state: typing.Any = None,
+                 allow_secondary_copy: bool = False) -> None:
         super().__init__(_("Insert Graphics"))
         self.__document_controller = document_controller
         self.__display_item_proxy = display_item.create_proxy()
@@ -1165,6 +1166,7 @@ class InsertGraphicsCommand(Undo.UndoableCommand):
         self.__graphics_properties = None
         self.__graphic_proxies = [graphic.create_proxy() for graphic in existing_graphics or list()]
         self.__undelete_logs: typing.List[Changes.UndeleteLog] = list()
+        self.__allow_secondary_copy = allow_secondary_copy
         self.initialize(display_item_modified_state)
 
     def close(self) -> None:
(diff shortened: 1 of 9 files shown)


Ion-e:
A couple of points here. I may have gotten them incorrectly though.

1.Expected: Copying "Pick" tool to a new data item should make the pick average 1D line graph to be created for the new data item. 
   Actual: Currently it doesn't.

2. Expected: Copying a interval from the pick average of a 2D raster display of 1D EELS spectra and pasting to a different pick average should make either of the 2d raster display of 1D EELS to be updated. 
   Actual: Currently the graphic is copied but not the "interval" tool. So it's only a visual copy with no computations underneath. They do show changes in the histogram correctly.

3. Expected: intervals shouldn't be able to be copied into 2D raster displays, and "rectangles" shouldn't be able to be copied to 1D line profiles.
    Actual: they are able to be copied across, they won't show in the display however they can be observed in the inspector.

cmeyer:
Good point about the pick tool. I decided to only address the line profile part of the original issue with this commit, since this was the immediate request from Ben. I will split the original issue into two parts and follow-up on the pick tools separately - they need a bit of a redesign, as mentioned yesterday during our chat, to work nicely in a copy/paste workflow.

Ion-e:
Sorry, I am a bit confused about the expected behaviour of the changes, would you mind elaborating the scope of this ticket?

Ion-e:
This should be limited to the copy/paste of standard line profiles.

Ion-e:
Verified as implemented, thanks!

Ion-e:
I'm removing this from my list as it looks good. However I would also consider if we would want to add the Subscan tool to be affected by the changes in this ticket. (Which I haven't tested yet)

PR #1333, opened 2025-02-04
Fix #1331. Draw graphics in alternate style when unfocused.

### LINKED ISSUE #1331: Selected graphics should draw selection markers as disabled when display panel does not have focus
Right now they draw bright green whether focused or not, making it difficult to quickly see what is focused.

cmeyer:
This PR makes unfocused graphics draw in a different style.


Ion-e:
what does "drawing in alternate style" means in this scenario?

Ion-e:
Graphics should show visually they are unselected (grey) fromselected (bright green)

Issue #1337, opened 2025-02-18
Add titles to items in grid browser


cmeyer:
Use titles to help distinguish items in the grid browser.

This will require switching to canvas item drawing, which will also improve performance and is a precursor to #297.

PR #1338, opened 2025-02-18
New grid browser with better performance

cmeyer:
PR for #1337. This switches the grid browsers in the data panel and display panel to use a new optimized canvas based layout. Also reworks the strip browser. The latter two are accessible using 'v' key in a display panel. The former is accessed using the icons at the bottom of the data panel.

Testing: the functionality should not be changed from previous version. Performance should be the same or better. And there should not be drawing anomalies.

There is a known issue with focus which I will submit a PR for separately. Click on another display panel. Then on the strip browser. The inspector doesn't respond to the clicks.

- **Rewrite data panel to use grid canvas item 2.**
- **Rewrite display panel browser to use grid canvas item 2.**
- **Rewrite strip browser to use list canvas item 2.**
- **Remove unused grid/item controller, display adapters, etc.**
- **Update release notes. Fix typos.**

This requires https://github.com/nion-software/nionui/pull/98 in `nionui` which I already merged since it was mainly a refactoring, but can still be reviewed and patched.


Ion-e:
This looks good. The only odd behaviour I noticed is that if you delete an item from the thumbnail/gridbrowser views, then the display panel will go blank, which may be a bit counter intuitive but I can see why it's set up this way. Give me a shout if it's bug worthy @cmeyer

Ion-e (commented):
LGTM

PR #1342, opened 2025-02-19
2025 02 19 command logger

cmeyer:
- **Modify undo commands to override _perform instead of perform.**
- **Add minimal command logger for debugging crashes.**

### DIFF
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -2,11 +2,13 @@ from __future__ import annotations
 
 # standard libraries
 import asyncio
+import dataclasses
 import datetime
 import functools
 import gettext
 import json
 import logging
+import logging.handlers
 import operator
 import os
 import pathlib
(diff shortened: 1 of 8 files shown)


nion/swift/Application.py
```diff
@@ -237,16 +239,55 @@ def __init__(self, ui: UserInterface.UserInterface, set_global: bool = True) ->
...
+                time_delta: datetime.timedelta
+
+            if log_dir_path.exists():
+                log_files = list[LogFile]()
+                now = datetime.datetime.now()
+                for file_path in log_dir_path.rglob("commands*.log*"):
+                    log_files.append(LogFile(file_path, now - datetime.datetime.fromtimestamp(file_path.stat().st_mtime)))
+                log_files.sort(key=operator.attrgetter("time_delta"))
+                for index, log_file in enumerate(log_files):
+                    if index > 100 or log_file.time_delta.days > 30:
```
Tiomat85, line 255: Is this intended to be >= 100 (keeping 100 files max)?

nion/swift/Application.py
```diff
@@ -237,16 +239,55 @@ def __init__(self, ui: UserInterface.UserInterface, set_global: bool = True) ->
...
+            now_slug = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
+            log_dir_path = pathlib.Path(self.ui.get_data_location()) / "Logs"
+            log_dir_path.mkdir(parents=True, exist_ok=True)
+            log_base_path = log_dir_path / f"commands_{now_slug}.log"
+            log_formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(threadName)s - %(message)s')
+            rotating_file_handler = logging.handlers.RotatingFileHandler(log_base_path, backupCount=100)
+            rotating_file_handler.setFormatter(log_formatter)
+            commands_logger = logging.getLogger("_commands")
+            commands_logger.propagate = False
+            commands_logger.setLevel(logging.DEBUG)
```
Tiomat85, line 275: Is there a config file anywhere that you can load this information from to not need code changing to enable/disable this?

nion/swift/Application.py
```diff
@@ -288,6 +287,7 @@ def initialize(self, *, load_plug_ins: bool = True, use_root_dir: bool = True) -
             logging.info("Qt version " + self.ui.get_qt_version())
             logging.info("Log files " + str(log_dir_path))
             self.__purge_log_files(log_dir_path)
+            logging.info("Log file for this launch " + str(log_file_path))
```
Tiomat85, line 290: If this is trying to be a list of log files generated, it is probably worth including the log level that is running at as well so you don't assume an empty log file is fine when it could just be the wrong logging level.

Tiomat85 (commented):
Nothing major, just a few little tweaks for you to consider.

nion/swift/Application.py
```diff
@@ -237,16 +239,55 @@ def __init__(self, ui: UserInterface.UserInterface, set_global: bool = True) ->
...
+                now = datetime.datetime.now()
+                for file_path in log_dir_path.rglob("commands*.log*"):
+                    log_files.append(LogFile(file_path, now - datetime.datetime.fromtimestamp(file_path.stat().st_mtime)))
+                log_files.sort(key=operator.attrgetter("time_delta"))
+                for index, log_file in enumerate(log_files):
+                    if index > 100 or log_file.time_delta.days > 30:
+                        logging.info(f"Purging log file {log_file.absolute_file_path}")
+                        log_file.absolute_file_path.unlink()
+        except Exception as e:
+            pass
```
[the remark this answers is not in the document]
cmeyer: Maybe... the reason I say "maybe" is I'm not sure what to put here. This is intention is that if something fails, i.e. directory access, or a file open, it doesn't prevent launching. Launching will presumably emit more detailed messages if things continue to be bad.

I think maybe a "Failed to purge all log files." would be sufficient, without a stack track. If that message appears on multiple launches with no additional further-into-launch error messages, it will require more investigation to resolve.

nion/swift/Application.py
cmeyer: Yes I'll change it, although it is a very arbitrary number that is only intended to prevent an unlimited number of files from filling up the disk.

nion/swift/Application.py
cmeyer: There are a lot of places in the code that could use a config file like this, i.e. loading plug-ins. But I'll put that off to a future version. Adding a config file just adds another point of configuration failure. Is there a case where you think you'll want to disable the logger on a regular basis? It's already "never enabled" during automated tests.

Tiomat85: I was thinking of it more from the opposite angle.  For a release presumably most of the logging should be defaulting to Warning or above, then if someone raises a potential issue with a released version being able to change that to Debug or Info without needing a code change would presumably be helpful.

nion/swift/Application.py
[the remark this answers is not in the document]
Tiomat85: That should be enough, just something to indicate that something has gone wrong tidying up otherwise a problem there would just linger until someone notices the folder has a thousand files in it.  If it becomes a common problem then as you say we can always extend.

nion/swift/Application.py
cmeyer: I left the level at DEBUG since this isn't user facing. We can reconsider if it becomes a problem.

cmeyer:
@Tiomat85 thanks for the review - very useful. I'm moving fast on this to get it on the local machines today. see also 3599aec2153e63836dffaed51644d9ffe09b26ce

Ion-e:
Where is this logged information saved?

Ion-e:
On Swift Startup a log file: filepath should be there. It's also recorded as an output.

Ion-e:
Looks good to me!

PR #1347, opened 2025-02-26
Record computation description metadata when data is computed.

cmeyer:


KRLango (approved):
LGTM

cmeyer:
I'm going to hold off on this until next week - I want to have discussion w/ @Brow71189 as to the purpose of recording this information.

On one hand it seems obvious that we want to record what computation was run; but in a project where the computation is active, it is easily derived from internal objects. So that begs the question: is this just a lack of a UI for internal objects? I don't think the metadata viewer is a good UI for this.

When exporting to a external file format, including this information somewhere is more sensible, but (a) can we generate it on-the-fly when exporting? (b) do we need to increase the information stored in 'specifiers' to be able to locate the project in which the original data is located?

This PR is serving as a security blanket, too, since I have removed the recording of this information from the relevant computations in nion-software/experimental#48.

nion/swift/model/Symbolic.py
```diff
@@ -2624,6 +2627,66 @@ def recompute(self) -> typing.Optional[ComputationExecutor]:
...
+                                base_items_l.append(item_d)
+                                if isinstance(base_item, DataItem.DataItem):
+                                    item_d["type"] = "data-item"
+                                    item_d["uuid"] = str(base_item.uuid)
+                                    if base_item.data_shape:
+                                        item_d["shape"] = base_item.data_shape
+                                    if base_item.data_dtype:
+                                        item_d["data_type"] = numpy.dtype(base_item.data_dtype).name
+                                    if base_item.data_metadata:
+                                        item_d[
```
Tiomat85, line 2674: Minor thing, is it worth extracting base_item.data_metadata.data_descriptor into a variable here just to make the following line a bit more legible?

Tiomat85 (approved):
Looks fine to me, one tiny formatting tweak just for readability to consider.

Brow71189:
Since I don't see it in the code, will this finally copy the metadata of the input data item into the result? This was one of the main reasons for explicit metadata handling in experimental, so it would be important to handle this here.

Brow71189:
That looks good to me now. I chained a few computations together and the `source_metadata` got propagated nicely. Recording the last run computation is a nice bonus and a good start towards recording everything what has been done to a piece of data.

The most important thing is copying the source metadata though, so since that is happening now I'm happy with this.

nion/swift/model/Symbolic.py
```diff
@@ -2607,6 +2616,277 @@ def recompute(self) -> typing.Optional[ComputationExecutor]:
...
+                graphic_d["radius_1"] = graphic.radius_1
+                graphic_d["radius_2"] = graphic.radius_2
+                graphic_d["mode"] = graphic.mode
+            if isinstance(graphic, Graphics.LatticeGraphic):
+                graphic_d["u_pos"] = make_size_description(graphic.u_pos)
+                graphic_d["v_pos"] = make_size_description(graphic.v_pos)
+                # future change will make the cell be defined by an ellipse size and rotation
+                graphic_d["size"] = make_size_description(Geometry.FloatSize(graphic.radius, graphic.radius))
+                graphic_d["rotation"] = 0.0
+            return graphic_d
```
Tiomat85, line 2724: Would these be better done in an OO way. So rather than having this function define what metadata each graphic type needs to store, the graphic class itself determines them? That way you could remove the big ininstance chains.

I would imagine something like:
graphic_d = graphic_d | graphic.get_description()

nion/swift/model/Symbolic.py
```diff
@@ -2607,6 +2616,277 @@ def recompute(self) -> typing.Optional[ComputationExecutor]:
...
+                if variable_bound_item._data_structure:
+                    parameter_d["data-structure"] = make_data_structure_description(variable_bound_item._data_structure)
+                if variable_bound_item._property_name:
+                    parameter_d["property-name"] = variable_bound_item._property_name
+            if isinstance(variable_bound_item, BoundGraphic):
+                if variable_bound_item._graphic:
+                    parameter_d["graphic"] = make_graphic_description(variable_bound_item._graphic)
+                if variable_bound_item._property_name:
+                    parameter_d["property-name"] = variable_bound_item._property_name
+            return parameter_d
```
Tiomat85, line 2769: Same as previous one.

nion/swift/model/Symbolic.py
```diff
@@ -2607,6 +2616,277 @@ def recompute(self) -> typing.Optional[ComputationExecutor]:
...
+                    computation_parameters = list[ComputationRecordParameter]()
+                    for variable in self.variables:
+                        variable_specifier = variable.specifier
+                        variable_bound_item = variable.bound_item
+                        if variable_specifier and variable_bound_item:
+                            if isinstance(variable_bound_item, BoundList):
+                                list_items = list[ComputationRecordObject]()
+                                bounds_items = variable_bound_item.get_items()
+                                specifiers = variable.object_specifiers
+                                if len(bounds_items) == len(specifiers):
```
Tiomat85, line 2857: What should happen if they don't match? Presumably that is an error scenario but it gets silently ignored.

cmeyer: It gets silently ignored. Open to suggestions.

nion/swift/model/Symbolic.py
cmeyer: Same comment. Could be improved.

PR #1349, opened 2025-02-26
Improve API to avoid accessing internal objects

cmeyer:
- **Add computation.get_input for use in plug-in computations.**
- **Add structure_type to Facade.DataStructure for use in plug-ins.**
- **Add methods to access data item and graphic inputs.**


KRLango (approved):
LGTM

PR #1357, opened 2025-03-06
Pass only read-only parameters to computation execute

cmeyer:
- **Introduce Region objects to represent Graphics for computations.**
- **Introduce computation_value to bound item and pass it to computation instead of value.**
- **Pass Regions instead of Graphics to computations.**
- **Pass DataItemComputationValue instead of DataItem to computations.**
- **Pass DataStructureComputationValue to computations instead of directly passing DataStructure.**
- **Add computation_value to BoundList.**
- **Always pass data source to computation executor instead of passing data directly.**
- **Ensure computations are run in tests. Also additional test cleanup.**

This changes parameter passing to computations to only pass read-only parameters. This also changes parameters that used to pass xdata directly to computations to now pass data sources. Graphics and data structures are no longer passed directly, but instead parameters with access to the model part of the graphic or data structure is passed. This change also improves tests to be more robust. All of these changes are preparation for threaded computations.

This change introduces Region objects to represent the Graphics when passed to computations. These may be changed in the future to be more computation specific; but they should be considered to be a stepping stone useful for the implementation of threaded computations (upcoming) for now.

### DIFF
--- a/nion/swift/model/Graphics.py
+++ b/nion/swift/model/Graphics.py
@@ -868,6 +868,9 @@ class Graphic(Persistence.PersistentObject):
     def test(self, mapping: CoordinateMappingLike, ui_settings: UISettings.UISettings, p: Geometry.FloatPoint, move_only: bool) -> typing.Tuple[typing.Optional[str], bool]:
         raise NotImplementedError()
 
+    def get_region(self) -> RegionBase:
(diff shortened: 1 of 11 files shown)


Brow71189:
Looks fine to me

Tiomat85 (approved):

PR #1359, opened 2025-03-10
Fix 217. Use data item names for line plot layer names by default.

cmeyer:
- **Refactor displays to take display layer as object rather than dict.**
- **Pass display layer info to display canvas items.**
- **Use display data channel label if layer has no label.**

The first two commits are refactoring precursor to the last commit which makes line plot layers default to the data item name.

There are many edge cases and I expect to have a few follow-up PR's with bug fixes that get revealed during testing and/or features that go along with this change.

However, I think it's worthwhile to get this PR merged first to see what else needs to be fixed after testing and use.


Brow71189:
Looks good to me.

Ion-e:
The inspector's "Data layer Name" doesn't seem to populate correctly. 
image

cmeyer:
@Ion-e please see follow-up PR https://github.com/nion-software/nionswift/pull/1360

PR #1360, opened 2025-03-10
fix 217 add data item placeholder to inspector

cmeyer:
- **Add a value model utility class to Observer tools.**
- **Show data item title as placeholder for empty line plot layer label in inspector.**


Brow71189:
LGTM

PR #1369, opened 2025-03-14
fix 1368 handle data panel focus

cmeyer:
- **Fix data panel focus issue.**
- **Fix problem where workspace split actions were enabled with no display panel selected.**

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -3641,7 +3641,11 @@ class WorkspaceSplitAction(Window.Action):
 
     def is_enabled(self, context: Window.ActionContext) -> bool:
         context = typing.cast(DocumentController.ActionContext, context)
-        return context.display_panel is not None
+        window = typing.cast(DocumentController, context.window)
+        workspace_controller = window.workspace_controller
+        display_panels = context.display_panels
+        selected_display_panel = context.selected_display_panel
+        return bool(workspace_controller and display_panels and selected_display_panel)
 
 
 class WorkspaceSplitHorizontalAction(WorkspaceSplitAction):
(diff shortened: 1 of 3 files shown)


KRLango (commented):
I cannot see any red flags in the code. 

When testing it the only odd behaviour I hit was that if you click on one the little "parent/child" thumbnails at that bottom of a display panel, it is not clear which item and processing will occur on. As far as I can tell it is whichever one of the data panel items or display panels was last selected, but now the only item that shows focus is the thumbnail.

nion/swift/DocumentController.py
```diff
@@ -3641,7 +3641,11 @@ def execute(self, context: Window.ActionContext) -> Window.ActionResult:
 
     def is_enabled(self, context: Window.ActionContext) -> bool:
         context = typing.cast(DocumentController.ActionContext, context)
-        return context.display_panel is not None
+        window = typing.cast(DocumentController, context.window)
+        workspace_controller = window.workspace_controller
+        display_panels = context.display_panels
+        selected_display_panel = context.selected_display_panel
+        return bool(workspace_controller and display_panels and selected_display_panel)
```
Tiomat85, line 3648: Can you have a situation where no display_panel is selected? (selected_display_panel is None)?

Tiomat85 (approved):

Ion-e:
Comment on 1368 asking for test cases for this PR. Thanks!

nion/swift/DocumentController.py
cmeyer: Yes - if the user clicks in the data panel there will be no display panel selected.

Ion-e:
There's an issue here. 

1 - have a project with at least two data items and one of those data items in a display panel , *Add a graphic on item B*
2 - click on item A in the display panel
3 - click on item B in the data panel
Actual: the graphic of item B appears in the inspector.
Expected: the inspector of item B should appear (including it's graphic section)

Issue #1378, opened 2025-03-17
A new item in data panel when first item already selected does not respond to mouse clicks


Ion-e:
When using the "Generate Data" tool, importing data, and potentially in acquisitions:

The top items (recently created items) in the data panel seem unusable the first few seconds after using these tools. (It may be possible that the size of the items created/imported may play a role in this behaviour).

Ion-e:
It seems to happen in to the first item in the Data panel.

Generating 2-3 standard raster 2D display items seem to trigger the issue.

cmeyer:
Linked PR #1379

Ion-e:
This seems to be working finely. Thanks!

PR #1379, opened 2025-03-17
Fix issue where selected data items not updated for all selection mutations. Test.

cmeyer:
Linked Issue #1378.

This PR fixes discrepancies between what is drawn as selected in the data panel, the actual selection indexes, and the selected display items. Also adds a messy regression test to test this specific case.

Reviews are appreciated. The key code is documented in comments. Other code is related refactoring. I plan to merge this after getting an approval review or after two working days.

Regarding testing: this can be tested against the instructions in #1378. There may. be other lurking focus issues similar to this. But I did do a reasonably comprehensive search for similar cases and did not find any.

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -279,8 +279,8 @@ class DocumentController(Window.Window):
         self.__weak_periodic_listeners: typing.List[_PeriodicListenerWeakRef] = list()
         self.__weak_periodic_listeners_mutex = threading.RLock()
 
-        self.selection = Selection.IndexedSelection()
-        self.selection.expanded_changed_event = True
+        self.__selection = Selection.IndexedSelection()
+        self.__selection.expanded_changed_event = True
 
         # the user has two ways of filtering data items: first by selecting a data group (or none) in the data panel,
         # and next by applying a custom filter to the items from the items resulting in the first selection.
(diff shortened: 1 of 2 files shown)


lisham2000:
Code read. Comments useful for code digestion

nion/swift/DocumentController.py
```diff
@@ -343,7 +343,13 @@ def queue_call_soon() -> None:
             collection_info_controller_list_model.append_item(collection_info_controller)
 
         def document_model_item_inserted(key: str, value: typing.Any, before_index: int) -> None:
-            if key == "data_groups":
+            if key == "display_items":
+                # ensure selected data items are invariant when display items are added. do it in two steps (clear and
+                # set) to trigger notifications
+                selected_display_items = copy.copy(self.__selected_display_items)
+                self.select_display_items_in_data_panel(list())
+                self.select_display_items_in_data_panel(selected_display_items)
```
Tiomat85, line 351: Having to set it to a blank list and then to the new list feels a bit wrong? Could any additional logic applied by running through it with empty list could be applied as part of it?

Tiomat85 (commented):
Just one question.

Ion-e (approved):

nion/swift/DocumentController.py
cmeyer: This _is_ the additional logic.

Overall, it would be nice if the selection were handled in a more uniform way. But the current implementation separates the `Selection` object, which tracks integer 'indexes' from the selected display items, which tracks objects. Keeping those in sync is problematic. But this is a reasonable solution available without rewriting significant bits of code.

The logic needs to handle the case when the selection indexes suddenly refers to different objects. The selection itself would not change but the objects it refers to would change. What logic would be applied without having an entirely new code path? This is a very simple way to accomplish the same thing in a reliable way with the existing code paths.

Tiomat85: That is ok, it just felt a bit ugly as that requires knowing you need to do that for that specific block of logic as that is the kind of debt that bites as a weird bug later on.  Not going to make a fuss on it if it solves the problem in the short term, but it feels like a little bit of technical debt to tidy up later on if the extra performance hit/redraw causes issues.

PR #1393, opened 2025-03-24
Ensure workspace is only saved when changed.

cmeyer:


Ion-e:
Issue persists on this branch. 

The workspace layout persists after restarting the pc, however the data items are lost.

I suspect this is a bug @cmeyer

Issue #1143, opened 2024-08-20
Annular ring low/high pass incorrectly labelled


KRLango:
When you make an annular ring graphic and set it to "High Pass" mode, it will behave as a low pass filter, similarly if you set it to "Low Pass" mode, it will behave as a high pass filter.

These should be rectified so the selected action is what the user gets.

However, we do need to be careful not to break existing workflows that may already include these graphics when we make this change.

Ion-e:
Happy with it, thanks!

PR #1396, opened 2025-03-27
Ensure graphics display match masks created from annular graphic

### LINKED ISSUE #1140: Ring mask calculates filter incorrectly for low/high pass cases

KRLango:
This should resolve #1140 and #1143 

This replaces #1254 Since the code that is modified has moved making a merge rather difficult and a new branch much easier.

### DIFF
--- a/nion/swift/model/Graphics.py
+++ b/nion/swift/model/Graphics.py
@@ -2481,15 +2481,16 @@ class RingGraphic(Graphic):
         bottom_marker_inner = mapping.map_point_image_norm_to_widget(Geometry.FloatPoint(calibrated_origin.y, calibrated_origin.x + self.radius_2))
         image_norm_test_point = mapping.map_point_widget_to_image_norm(p)
         test_radius = abs(image_norm_test_point - calibrated_origin)
-        if test_point(top_marker_outer, p, ui_settings.cursor_tolerance):
-            return "radius_1", True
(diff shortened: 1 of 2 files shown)


Ion-e:
Draft looks good to me!

cmeyer:
Do any tests cover this code already? If not, we should add a test that (a) fails before the changes; (b) succeeds after the changes; (c) helps avoid a regression on this in the future.

KRLango:
@cmeyer I would guess not since I have changed the behaviour and it has not failed any tests. I will double check and add one.

Tiomat85 (approved):

lisham2000 (approved):
Changes and tests LGTM

cmeyer:
Merged 83f24ba50913d4ffb6969390ad7bb75f7cc8dbb0 and 44e0b11a91a66bf0e4a79ccc1a72138ba711aabd

Issue #1193, opened 2024-10-07
Export error File name is incorrect


Ion-e:
![Image](https://github.com/user-attachments/assets/207f23de-f7ed-4d18-b780-7d2b6693f17f)

Fail dialog shows the attempted filename instead of the failed data/display item name.

cmeyer:
With the current naming scheme in the application, it might make sense to give both the title and the attempted file name.

Ion-e:
Perhaps, at least the user *should know* which data item was not exported. For reference: https://github.com/nion-software/nionswift/issues/1193

cmeyer:
@Ion-e Can you revisit this? I think this is fixed.

jamesrussell216:
Need to be careful here as "Multiple Data Items" don't have a "Title" - not entirely sure what's going in the code with these in how they get named, so just a note for whoever picks this up they might have to take care with those!

Ion-e:
![Image](https://github.com/user-attachments/assets/9f736676-bb3c-4ae8-857b-2bf956304a98)
An additional user case where this may be confusing. If  a user unticks the name (and other options) they may end up with a situation where some of their exports have failed and they cannot locate which ones.

KRLango:
This is covered by #1398

Ion-e:
Looks good to me.

Ion-e:
Verified as implemented.

PR #1398, opened 2025-03-27
Update UI to show only failed exports and to always show the dialog

KRLango:
Addresses the points raised in https://github.com/nion-software/nion-internal/issues/136

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -223,22 +223,20 @@ class ExportDialog(Declarative.Handler):
                     if viewmodel.include_sequence.value:
                         components.append(str(index))
                     filepath = ExportDialog.build_filepath(components, writer.extensions[0], directory_path=directory_path)
-                    file_name = filepath.name
                     file_extension = filepath.suffix[1:].lower()
                     if writer.can_write_display_item(display_item, file_extension):
                         ImportExportManager.ImportExportManager().write_display_item_with_writer(writer, display_item, filepath)
-                        export_results.append(ExportResult(file_name))
+                        export_results.append(ExportResult(display_item.displayed_title))
                     else:
                         error_message = _("Cannot export this data to file format")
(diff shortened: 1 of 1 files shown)


cmeyer:
Quick look at code, no huge red flags. Did not test it. Are we eliminating successful results from the dialog? What if everything succeeds? (again: didn't try it, so these may be obvious)

KRLango:
@cmeyer When everything succeeds, the dialog appears with the message "7 of 7 succeeded" and the button to open the folder is there, but the box with the failed exports is not there.

KRLango:
I have just realised this also covers #1193

nion/swift/ExportDialog.py
```diff
@@ -681,18 +679,24 @@ def _build_ui(self, u: Declarative.DeclarativeUI) -> None:
...
         file_name_labels = list()
         status_labels = list()
 
         for export in self.exports:
-            status_text = _("Succeeded") if not export.error else f"\N{WARNING SIGN} {export.error}"
-            color = 'green' if not export.error else 'red'
-            file_name_labels.append(u.create_label(text=export.file, tool_tip=export.file, width=FILE_FIELD_WIDTH))
-            status_labels.append(u.create_label(text=status_text, tool_tip=export.error, color=color, width=STATUS_FIELD_WIDTH))
+            if export.error:
+                status_text = _("Succeeded") if not export.error else f"\N{WARNING SIGN} {export.error}"
```
Tiomat85, line 691: Is that second check redundant?  In fact is the whole 'Succeeded' chain here not able to be removed as this should only need the code for when export.error is populated.

Tiomat85 (changes requested):
Just one comment about some code I think is redundant.

Tiomat85 (approved):

cmeyer:
Merged b4482a980774ea6569d33ae8a47e764f159828de

Issue #1264, opened 2024-11-25
Separate data storage fully from data item to facilitate data compression


cmeyer:
A project contains various objects such as data items, display items, computations, etc. Most objects are "small" meaning that they can be stored in a small JSON-like dictionary in the main project database index file.

Data items, however, include both a small JSON-like dictionary and the data array. Since the data array is typically very large, data items are stored in an application-specific flat file format (either 'ndata' or 'h5').

Storing the data items in this way, though, limits development in a number of ways, including making it more difficult to migrate the entire project to a new project file format, since many data files much be touched during a migration; and limiting use of common file formats for the internal storage, since the format must include the ability to store and quickly update the object-part of the data item.

This issue proposes to move the object portion of the data item to the project file while leaving the data array portion in the flat file.

As part of this proposal, there should be a system whereby key information such as the data description, coordinate system descriptions, and metadata can be stored in duplicate, both in the flat file and in the project file, with rules about which is definitive depending on the specific flat file format, e.g. the flat file could be a 'bmp' which has no ability to store auxiliary information in which case the project file would be definitive; but could also be 'ndata' which fully stores the auxiliary information and could be definitive in older projects. Other information would only be stored in the project file going forward, including items such as 'category' or 'dynamic_title_enabled'. Some information may exist in both places, such as 'title'.

Also as part of this proposal, the current system of scanning the data folder for data items may need to be replaced with a system that tracks the file references within the project.

PR #1405, opened 2025-03-31
Improve file format handling when data item data size changes

cmeyer:
- **Improve error reporting when project can't be opened.**
- **Consolidate large format handling. Change file format to match changing data size.**

This changes behavior in the file storage system to automatically change the internal file format depending on the size of the data. A small data item stored as `.ndata` may be changed to `.h5` if the data size goes over 16MB.

Please review. The file storage system is not the cleanest code, but this change should be pretty clear that it tries to check the file type when the data changes. I'd like to merge this with or without reviews after two working days.

### DIFF
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -491,6 +491,8 @@ class Application(UIApplication.BaseApplication):
             try:
                 document_controller = self.open_project_window(project_reference, update_last_project_reference)
             except Exception:
+                import traceback
+                traceback.print_exc()
                 self.show_ok_dialog(_("Error Opening Project"), _("Unable to open default project."), completion_fn=self.show_choose_project_dialog)
                 return True
 
--- a/nion/swift/GeneratorDialog.py
+++ b/nion/swift/GeneratorDialog.py
@@ -193,11 +193,9 @@ class GenerateDataDialog(Declarative.WindowHandler):
             calibrations.append(Calibration.Calibration(units="nm"))
             calibrations.append(Calibration.Calibration(units="eV"))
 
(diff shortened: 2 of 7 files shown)


Tiomat85:
I can't see any particular problems with the code, but it felt a little clunky as you mentioned.  
What could be a nicer rework would be to have the StorageHandler class itself validate whether it can save a particular data item rather than above.  So the ndata handler would return that it cannot save an item above the threshold.  This would then allow for a bit more granular control, such as if certain formats can't handle certain data types etc.

cmeyer:
The problem with having the storage handlers themselves decide based on size is that it involves heuristics and I didn't want to distribute those heuristics into each storage handler. As an example: ndata can handle data sizes > 16MB, but we don't _want_ it to since there is better performance using h5. To that end, and agreeing with your comments, I think the storage handlers should have methods that allow the "system" to decide which one to use. However, I didn't want this PR to go too far beyond what is already in there, so maybe we leave that to a future PR? This code is going to be reworked as I implement #1264.

Tiomat85:
> The problem with having the storage handlers themselves decide based on size is that it involves heuristics and I didn't want to distribute those heuristics into each storage handler. As an example: ndata can handle data sizes > 16MB, but we don't _want_ it to since there is better performance using h5. To that end, and agreeing with your comments, I think the storage handlers should have methods that allow the "system" to decide which one to use. However, I didn't want this PR to go too far beyond what is already in there, so maybe we leave that to a future PR? This code is going to be reworked as I implement #1264.

That is fine with me.  I had taken it that ndata was unable to handle >16MB and not that it was an application level choice, so am happy with it as is.

Tiomat85 (approved):

PR #1415, opened 2025-04-03
Make line plot displays fill transparent when creating composite line…

lisham2000:
… plot displays

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -1257,6 +1257,10 @@ class DocumentController(Window.Window):
         if data_item and target_display_item:
             if target_display_item.display_data_channel:
                 new_display_item = target_display_item.snapshot()
+                display_layer = new_display_item.display_layers[0]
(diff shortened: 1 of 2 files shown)


Ion-e:
Stroke width is not set to 2 as default when creating composite line plots. This may be related to https://github.com/nion-software/nionswift/issues/1417

KRLango (commented):
Code looks good. But when I accidentally dragged a 1D display item onto itself, I got a version with the empty fill, but only the one graph. Are we OK with that behaviour? Or is this a bug?

cmeyer:
Sounds like a bug.

cmeyer:
I haven't reviewed this PR yet, btw. Just noting that dragging onto itself it probably an existing bug.

cmeyer:
Merged aeba6c0767306de246242f489c879bee56e447d5 and 41dd4b1ee7f252f8ce2a0cdd25e61130ca42fabf

Thanks @lisham2000!

Issue #960, opened 2023-06-20
Locked graphics should visually indicate that they are locked


cmeyer:
See also:
- #105

jamesrussell216:
@cmeyer how do you lock a graphic??

cmeyer:
In the inspector

<img width="317" alt="image" src="https://github.com/user-attachments/assets/9c309304-e000-466a-8536-5149d69b6077">

jamesrussell216:
How do we want to visually indicate this @cmeyer ? Change the colour of the "little green squares at the corners and one at the top for rotation that James can't think of the name of right now"? We'll need to be mindful that Luke's graphics PR is going to allow graphics to change colour, but that doesn't extend to those [data cursor things].

jamesrussell216:
Discussed this one today. Agreed on...

- The handles changing to crosses when shape is locked.
- The central "+" changing to a cross when position is locked.

Ion-e:
Duplicated closed issue: https://github.com/nion-software/nionswift/issues/1331

cmeyer:
After trying out the changes in PR #1424, I'm very negative on using 'x' for an endpoint. It looks non-standard and not intuitive. it conveys "danger" to me rather than something as benign as being locked. I think we should abandon that idea and explore alternatives, probably just a square with white interior, or a separate lock attached to the graphic.

cmeyer:
PR
- #1424

Ion-e:
This ticket (and https://github.com/nion-software/nionswift/pull/1424) seem very confusing. Could I get the final specs of what was introduced as bullet points so that I can create a set of tests for it?

jamesrussell216:
Review this one after:

https://github.com/nion-software/nionswift/issues/1535
https://github.com/nion-software/nionswift/issues/1532

PR #1424, opened 2025-04-07
Setting markers to X if locked shape, and central marker if locked position.

Tiomat85:
Addresses issue #960 .
Wrapped drawing of the markers into single functions to remove duplication of code.
Extended them to include whether the marker was interactable (squircle) or locked (cross).  Similarly for the centre marker, plus or cross.

This does not address rotation locking (though the same draw_marker function can be re-used), nor any inconsistent behaviour of any of the graphics with locking.

### DIFF
--- a/nion/swift/model/Graphics.py
+++ b/nion/swift/model/Graphics.py
@@ -437,23 +437,40 @@ def draw_arrow(ctx: DrawingContextLike, p1: Geometry.FloatPoint, p2: Geometry.Fl
     ctx.line_to(p2.x - arrow_size * math.cos(angle + math.pi / 6), p2.y - arrow_size * math.sin(angle + math.pi / 6))
 
 
-def draw_marker(ctx: DrawingContextLike, p: Geometry.FloatPoint, is_enabled: bool) -> None:
-    with ctx.saver():
-        marker_color = '#00FF00'
-        ctx.begin_path()
-        ctx.move_to(p.x - 3, p.y - 3)
-        ctx.line_to(p.x + 3, p.y - 3)
-        ctx.line_to(p.x + 3, p.y + 3)
-        ctx.line_to(p.x - 3, p.y + 3)
-        ctx.close_path()
-        if is_enabled:
-            ctx.fill_style = marker_color
-            ctx.fill()
-        else:
-            ctx.stroke_style = '#444'
-            ctx.stroke()
-            ctx.fill_style = '#ccc'
-            ctx.fill()
+def draw_marker(ctx: DrawingContextLike, p: Geometry.FloatPoint, is_enabled: bool, is_interactable: bool) -> None:
+    if is_interactable:
+        with ctx.saver():
+            marker_color = '#00FF00'
+            ctx.begin_path()
+            ctx.move_to(p.x - 3, p.y - 3)
+            ctx.line_to(p.x + 3, p.y - 3)
+            ctx.line_to(p.x + 3, p.y + 3)
+            ctx.line_to(p.x - 3, p.y + 3)
+            ctx.close_path()
+            if is_enabled:
+                ctx.fill_style = marker_color
+                ctx.fill()
+            else:
+                ctx.stroke_style = '#444'
+                ctx.stroke()
+                ctx.fill_style = '#ccc'
+                ctx.fill()
+    else:
+        with ctx.saver():
+            marker_color = '#00FF00'
+            ctx.line_width = 2
+            ctx.begin_path()
+            ctx.move_to(p.x - 3, p.y - 3)
+            ctx.line_to(p.x + 3, p.y + 3)
+            ctx.move_to(p.x + 3, p.y - 3)
+            ctx.line_to(p.x - 3, p.y + 3)
+            ctx.close_path()
+            if is_enabled:
+                ctx.stroke_style = marker_color
+                ctx.stroke()
+            else:
+                ctx.stroke_style = '#ccc'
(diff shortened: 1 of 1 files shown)


nion/swift/model/Graphics.py
```diff
@@ -465,20 +482,41 @@ def draw_circular_marker(ctx: DrawingContextLike, p: Geometry.FloatPoint, is_ena
         draw_ellipse(ctx, p, size, '#444', '#ccc')
 
 
-def draw_rect_marker(ctx: DrawingContextLike, r: Geometry.FloatRect, is_enabled: bool) -> None:
-    draw_marker(ctx, r.top_left, is_enabled)
-    draw_marker(ctx, r.top_right, is_enabled)
-    draw_marker(ctx, r.bottom_right, is_enabled)
-    draw_marker(ctx, r.bottom_left, is_enabled)
+def draw_centre_marker(ctx: DrawingContextLike, origin: Geometry.FloatPoint, graphic_size: Geometry.FloatSizeTuple, stroke_style: typing.Any, can_move_shape: bool, mark_size: int = 8) -> None:
```
KRLango, line 485: This function took me a while to parse. Would it be better to split it into a `can_move_shape` function and a `not can_move_shape` function to make it more obvious what is going on?

nion/swift/model/Graphics.py
```diff
@@ -1263,7 +1258,7 @@ def _draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, map
         bounds = Geometry.FloatRect.make(self.bounds)
         center = mapping.map_point_image_norm_to_widget(bounds.center)
         size = mapping.map_size_image_norm_to_widget(bounds.size)
-        draw_ellipse_graphic(ctx, center, size, rotation, is_selected, is_focused, stroke_style, fill_style)
+        draw_ellipse_graphic(ctx, center, size, rotation, is_selected, is_focused, not self.is_shape_locked, not self.is_position_locked, self.used_stroke_style, fill_style)
```
KRLango, line 1261: Passing these all in as `not self.is_..._locked` feels like code smell. Would we be better having the inputs being `is_locked`, rather than `can_move`?

KRLango (changes requested):
I haven't tested it, but wanted to raise a few points from looking at the code changes.

nion/swift/model/Graphics.py
```diff
@@ -465,20 +482,41 @@ def draw_circular_marker(ctx: DrawingContextLike, p: Geometry.FloatPoint, is_ena
...
+            ctx.move_to(mid_x, mid_y - 0.5 * mark_size)
+            ctx.line_to(mid_x, mid_y + 0.5 * mark_size)
+            ctx.stroke_style = stroke_style
+            ctx.stroke()
+
+def draw_rect_marker(ctx: DrawingContextLike, r: Geometry.FloatRect, is_enabled: bool, can_adjust_shape: bool) -> None:
+    draw_marker(ctx, r.top_left, is_enabled, can_adjust_shape)
+    draw_marker(ctx, r.top_right, is_enabled, can_adjust_shape)
+    draw_marker(ctx, r.bottom_right, is_enabled, can_adjust_shape)
+    draw_marker(ctx, r.bottom_left, is_enabled, can_adjust_shape)
```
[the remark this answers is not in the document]
Tiomat85: This was the name it had before, so I was avoiding trying to change the name unless needed.

nion/swift/model/Graphics.py
```diff
@@ -437,23 +437,40 @@ def draw_arrow(ctx: DrawingContextLike, p1: Geometry.FloatPoint, p2: Geometry.Fl
...
-        ctx.close_path()
-        if is_enabled:
-            ctx.fill_style = marker_color
-            ctx.fill()
-        else:
-            ctx.stroke_style = '#444'
-            ctx.stroke()
-            ctx.fill_style = '#ccc'
-            ctx.fill()
+def draw_marker(ctx: DrawingContextLike, p: Geometry.FloatPoint, is_enabled: bool, is_interactable: bool) -> None:
```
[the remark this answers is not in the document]
Tiomat85: I was contemplating that, splitting out the interactable/non-interactable options but it was borderline as it fit in a single screen fine. If it got much longer with other options then it might make sense to split out.

nion/swift/model/Graphics.py
Tiomat85: Similar to the draw_marker one, it was borderline on whether to have them as a single function split into two halves or individual functions.  I don't mind splitting them out if we agree, along with tweaking the naming to match but I was trying to avoid having really long function names.

nion/swift/model/Graphics.py
Tiomat85: Certainly could invert those flags.

cmeyer:
It looks like you're planning a few more commits - so I'm not looking at the code.

Subjective opinion: the 'x' endpoint looks non-standard and not intuitive. it conveys "danger" to me rather than something as benign as being locked. I think we should explore alternatives.

But some detail comments.
- the drawing is inconsistent. here is a locked endpoint on a line, lopsided and too thick:
image

- the locked-position rectangle center cross is too large. the _length_ of the lines should be the same between locked and unlocked. are you drawing the 'x' to be the corners of the box? this makes each line too large by sqrt(2). drawing may need a line width adjustment too.
image 
image

- the point graphic does not change after being position locked

KRLango:
LGTM

KRLango (approved):

Issue #1443, opened 2025-04-11
Line plot drawing incorrect on last pixel


cmeyer:
<img width="1107" alt="Image" src="https://github.com/user-attachments/assets/e1a31515-a10a-4cb4-affd-0d21d1ba02ac" />

cmeyer:
Linked PR:
- https://github.com/nion-software/nionswift/pull/1452

cmeyer:
New drawing style: strokes do not draw along sides.

<img width="537" alt="Image" src="https://github.com/user-attachments/assets/3ce3fcbe-2818-4206-97b6-d97ab161e287" />

Old drawing style (note: left-most stroke is missing):

<img width="541" alt="Image" src="https://github.com/user-attachments/assets/1b58d57c-781d-4aaa-9e47-205a22d7505b" />

Issue #1418, opened 2025-04-03
Additional line profiles not immediately visible when added into a composite line profile


lisham2000:
Steps to recreate:

1. Have 3 line profiles on a (simulated) image:
2. Create a composite line profile using 2 of these line profiles by dragging one onto another ( works as intended)
3. Drag the third line profile onto this composite display and it will appear in the legend but not in the composite line profile

If you adjust any of the line profiles it then appears

Highlighted by #1415

cmeyer:
Linked PR:
- https://github.com/nion-software/nionswift/pull/1452

cmeyer:
Fixed 91f079e93a6ac50cd6fd34eaa2a59b3d46eb764d.

@lisham2000 Close?

PR #1452, opened 2025-04-15
line plot tweaks

cmeyer:
- **Change line plot to not draw returns to baseline on strokes.**
- **Ensure display data channel display values initially get calculated.**

Please review these two commits. The first (don't draw returns to baseline) change how drawing looks on the ends of segments #1443. The second fixes an issue where sometimes line plots would not draw initially #1418. I would like to merge them after a day. See #1443 for drawing notes.


jamesrussell216:
Just from me - this looks like a good change. I've not tested it or anything (having install issues at the moment) but from what Chris has shown it looks good to me.

Issue #1463, opened 2025-04-30
Restore ability to rename data group


jamesrussell216:
Currently you can add new groups, but there's no menu option to rename them, so you just end up with a bunch of "Untitled Group" groups. The code seems to be there under the hood to do, it's just not accessible from the UI. I suspect that the functionality to delete a group may fall into the same catagory!

jamesrussell216:
Just realised @cmeyer I never sent this one your way when we were on the train!

jamesrussell216:
Probably needs a delete function as well.

cmeyer:
This should be available by
- click-hovering the data group, like Explorer rename
- menu item when data group is selected
- right clicking data group
- pressing return when data group is focused

Any of the actions above should pose a small pop-up dialog to rename, similar to title edit. Pop-up dialog should respond to escape and return keys to cancel and accept.

cmeyer:
Can this be closed in favor of the more specific
- #1486

PR #1481, opened 2025-05-16
Add ability to rename data group by hitting return or double clicking on list item.

cmeyer:
This adds the missing ability to rename a data group. It requires the recent changes in `nionui`. The basic functionality is to click on a data group in the Collections panel, and either hit return or double click the item to change the name. The resulting pop-up dialog should be able to copy/paste/cut/select-all. The pop-up dialog should immediately respond to typing, i.e. not require any additional focus clicks; and should also respond to return to accept and escape to cancel.

Please review with the consideration that this is an acceptable but probably incomplete solution to #1463. I would suggest reviewing to ensure it is working as expected. I would also suggest testing this pop-up in varying situations including using return/escape, switching out of the application, exiting the application, having the pop-up appear on different sides of the screen, in various stages of scroll bar, etc. I tested as thoroughly as I could on macOS only.

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -125,9 +125,19 @@ class CollectionInfoController(Observable.Observable):
         oo.source(container).sequence_from_array("display_items", predicate=self.__filter_predicate.matches).len().action_fn(count_changed)
         self.__count_observer = typing.cast(Observer.AbstractItemSource, oo.make_observable())
 
+        def title_changed(key: str) -> None:
+            if key == "title":
+                if self.__data_group:
+                    self.__base_title = self.__data_group.title
+                    self.collection_info_stream.value = self.collection_info
+                    self.notify_property_changed("collection_info")
+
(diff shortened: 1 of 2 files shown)


KRLango:
The popup is working as described, but on windows it seems to be the wrong type of popup as it has the menu bar along the top:

Screenshot 2025-05-16 154557

From here you can also do things such as "Exit" which closes Swift:

Screenshot 2025-05-16 154612

This is undesirable.

Probably not relevant if we decide to change the popup type to remove the menu bar, but the cut/copy/paste actions are all in the edit menu, but greyed out. However the keyboard shortcuts for them worked fine.

Also, the popup remains open when you change focus. I think it would be better if it closed once you change focus to something else in Swift.

cmeyer:
The menu problem is fixed in nion-software/nionui@c85c9ddc91c64ec00de09d4bb173edad162f6afc

Ion-e:
New ticket is made for the existing issue: https://github.com/nion-software/nionswift/issues/1486

Otherwise it looks good to me.

PR #1545, opened 2025-06-06
Fix #1508. Standardize spot filter, angular filter, band-pass filter, lattice-filter names.

### LINKED ISSUE #1508: Graphics: Band pass using 2 different nomenclatures
In some places in the software "Band pass" also appears as "ring tool".

cmeyer:


jamesrussell216:
Agreed, but are they filters? They can be used with filtering (as many graphics can be), but if you put a lattice graphic down you still have to set up a filter computation - the graphic in and of itself isn't a filter. Happy to be convinced either way!

Brow71189:
I agree with James that these graphics are not really filters until you apply them, but I also cannot think of a better naming scheme. Generally I like the idea of distinguishing "Fourier filters" and "Masks", because they behave differently.
So overall I'm fine with these changes.

jamesrussell216:
Gah OK... I appreciate the difficulties here. I could be convinced about calling these "filters". However this raises a new problem for me that I think is as bad as the original inconsistency - we already have other things, different things, that are called filters!...

image

I'm fine on the "names", just not sold on the use of "filter". I still think "graphic" would be better.

cmeyer:
Yep the dual use of filters is a problem. Hmm.

Brainstorming:
- "filter graphics"
- "filter region"
- "filter mask"
- "Fourier mask"
- "Fourier filter"
- "Fourier filter graphic"
- "frequency mask"
- "frequency filter"
- "frequency filter graphic" (too wordy)
- "Fourier region"
- "Fourier filter region"

cmeyer:
We could just call them:
- "Spot Graphic"
- "Band-Pass Graphic"
- "Angular Graphic"
- "Lattice Graphic"

Ion-e:
Issue verified as fixed on branch.

jamesrussell216:
@cmeyer apparently I thumbed up your name suggestion from Jun 26th, they look sensible so I'd go with those :-)

Ion-e:
This looks good to me, I had a merge conflict on testing but it's all good now.

Thank you everyone.

cmeyer:
Squashed and merged 6fbbaf751c3eb7afc64bb3b21eb04e6d14ff4de0

Issue #1532, opened 2025-06-05
Graphics Tools show and allow changing of states that don't make sense for the type of Graphic


Tiomat85:
Some Graphic types have no concept of position or shape, and yet the checkboxes for locking those are still there and enabled. Propose that they should instead be disabled where they do not make sense.  So where the options do not make sense for a graphic type, the corresponding checkbox in the Inspector is disabled (greyed out).

| Type                    | Position | Shape |
|-------------------------|----------|-------|
| Ellipse                 | Yes      | Yes   |
| Point                   | Yes      | No    |
| Line (and line profile) | Yes      | Yes   |
| Rectangle               | Yes      | Yes   |
| Wedge                   | No       | Yes   |
| Lattice                 | Yes      | Yes   |
| Spot                    | Yes      | Yes   |
| Ring                    | No       | Yes   |
| Interval                | Yes      | Yes   |
| Channel                 | Yes      | No    |

Tiomat85:
Closing as Merged.

PR #1554, opened 2025-06-13
Add bool properties of whether a graphic is lockable. Disable checkboxes appropriately.

Tiomat85:
Addresses #1532.

Adding property of whether a graphic type is lockable for states.
Disabling checkboxes in the inspector appropriately.

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -2893,9 +2893,9 @@ class GraphicsInspectorHandler(Declarative.Handler):
         lock_row = u.create_row(
             u.create_spacing(3),
             u.create_label(text=_("Lock"), width=60, text_alignment_vertical="center"),
-            u.create_check_box(text=_("Position"), checked="@binding(_lock_position_model.value)", text_alignment_vertical="center"),
+            u.create_check_box(text=_("Position"), checked="@binding(_lock_position_model.value)", enabled="@binding(_graphic.can_lock_position)", text_alignment_vertical="center"),
             u.create_spacing(12),
-            u.create_check_box(text=_("Shape"), checked="@binding(_lock_shape_model.value)", text_alignment_vertical="center"),
+            u.create_check_box(text=_("Shape"), checked="@binding(_lock_shape_model.value)", enabled="@binding(_graphic.can_lock_shape)", text_alignment_vertical="center"),
             u.create_stretch(),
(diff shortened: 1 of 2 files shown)


nion/swift/Inspector.py
```diff
@@ -2893,9 +2893,9 @@ def __init__(self, document_controller: DocumentController.DocumentController,
         lock_row = u.create_row(
             u.create_spacing(3),
             u.create_label(text=_("Lock"), width=60, text_alignment_vertical="center"),
-            u.create_check_box(text=_("Position"), checked="@binding(_lock_position_model.value)", text_alignment_vertical="center"),
+            u.create_check_box(text=_("Position"), checked="@binding(_lock_position_model.value)", enabled="@binding(_graphic.can_lock_position)", text_alignment_vertical="center"),
             u.create_spacing(12),
-            u.create_check_box(text=_("Shape"), checked="@binding(_lock_shape_model.value)", text_alignment_vertical="center"),
+            u.create_check_box(text=_("Shape"), checked="@binding(_lock_shape_model.value)", enabled="@binding(_graphic.can_lock_shape)", text_alignment_vertical="center"),
```
KRLango, line 2898: Is there any situation for a given graphic where can_lock_position and can_lock_shape can change? If not wouldn't we be better just setting the enabled value directly?

cmeyer (commented):
A method here doesn't seem right since there is never a need to return anything but a constant. What about a capitalized property defined on the class itself?

```
class Graphic:
    @property
    @abstractmethod
    def CAN_LOCK_POSITION(self) -> bool:
        # must be implemented by subclass, no default
        pass

class RectangleTypeGraphic(Graphic):
    CAN_LOCK_POSITION = True

    def __init__(...)...
```

Tiomat85:
> My comments weren't entered as "request changes" so I'm putting them here again.
> 
> * `can_lock_position` doesn't seem like the right way to do this. What about using a constant defined on the class?
> * I think disabled UI elements should have a tool tip to explain why disabled.

Tooltip I agree with, so will look at adding something when I can figure out some nice wording for it.

The property vs function, I have no real preference either way so can change it to a capitalized property.  I still think having a defined default of True is better though rather than requiring it specifying in each one, simply for less overhead required. 

---

On the query about inspecting multiple items, if we can define the logic of what we would (Any lockable so it is enabled?) we could roll a new property that handles that. I don't see it causing big problems.

PR #1567, opened 2025-06-18
Add metadata paths for scan and camera orientation to metadata.py

KRLango:
This is part of https://github.com/nion-software/nion-instrumentation/issues/390

It is to add the additional metadata paths to metadata.py for ease of access. However it may change to accommodate changes in https://github.com/nion-software/nion-instrumentation/pull/482.

### DIFF
--- a/nion/swift/model/Metadata.py
+++ b/nion/swift/model/Metadata.py
@@ -50,7 +50,7 @@ key_map = {
     'stem.beam_current_a': {'paths': ['instrument.beam_current', 'hardware_source.beam_current_a'], 'type': 'real'},
     'stem.defocus': {'paths': ['instrument.defocus', 'hardware_source.defocus_m'], 'type': 'real'},
     'stem.defocus_m': {'paths': ['instrument.defocus', 'hardware_source.defocus_m'], 'type': 'real'},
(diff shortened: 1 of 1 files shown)


Tiomat85 (approved):

KRLango:
@cmeyer I don't think so. https://github.com/nion-software/nion-instrumentation/pull/482 has changed and will require changes here as well.

cmeyer:
I'm more convinced now that we should just close this PR and not implement it. The problem is that the simplistic metadata approach in `Metadata.py` fails to cover the complicated semantics of axes. The axes are fundamentally different from the other scalar tags. If we have specific angles that we want to provide as standard metadata, and it is possible to do so, we can put those in as scalar-valued metadata and add tags for them in `Metadata.py`.

In the larger discussion, I think we should consider whether it would be good to create a metadata access library that provides programmatic access, perhaps in `nionswift-instrumentation-kit` or perhaps in a new library.

KRLango:
I am inclined to agree. We have all the necessary information in the metadata now, the only access is via the metadata dictionary. I think a class that can read this information and provide the angles etc in a clean form would be useful, but that is different to what this PR provides.

Issue #1312, opened 2025-01-17
Doing "Alt + click and drag from line graphic" creates an odd behaviour


Ion-e:
As title, when doing "Alt + click and drag from line graphic" (However, not from one of the extremes of the line graphic) the software behaves oddly.

Couldn't take a screenshot however it's easy to reproduce. 

Actual: It seems a thumbnail of the current data item is created and dragged along, however it disappears when left click is released and instead of dropping the "data item" in a different location (say in  a different display panel) it drags the line graphic to that last position.

Ion-e:
Thanks for the explanation!

jamesrussell216:
I think this (or another issue) needs reopening here; there are weird behaviours that go on with this functionality. For instance if you...
- Press Alt
- Left-click on the middle line
- Drag; this brings up the export interface.
- However if you release the left-click whilst still over the display item, _then move the mouse_, the line profile moves, which I'm fairly sure isn't intended?

PR #1569, opened 2025-06-19
Fix #1482. Do not move graphic after option-dragging to produce external drag. Note: no test.

### LINKED ISSUE #1482: Option drag line and then cancel (by dropping in the same panel) moves the line when it shouldn't
Draw a line graphic (note: not a line profile) on an display item...

Image

Press the ALT key, then left click and drag somewhere in the middle of the line graphic (not the ends). This brings up the export interface. However if you release the left click somewhere in the display item, the export cancels as it should be the line graphic moves to the release location which presumably is undesired...

Image

cmeyer:


Ion-e:
Issue verified as fixed. These changes should cover: https://github.com/nion-software/nionswift/issues/1312

However the fact that the thumbnail still is created persist. Do we want a new ticket for that? (Please note that the thumbnail only is created by doing this from a graphic, and not when selecting the display item directly)

KRLango (commented):
Code looks good. When testing it I found it sometimes moved the graphic, sometimes did the SVG copy/paste drag. It was always clear what it was doing, so this may be better as a separate issue.

Ion-e:
@cmeyer My question hasn't been addressed, should I make a new ticket for it if it's an actual issue?

Issue #1295, opened 2024-12-31
Add ability to switch collection directly from the data panel


cmeyer:
The Collections panel allows the user to change collections; and the Data Panel shows the currently selected collection. Sometimes the user will not have the Collections panel open and the extra line in the Data Panel provides a good place to put a UI element to be able to switch the collection without the Collections panel being open - either a down arrow triggering a pop-up menu on the right; switching collections when the user clicks on the collection name, or a pop-up menu when the user clicks anywhere in the collection name section.

<img width="363" alt="Image" src="https://github.com/user-attachments/assets/4b0f11d2-d047-4d98-856c-1ae2e7d0fc5a" />

In the image above, the collection area is the area with the text "Persistent (0)".

jamesrussell216:
@cmeyer seems like a potential easy win to just change that to a dropdown? I think that would be a more intuitive UI interaction that a swapping mechanism.

Ion-e:
-- Test plan:
- Open the collections and the data panel.
- Create several groups with different names in “File > Add group”.
- Create a few data items, live items and distribute them amongst those groups.
- Double click on the bar under the “Data panel”.
- A combobox menu should appear.
- The list of all the groups should appear.
- Click on Live.
	-The data panel should change to only live data items.
- Iterate this process throughout the different groups.
	- The data panel should update to reflect the chosen group.
- Click on the “Live” group in the collections panel.
	- The data panel should update with the “Live” group.
	- The combobox should update to show “Live”.
- Click the combobox.
- Hover the mouse over a different group as the chosen one.
- Press scape.
	- The combobox should close, and remain on the “Live” group.
- Create another group “New group”.
- The group should now appear in the combobox.

Ion-e:
Verified as implemented. Thanks!

PR #1571, opened 2025-06-19
Add combo box in place of the collection title for easier collection swapping

KRLango:
This is for #1295 

Add a combo box rather than the existing collection title to allow users to switch the current collection without having to swap panels.

For reference, this is what it looks like on windows:
image

### DIFF
--- a/nion/swift/DataPanel.py
+++ b/nion/swift/DataPanel.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import asyncio
 import copy
 import gettext
+import operator
 import pkgutil
 import threading
 import typing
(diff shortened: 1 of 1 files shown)


KRLango:
@jamesrussell216 the latest commit addresses most of your points. The only exception is:

> Also when you select the long group name it cuts it differently to when the dropdown is expanded...

I haven't touched that as it is currently using the default behaviour for the dropdown and I'm not sure if we should be changing that or not.

Ion-e (commented):
LGTM

Tiomat85 (approved):

Tiomat85:
I would lean towards option #3, but I don't have a huge preference either way.

jamesrussell216:
> I would lean towards option #3, but I don't have a huge preference either way.

Second this.

Regarding @Ion-e comment on the real estate, I think if you tried to put anything else in there it would look too cluttered. Weirdly I think having it full length make it look less cluttered. I think it's a good improvement and will make swapping collections easier.

jamesrussell216 (approved):
LGTM from user POV, obviously not commenting on the code.

Tiomat85 (approved):
Looks fine and more balanced.

PR #1576, opened 2025-06-20
Ensure that error is raised when data item is deleted

KRLango:
This is for https://github.com/nion-software/nionswift-instrumentation-kit/issues/249

The acquire function already has a cancel on error, but nothing was throwing an error to let it know that the dataitem no longer existed. 

This was the closest place I could find that knows about the item having being closed/deleted.

I couldn't think of a reason we would want to continue updating a dataitem that has been closed. When we raise this error, the error handling in the acquire function kicks in and cancels the acquisition.

### DIFF
--- a/nion/swift/model/DataItem.py
+++ b/nion/swift/model/DataItem.py
@@ -921,6 +921,7 @@ class DataItem(Persistence.PersistentObject):
 
     def queue_partial_update(self, partial_xdata: DataAndMetadata.DataAndMetadata, *, src_slice: typing.Sequence[slice],
                              dst_slice: typing.Sequence[slice], metadata: DataAndMetadata.DataMetadata) -> None:
+        assert not self._closed, _("Cannot update deleted data item")
         with self.__pending_xdata_lock:
             self.__pending_queue.append((partial_xdata, src_slice, dst_slice, metadata))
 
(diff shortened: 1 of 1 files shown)


cmeyer (commented):
I'll look at this in more detail, but I'm hesitant to add more usage of `_closed` which is a hack (by me), at best. Even the existence of `_closed` (again: my doing) is a strong code smell. This PR, though, may end up to be the "best" solution for now.

Tiomat85 (approved):
Ignoring the _closed discussion, the usage of it here makes sense and is good defensive style.

Tiomat85:
> I'll look at this in more detail, but I'm hesitant to add more usage of `_closed` which is a hack (by me), at best. Even the existence of `_closed` (again: my doing) is a strong code smell. This PR, though, may end up to be the "best" solution for now.

I would say having an internal field whether an object is 'valid' or not is fine (assuming it is properly used).  Whether it needs refactoring/protected a bit more I think would be fine as an improvement.

KRLango:
What do you mean by a "fast acquisition"? I have tried several from ~0.05 seconds up to 1 second, and I either see the acquisition end before I manage to delete the data item, or I get the error I added.

cmeyer:
Reluctantly merged this. See comments above. 92ca8aa79eb457a8963c8332e7e1b4e356f326a8

Issue #1469, opened 2025-05-08
Unfocused text in line plot intervals should not be dimmed


cmeyer:
There should be some indication that it is not focused; but the text should be visible as these are used for information display and dimming them makes them difficult to read.

cmeyer:
I don't think we should change the behavior when the interval is not selected. It is way too noisy if text is drawn for all intervals; and the user has an easy way currently to decide if the text is drawn or not: select the interval or don't.

As with the original issue: when the interval is selected but the focus is not on the display panel, we should simply not dim the text.

jamesrussell216:
> I don't think we should change the behavior when the interval is not selected. It is way too noisy if text is drawn for all intervals; and the user has an easy way currently to decide if the text is drawn or not: select the interval or don't.
> 
> As with the original issue: when the interval is selected but the focus is not on the display panel, we should simply not dim the text.

As I understood it @cmeyer that behaviour was specifically requested by the customer though? If we're going to change an agreed spec, and the works already been done, then I would at least like to go back to the customer to clarify.

KRLango:
@jamesrussell216 I think that comes from the concern I raised about how it looked in the PR: https://github.com/nion-software/nionswift/pull/1579

jamesrussell216:
Ahh OK that makes a bit more sense. Yes I see the problem - but is that a realistic use case though? I mean, I can shove a whole bunch of graphics done on a display item and make it a right mess too, but if it's not how people actually use the software in practice we maybe solving a problem that doesn't exist.... so I think I'll drop the customer an email anyways and just get their two pence on it and feed that back, thanks for raising both.

PR #1579, opened 2025-06-23
Change the conditions to apply new formatting criteria on interval graphics

KRLango:
Fix for https://github.com/nion-software/nionswift/issues/1469

This updates the criteria for the display of interval graphics on 1D plots.

It looks congested when you have several intervals:
Generated Data

But it does keep the numbers visible.

### DIFF
--- a/nion/swift/LineGraphCanvasItem.py
+++ b/nion/swift/LineGraphCanvasItem.py
@@ -912,14 +912,10 @@ class LineGraphRegionsCanvasItemComposer(CanvasItem.BaseComposer):
 
                     region_color = region.color
                     selection_color = region_color
-                    if not self.__is_focused:
-                        r, g, b = Color.Color(region_color).to_rgb_255()
-                        p = 0.25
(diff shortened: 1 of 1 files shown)


cmeyer (commented):
I don't think we want to make the change proposed in this PR, but we can discuss offline.

Also, the original issue is simply that the text gets dimmed when it doesn't have keyboard focus, not that it doesn't have the text. Again, maybe we can discuss offline,

lisham2000 (approved):
Code wise all good from me

Tiomat85 (approved):
LGTM

cmeyer:
I'm going to reject this PR since it changes fundamental behavior of what selected intervals vs. non-selected intervals display. The original issue #1469 is addressed with this one line change 2b21fc7f3e03d655c2efb0d27d25499b21e92175, which fixes the regression introduced in c74a4cb91c590e41f3e481d30dbd6db684cef832.

KRLango:
Given the level of changes here, I think we would be better just closing this PR for now and if we decide we do want different behaviour to what we have now, it is also likely to be different to what this PR has and would be easier to simply make a new PR with the new behaviour rather than try and edit this one.

PR #1581, opened 2025-06-24
Change lock functionality for Lattice, LineType, Ring, Wedge and Interval

lisham2000:
Solve graphic locking issues:

1. The line graphic doesn't respect position lock
2. Ring tool shape lock doesn't lock the annular radii
3. Interval tool, position lock should lock the centroid
4. Wedge tool shape lock should lock the angles
5. Lattice tool, shape locks both position and shape at the moment. Position lock should lock the lattice periodicity.

### DIFF
--- a/nion/swift/test/Graphics_test.py
+++ b/nion/swift/test/Graphics_test.py
@@ -225,7 +225,6 @@ class TestGraphicsClass(unittest.TestCase):
         self.assertAlmostEqual(end0 - start0,
                                wedge.end_angle - wedge.start_angle,
                                delta=0.001)
-        print(end0 - start0, wedge.end_angle - wedge.start_angle)
         wedge.close()
 
     def test_lattice_graphic_position_lock(self):


nion/swift/model/Graphics.py
```diff
@@ -2317,6 +2320,14 @@ def adjust_part(self, mapping: CoordinateMappingLike, original: Geometry.FloatPo
...
         if d_angle > math.pi:
             d_angle = -(math.pi * 2 - d_angle)
+
+        if self.is_shape_locked:
+            if part[0] == "all" or part[0] == "inverted-all":
+                dtheta = o_angle - c_angle
+                self.__start_angle_internal = start_angle_original - dtheta
+                self.__end_angle_internal = end_angle_original - dtheta
+            return
+
```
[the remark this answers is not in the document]
lisham2000: This is to differentiate between the rotation lock that will come into affect within PR 1573 which will lock the start angle so both dimensions are lockable independently

Ion-e:
Grabbing the interval graphic by the centroid allows the user to change the width when position is locked.

This is an inconsistent behaviour with the line graphics, where doing the same don't allow for the graphic to be changed.

cmeyer:
@lisham2000 Can you split the "The line graphic doesn't respect position lock" functionality out into a separate PR? Also, it will need to include integration tests.

Ion-e:
For testing: the angle graphic will need to respect position too.

Ion-e:
I think I will make an exemption on this one and wait for the testing until all 3 graphics issue have been merged, once that happens I will be using this PR as a testing point for a more comprehensive testing.

cmeyer:
I changed this to 'draft' pending the changes to tests suggested by Kevin - the tests should be structured so they create the graphic, lock it, try to use the mouse to change it, confirm it is unchanged, then unlock it, then use the mouse to change it, and confirm it changes. The two 'mouse move' lines should be identical and add comments to explain how why they should be identical.

nion/swift/test/Graphics_test.py
```diff
@@ -192,16 +192,227 @@ def test_line_graphic_position_locked(self):
...
+        current = polar_deg(55)
+        part = ("all",) + wedge.begin_drag()
+        modifiers = CanvasItem.KeyboardModifiers()
+
+        # Replicate WedgeGraphic angle math for expected values.
+        def compute_angle(p: Geometry.FloatPoint) -> float:
+            return math.pi - math.atan2(center[0] - p[0], center[1] - p[1])
+
+        o_angle = compute_angle(original)
+        c_angle = compute_angle(current)
```
KRLango: To calculate these, you currently take the angle in degrees, pass it into `polar_deg` that converts it to radians, then to cartesian coordinates, then those go into 'compute_angle' to get the angle in radians.

A better design would be to have the angle sin radians from the start then we would already have the angles we need.

nion/swift/test/Graphics_test.py
```diff
@@ -192,16 +192,227 @@ def test_line_graphic_position_locked(self):
...
-        point_graphic.label = "Test"
-        self.assertEqual(point_graphic.test(mapping, ui_settings, Geometry.FloatPoint.make((250 - 18 - 6, 250)), move_only=True)[0], "all")
-        point_graphic.close()
+        center = Geometry.FloatPoint.make(mapping.calibrated_origin_widget)
+
+        wedge = Graphics.WedgeGraphic()
+        wedge.start_angle = math.radians(30)
+        wedge.end_angle = math.radians(60)
+
+        def polar_deg(angle_deg: float, radius: float = 200.0) -> Geometry.FloatPoint:
```
KRLango: `polar_to_cartesian` would be a more descriptive name. I would also take the angle in radians to match the other comment.

cmeyer:
Looks great! Thanks. Squashed and merged 960e9576351296077d547cf44935cfcca17b75fc

PR #1582, opened 2025-06-24
Add Range Slider to Slice Section

Tiomat85:
Removed the previous Slice and Width individual sliders with a single combined Range slider.

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -2363,31 +2363,42 @@ class SliceSectionHandler(Declarative.Handler):
         self.slice_center_model = DisplayDataChannelPropertyCommandModel(document_controller, display_data_channel, "slice_center", title=_("Change Slice"), command_id="change_slice_center")
         self.slice_width_model = DisplayDataChannelPropertyCommandModel(document_controller, display_data_channel, "slice_width", title=_("Change Slice"), command_id="change_slice_width")
 
-        self.slice_center_maximum = display_data_channel.data_item.dimensional_shape[-1] - 1 if display_data_channel.data_item else 0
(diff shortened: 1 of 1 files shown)


KRLango (changes requested):
I also noticed a couple of other points:

- If I have the slider set min-max, it shows the same range as if I set either minimum or maximum 1 in from the edge. (probably a symptom of the off-by-one error, but I wanted to mention it specifically in case it is not)
- Changing the min/max in the edit boxes does not update the range/centre edit boxes
- With 50 slices in the data, if I have the min and max bars at the limits then enter 10 in the range box, it gives me a range of 9 (and the center changed from 25 to 24)

Ion-e:
There is currently a bug, when swapping from this data item with a slider: 
image

To this one:
image

iteratively, then the range slider of the second data item starts to automatically update and change (The image changes too).

(When reproducing, start from the slider with top left to 0 and the top right part of the slider to somewhere in the centre of the section.

Ion-e:
Bugs: 

1. Use the interval graphic to swap the left and right axis, range doesn't update.
2. Drag only one of limits of the interval and see it hopping.
3. Drag the range slider, observe the interval hopping.
4. The range slider changes it's values when clicking between the display panel where the 1d plot is and the display panel where the range slider is in.
5. See screenshot 1.
6. When I introduce a specific range value in the edit box, and then I introduce a value in "centre" the first one gets overwritten by something else.
7. Similar happens as above when introducing a value in "max".
8. You can introduce floats in the edit boxes, but they don't seem to change anything.
9. It's possible to introuduce min/max values out of range, however the interval won't show it.

Screenshot 1:
image

PR #1584, opened 2025-07-01
Add Prefix character validation

### LINKED ISSUE #1490: Export multiple: it's possible to introduce invalid symbols in the prefix edit box
Open Export multiple window
Introuduce: \//\//\\\//\\////
Press export: Observe the software will crash.

Expected from RL: 
The "Export" button should become disable.
A red warning should appear to indicate that those are not valid characters for export.

*Please note that if an export of a data item that has any of these invalid characters and with the "include title" box ticked, it should also show the red warning (or at least fail with an error appropiately addressing the cause of the fail). Thanks!

Tiomat85:
Fixes #1490 

When editing the prefix, add validation and if invalid disable the ability to start an export with a reason given as a tooltip.  Further development could include visual identification of the problematic section.

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -87,7 +87,7 @@ class ExportDialog(Declarative.Handler):
         title_text = f"{export_text} ({len(display_items)} {items_text})"
         dialog = typing.cast(Dialog.ActionDialog, Declarative.construct(document_controller.ui, document_controller, u.create_modeless_dialog(self.ui_view, title=title_text), self))
         dialog.add_button(_("Cancel"), self.cancel)
-        dialog.add_button(_("Export"), handle_export_clicked)
+        self.__export_button = dialog.add_button(_("Export"), handle_export_clicked)
         dialog.show()
 
     def choose_directory(self, widget: Declarative.UIWidget) -> None:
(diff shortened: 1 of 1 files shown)


nion/swift/ExportDialog.py
```diff
@@ -101,6 +101,28 @@ def on_writer_changed(self, widget: Declarative.UIWidget, current_index: int) ->
...
         self.viewmodel.writer.value = writer
 
+    def on_prefix_changed(self, widget: Declarative.UIWidget, text: str) -> None:
+        self.validate_options()
+
+    def validate_options(self) -> bool:
+        invalid_reason: str | None = None
+
+        prefix_text = self.viewmodel.prefix.value or ""
+        if prefix_text != Utility.simplify_filename(prefix_text) or "\\" in prefix_text or "/" in prefix_text:
```
cmeyer, line 111: Doesn't `simplify_filename` already catch forward and backslashes? If not, why doesn't it?

cmeyer: Ping

cmeyer:
In addition to some stylistic items, I'm delaying this until the threaded export dialog is merged, which is a larger change and conflicts with this one.

nion/swift/ExportDialog.py
```diff
@@ -101,6 +101,26 @@ def on_writer_changed(self, widget: Declarative.UIWidget, current_index: int) ->
         writer = self.__writers[current_index]
         self.viewmodel.writer.value = writer
 
+    def on_prefix_changed(self, widget: Declarative.UIWidget, text: str) -> None:
+        self.validate_options(text)
+
+    def validate_options(self, prefix_text: str) -> bool:
```
cmeyer, line 107: Can this just be merged into `_handle_prefix_changed`?

nion/swift/ExportDialog.py
```diff
@@ -101,6 +101,26 @@ def on_writer_changed(self, widget: Declarative.UIWidget, current_index: int) ->
         writer = self.__writers[current_index]
         self.viewmodel.writer.value = writer
 
+    def on_prefix_changed(self, widget: Declarative.UIWidget, text: str) -> None:
```
cmeyer, line 104: Can this be named `_handle_prefix_changed`. The variables named `on_xyz` are usually function pointers, not handlers. And the `_` prefix will denote it as "as private as can be when passing to Declarative UI". Thanks.

cmeyer (changes requested):
Minor change requests.

cmeyer:
The `_handle_prefix_changed` should not have a `text` parameter:

```
TypeError: ExportDialog._handle_prefix_changed() got an unexpected keyword argument 'text'
```

cmeyer:
Merged b7a9556a2b8608f1442179b28a241dad4f409349

PR #1598, opened 2025-07-16
Fix calibration issues in display adjustment processors. Add tests. Fixes #1590.

### LINKED ISSUE #1590: RGB output loses calibration
Steps to reproduce:

Start with two calibrated images (in nm or rad, for instance). The images should have the same dimensions and calibration. Use processing-RGB-make RGB to make a combined RGB data item. The resulting RGB data item should have the same calibration as the input images. Next, in the inspector for one of the input data items, change the "adjustment" from "none" to "log". The RGB data item then becomes uncalibrated. 

*Please note that the step > change the "adjustment" from "none" to "log" > Will only work when it's done on the first selected item when the RGB computation was done.

cmeyer:
Fixes #1590.

Please review code, integrated test, and test against original issue.


KRLango (approved):
LGTM

PR #1599, opened 2025-07-16
Fix #1568. Ensure proper cursor position is shown for line plots.

### LINKED ISSUE #1568: The info section doesn't show the right value on 1D line plots
The info section doesn't show the right values as it seems to round up the X value.

STR:
Create a 1D line plot with 20 or so data points.
Create a channel graphic to make it easier to visualise
Hover the mouse or graphic over the data point # 9
In the "bar" in the 9 position from 9.0 to 9.4 > The info panel will round down the X value but will show the correct Y value.
In the "bar" in the 9 position from 9.5 to 9.9 > The info panel will round up the X value and will show the wrong Y value. The value shown will be the one from the next X position (so 10.0 to 10.4).

Image

cmeyer:


Ion-e (approved):

cmeyer:
Waiting to merge until projects re-open for patches.

PR #1641, opened 2025-08-27
Enable compression for hdf5 files. Improve chunking algorithm.

Brow71189:
- Enable compression by default for hdf5 files: Use Blosc filter with lz4 algorithm, compression level 9 and a bitshuffle filter.
- Improve chunking algorithm for a better balance between accessing the data along different axis.

### DIFF
--- a/nion/swift/model/HDF5Handler.py
+++ b/nion/swift/model/HDF5Handler.py
@@ -24,6 +24,7 @@ from nion.utils import DateTime
 from nion.utils import Geometry
 from nion.utils import Registry
 
+
 if typing.TYPE_CHECKING:
     from nion.data import DataAndMetadata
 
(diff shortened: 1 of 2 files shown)


nion/swift/model/HDF5Handler.py
```diff
@@ -41,31 +42,58 @@ def make_directory_if_needed(directory_path: str) -> None:
...
+    if len(data_shape) > 2 and (
+        (numpy.prod(data_shape[-3:]) > 8 * target_chunk_size and is_2d_data) or
+        (numpy.prod(data_shape[-2:]) > 2 * target_chunk_size)):
+        if data_shape[-1] <= 1.5 * data_shape[-2]:
+            # Assume these are 2D camera frames if the x-dimension is not much larger than the y-dimension
+            divisors[-2:] = [4, 4]
+            is_2d_data = True
+        else:
+            divisors[-1] = 4
+            is_2d_data = True
```
KRLango, line 83: Isn't this 1D data? so should this be False? If not this conflicts with line 68.

Tiomat85 (approved):

cmeyer:
I've been holding off on this a bit while I got the 16.16.1 release together.

So how important is #1643 to this issue? It seems we might want to activate this after implementing that?

Also, please consider implementing https://github.com/nion-software/nionswift-io/issues/39 which is similar functionality but much lower risk.

Brow71189:
I would still like to test this a while in-house before we roll it out. So merging right after a release would be good, that gives us time to test.
#1643 is not required for this PR to work, but it will improve the accuracy of the chunk selection algorithm.

I'll also look into https://github.com/nion-software/nionswift-io/issues/39, that should be trivial. But just just like this PR, the NHDF compression will benefit from #1643 for more reliable chunking.

Issue #1625, opened 2025-07-26
Disable the mouse scroll wheel panning images (currently up and down)


jamesrussell216:
From @bentps... Making this ticket as additional to the others listed below as it's a clearly stated main point of pain, separate from making a working scroll wheel zoom.

> Please disable the scroll wheel panning the image up and down. This is ultimate low hanging fruit. It is really annoying and should be trivial to remove this connection.
>
> Having the scroll wheel control the image zoom would be great. Don't let adding that feature distract from disabling the image pan scrolling.
>
> It would be great if the zoom point when pressing +/- (and eventually when using the scroll wheel) was around the mouse curser, instead of the center of the image.

Note: Additional adjacent issues...
https://github.com/nion-software/nionswift/issues/7
https://github.com/nion-software/nionswift/issues/924

cmeyer:
This may also affect trackpad dragging on laptops.

PR #1669, opened 2025-10-01
Disable pan action on mouse wheel scroll

KRLango:
#1625

Removes the pan call in the wheel_changed handler function. For now does not replace it with anything else


cmeyer (approved):
LGTM - any reason this is still marked as a draft?

lisham2000 (approved):
Looks good as a transition PR

PR #1672, opened 2025-10-07
Make atomic file writing more robust by using retries.

cmeyer:
There are a few Windows systems where overwriting a file isn't possible a short time after closing it. This handles that situation.


KRLango (approved):
Code looks good to me. Only question I have is have we tested that 0.35 seconds total wait time is enough? My gut feel is that it should be in most cases, but I have nothing to base that on.

PR #1696, opened 2025-10-14
fix 1695 workspace created date

### LINKED ISSUE #1695: Workspace creation date gets reset if it is the active workspace when project reloaded
Rough steps to reproduce

1. open project with multiple workspaces
2. activate an older workspace
3. modify workspace
4. exit, relaunch
5. notice workspace creation date is now updated

`pprint.pprint([(w.name, w.created) for w in _model._project.workspaces])`

cmeyer:
- **Improve handling of workspace creation date.**
- **Ensure workspace clones creation date too. Test.**

Fixes #1695.

    The first attempt Mar 2024 did not properly write out the updated
    creation date; nor add a creation date for new workspaces.
    
    This version fixes those two issues. New workspaces get a valid
    creation date. And if creation is missing it is copied from modified
    and updated when next file write takes place.

### DIFF
--- a/nion/swift/Workspace.py
+++ b/nion/swift/Workspace.py
@@ -148,6 +148,7 @@ class CloneWorkspaceCommand(Undo.UndoableCommand):
         self.__workspace_layout_uuid = workspace_controller._workspace.uuid
         self.__new_name = name
         self.__new_layout = workspace_controller._workspace.layout
+        self.__new_created = workspace_controller._workspace.created
         self.__new_workspace_id: typing.Optional[str] = None
         self.initialize()
 
@@ -159,6 +160,7 @@ class CloneWorkspaceCommand(Undo.UndoableCommand):
 
     def _perform(self) -> None:
         new_workspace = self.__workspace_controller.new_workspace(name=self.__new_name, layout=self.__new_layout, workspace_id=self.__new_workspace_id)
+        new_workspace._set_created(self.__new_created)
         self.__workspace_controller._change_workspace(new_workspace)
 
     def _undo(self) -> None:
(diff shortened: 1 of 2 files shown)


nion/swift/model/WorkspaceLayout.py
```diff
@@ -24,20 +24,44 @@ class WorkspaceLayout(Persistence.PersistentObject):
...
-        self.define_property("created", None, hidden=True, converter=Converter.DatetimeToStringConverter())
+        self.define_property("created", DateTime.utcnow(), hidden=True, converter=Converter.DatetimeToStringConverter())
         self.define_property("name", str(), hidden=True)
         self.define_property("layout", dict(), hidden=True)
         self.define_property("workspace_id", str(uuid.uuid4()), hidden=True)
+        # created was not originally available, so now we always set created when creating a new object.
+        # however, there is special logic so that if we load an old object without created, we set it
+        # when the object context is set. this variable is used to facilitate that. this scheme avoids
+        # writing the file until there is another reason to write the file.
+        self.__created_was_missing = False
```
KRLango: __created_is_missing would be a better name since this value changes to False when the created value is set.

KRLango (changes requested):
Quick test seems to work well. I think the property __created_was_missing would be better named __created_is_missing since it is referencing the current state of the workspace rather than a past one, but otherwise looks good to me.

Ion-e:
This looks good. Older projects are also organised correctly and consistently using the time of creation.
The Next/Previous project tool is now iterating according to the shown list, which also is expected behaviour.

The "Cloning workspace" hasn't been defined just yet. Do we want the newly cloned workspace to have the date of the original or the date is cloned? @bentps (This is also mentioned in https://github.com/nion-software/nionswift/issues/1695#issuecomment-3403002318)

Otherwise LGTM.

PR #1709, opened 2025-11-07
Extend thumbnail widget, canvas, and source for declarative usage.

cmeyer:
Add documentation to BitmapOverlayCanvasItemComposition and cleaned up
handling of bitmap and overlays with the set_overlay_canvas_items
method.

In ThumbnailCanvasItem, the size and overlay items are now set in
set_thumbnail_source instead of duplicated code in the init method.

Change ThumbnailWidget to make initial thumbnail source optional. This
allows declarative usages to build widget before binding. Also add a
set_thumbnail_source method to allow changing the thumbnail source.
The method passes the new thumbnail source to the ThumbnailCanvasItem.

Expand DataItemReferenceThumbnailSource to make the initial
DataItemReference optional.  This allows declarative usages to build
before binding.

### DIFF
--- a/nion/swift/DataItemThumbnailWidget.py
+++ b/nion/swift/DataItemThumbnailWidget.py
@@ -132,8 +132,15 @@ class BitmapOverlayCanvasItem(CanvasItem.AbstractCanvasItem):
 
 
 class BitmapOverlayCanvasItemComposition(CanvasItem.CanvasItemComposition):
+    """A canvas item composition that contains a bitmap canvas item and an overlay canvas item.
 
-    def __init__(self) -> None:
+    The overlay canvas item is used to handle drag and drop and focus events.
+
+    The caller can also add additional overlay canvas items using set_overlay_canvas_items that will be displayed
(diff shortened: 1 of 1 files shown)


nion/swift/DataItemThumbnailWidget.py
```diff
@@ -515,32 +526,58 @@ def unbind_display_item(self) -> None:
...
+    @data_item_reference.setter
+    def data_item_reference(self, data_item_reference: DocumentModel.DocumentModel.DataItemReference | None) -> None:
+        self.__data_item_reference = data_item_reference
+
+        if data_item_reference:
+            document_model = self.__document_model
+
+            async def async_data_item_changed() -> None:
+                data_item = data_item_reference.data_item
+                display_item = document_model.get_display_item_for_data_item(data_item) if data_item else None
```
Tiomat85: Is there a guarantee that even after that extra event_loop tick the display item will be ready? Or should this re-queue the display item check if no display item is found?

PR #1712, opened 2025-11-17
Add Datastyle Protocol to LineGraphCanvasItem.py

lisham2000:


nion/swift/LineGraphCanvasItem.py
```diff
@@ -59,90 +59,228 @@ def nice_label(value: float, precision: int) -> str:
...
-
-    ticker: Geometry.Ticker
-    if data_style == "log":
-        ticker = Geometry.LogTicker(calibrated_data_min, calibrated_data_max)
-    else:
-        ticker = Geometry.LinearTicker(calibrated_data_min, calibrated_data_max)
+    def to_display(self, value_cal: float) -> float:
+        """Convert a calibrated value to a display value."""
+        ...
+    def from_display(self, value_disp: float) -> float:
```
cmeyer: Same. `from_display` needs a more descriptive name. `convert_display_value_to_calibrated_value`?

nion/swift/LineGraphCanvasItem.py
```diff
@@ -59,90 +59,228 @@ def nice_label(value: float, precision: int) -> str:
...
+        min_specified = explicit_min_cal is not None
+        max_specified = explicit_max_cal is not None
+
+        if min_specified:
+            cal_min = typing.cast(float, explicit_min_cal)
+        else:
+            cal_min_opt: float|None = None
+            for xdata in xdata_list:
+                if xdata and xdata.data_shape[-1] > 0:
+                    un = xdata.data if numpy.issubdtype(xdata.data.dtype, numpy.floating) else xdata.data.astype(float)
```
cmeyer: `un` was changed from the descriptive `uncalibrated_data`. Why?

nion/swift/LineGraphCanvasItem.py
```diff
@@ -59,90 +59,228 @@ def nice_label(value: float, precision: int) -> str:
...
+        ic = xdata.intensity_calibration
+        if ic:
+            data = xdata.data if not xdata.is_data_rgb_type else Image.convert_to_grayscale(xdata.data)
+            cal = ic.offset + ic.scale * data
+            return DataAndMetadata.new_data_and_metadata(
+                cal,
+                intensity_calibration=Calibration.Calibration(units=ic.units),
+                dimensional_calibrations=xdata.dimensional_calibrations
+            )
+        return xdata
```
cmeyer: Again, this is the wrong point to make the abstraction. There is enough common code between the two `transform_intensity_for_plot` implementations, that it is apparent it is the wrong abstraction.

Also, `ic`, as with `un` is a bad variable name. Can you leave the original descriptive variable names the same unless they are incorrect?

nion/swift/LineGraphCanvasItem.py
```diff
@@ -59,90 +59,228 @@ def nice_label(value: float, precision: int) -> str:
...
+
+    def display_origin(self, display_min: float, display_max: float) -> float:
+        return display_min
+
+_styles: dict[str|None, DataStyle] = {
```
cmeyer: Nit: Two blank lines before top level defs and classes.

cmeyer (changes requested):
Overall this is a good structure and approach, but the PR needs some work on proper abstractions, function/variable naming, and syntax nitpicks.

I tested the functionality, but not exhaustively.

Generally, this should be easy to compare old code and new code because anywhere there is a `self.data_style == "log"` occurs should have a method. I haven't done a detailed comparison between old and new yet - and I'll wait until the review comments are addressed.

nion/swift/LineGraphCanvasItem.py
```diff
@@ -317,23 +387,8 @@ def calculate_calibrated_xdata(self, xdata: DataAndMetadata.DataAndMetadata) ->
...
-            if self.data_style == "log":
-                calibrated_data = intensity_calibration.offset + intensity_calibration.scale * data
-                calibrated_data[calibrated_data <= 0] = numpy.nan
-                numpy.log10(calibrated_data, out=calibrated_data)
-            else:
-                calibrated_data = intensity_calibration.offset + intensity_calibration.scale * data
-            return DataAndMetadata.new_data_and_metadata(calibrated_data, intensity_calibration=Calibration.Calibration(units=intensity_calibration.units), dimensional_calibrations=xdata.dimensional_calibrations)
-        else:
-            return xdata
+        style = _styles.get(self.data_style or "linear", _styles["linear"])
```
Tiomat85: This seems odd, the default value will either never get used, or fail to be read twice with a key not found?  This pattern is there in multiple places.

cmeyer:
I've added additional changes to this PR, you can see my changes in the a new PR:
- https://github.com/nion-software/nionswift/pull/1747

cmeyer:
Merged with changes 93bcf31471450eae09ccc027febc38483d48a871

PR #1721, opened 2025-12-11
Fix issue where new workspace would not save contents. Add test.

### LINKED ISSUE #1720: Workspace does not always retain images
1. create a new workspace
2. split 2x2
3. generate 3 images, which should appear in the workspace
4. switch to another workspace
5. switch back to new workspace
6. observe it does not hold 3 images anymore

cmeyer:
Fixes #1720. Also add an integration test.

Note for reviewers: the workspace handling code is old and inelegant and should be considered technical debt. This change fixes an obvious bug (not properly setting up each DisplayPanel to listen for changes), but doesn't do much to address the technical debt.


Ion-e:
Does this also fixes https://github.com/nion-software/nion-internal/issues/169?

PR #1750, opened 2026-01-20
Add implementation for IE2 

lisham2000:
using the an abstract class and then overiding the calculate Y axis. As an option as discussed on https://github.com/nion-software/nion-instrumentation/issues/576

### DIFF
--- a/nion/swift/Inspector.py
+++ b/nion/swift/Inspector.py
@@ -2213,11 +2213,11 @@ class LinePlotDisplaySectionHandler(Declarative.Handler):
 
             def convert(self, value: typing.Optional[str]) -> typing.Optional[bool]:
                 """ Convert bool to checked or unchecked string """
-                return value == "ie2"
+                return value == "log"
 
             def convert_back(self, value: typing.Optional[bool]) -> typing.Optional[str]:
                 """ Convert checked or unchecked string to bool """
-                return "ie2" if value else "linear"
+                return "log" if value else "linear"
 
         self._log_checked_to_check_state_converter = LogCheckedToCheckStateConverter()
 
--- a/nion/swift/LineGraphCanvasItem.py
+++ b/nion/swift/LineGraphCanvasItem.py
@@ -92,7 +92,7 @@ class DataStyle(ABC):
     @abstractmethod
     def adjust_calibrated_limits(self, cal_min: float, cal_max: float, min_specified: bool, max_specified: bool) -> tuple[float, float]: ...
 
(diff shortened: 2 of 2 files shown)


nion/swift/LineGraphCanvasItem.py
```diff
@@ -128,8 +197,10 @@ def adjust_calibrated_limits(self, cal_min: float, cal_max: float, min_specified
         return cal_min, cal_max
 
 
-class _LogStyle(DataStyle):
-    style_id = "log"
+class _LogStyle(DataStyle, ABC):
```
KRLango: This doesn't need to subclass ABC

nion/swift/Inspector.py
```diff
@@ -2213,11 +2213,11 @@ class LogCheckedToCheckStateConverter(Converter.ConverterLike[str, bool]):
 
             def convert(self, value: typing.Optional[str]) -> typing.Optional[bool]:
                 """ Convert bool to checked or unchecked string """
-                return value == "log"
+                return value == "ie2"
```
KRLango: This change does not make sense - the converter is "LogChecked..." but this is checking of the value is ie2

KRLango (changes requested):
This abstract class version would be my preferred option as I think it would be easier to maintain going forward if new data styles get added.

nion/swift/LineGraphCanvasItem.py
```diff
@@ -171,9 +242,132 @@ def adjust_calibrated_limits(self, cal_min: float, cal_max: float, min_specified
...
+        min_specified = data_min is not None
+        max_specified = data_max is not None
+
+        tmin_opt: typing.Optional[float] = None
+        tmax_opt: typing.Optional[float] = None
+
+        if not min_specified or not max_specified:
+            for xdata in xdata_list:
+                if xdata and xdata.data_shape[-1] > 0:
+                    tx = self.transform_intensity_for_graph(xdata)
```
cmeyer: As far as I can see, the _only_ difference between `calculate_y_axis` in `_IE2Style` and `_LinearStyle` is that the `_IE2Style` version pre-computes using `self.transform_intensity_for_graph(xdata)`. Other than that, it seems like there is a lot of duplicated code with subtly different naming and ordering of the algorithm.

Can you just generalize the original version of `calculate_y_axis` to use the `_IEStyle` method and then you don't have to change the base class nor implement a `calculate_y_axis` method?

Restated can the original `calculate_y_axis` ([these lines](https://github.com/nion-software/nionswift/blob/ab236ae06e91925675f54136759b9bf711e27e03/nion/swift/LineGraphCanvasItem.py#L196-L220)) use `self.transform_intensity_graph(xdata)`, pretty close to how `_IE2Style` implements it?

PR #1759, opened 2026-02-05
Use file path for project reference key rather than project UUID

### LINKED ISSUE #1489: Software gets in a weird state when reopening a different project with the same name
I might be mistaken on the cause of the issue, it may not be swift but windows 11 being "smart". This ticket addresses 3 potential issues:

1. Software doesn't allow to open a project with the same name as the current project.
2. Making a copy in windows explorer of an existing project > Open one of the copies > Make a change > Change to a third project > Open the other copy < You will observe that the changes made in the first copy of the project will propagate to this copy of the project. Similarly, if you delete the original copy, the data items will disappear from the copy.
3. Create a project > Place some data items in the workspace > Make a copy of the project and open the project < Observe that some of the data items will disappear from the workspace.
Thoughts? @cmeyer @jamesrussell216

cmeyer:
Fixes #1489

- **Eliminate get_project_reference_by_project_uuid, unused SwitchProjectAction.**
- **Change project references to use project_reference_id as key instead of project UUID. Tests.**

### DIFF
--- a/nion/swift/model/Profile.py
+++ b/nion/swift/model/Profile.py
@@ -126,6 +126,10 @@ class ProjectReference(Persistence.PersistentObject):
     def project_reference_parts(self) -> typing.Sequence[str]:
         raise NotImplementedError()
 
+    @property
+    def project_reference_id(self) -> str:
+        return ".".join(self.project_reference_parts)
+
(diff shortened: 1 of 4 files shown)


Ion-e (approved):
LGTM

PR #1762, opened 2026-02-09
make renaming a workspace update the title bar

### LINKED ISSUE #627: Renaming workspace does not update name in title bar

MattRoyle:
Fixes #627. The _preform and _undo functions in the RenameWorkspaceCommand now call the document controller _workspace_changed. The other commands in Workspace.py were calling this method in the command so I did the same rather than placing it in the WorkspaceRenameAction in DocumentController.

### DIFF
--- a/nion/swift/Workspace.py
+++ b/nion/swift/Workspace.py
@@ -133,9 +133,11 @@ class RenameWorkspaceCommand(Undo.UndoableCommand):
 
     def _perform(self) -> None:
         self.__workspace_controller._workspace.name = self.__new_name
+        self.__workspace_controller.document_controller._workspace_changed(self.__workspace_controller._workspace)
 
     def _undo(self) -> None:
         self.__workspace_controller._workspace.name = self.__old_name
+        self.__workspace_controller.document_controller._workspace_changed(self.__workspace_controller._workspace)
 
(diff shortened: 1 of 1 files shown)


KRLango (approved):
The code seems sensible. I am concerned there may be some side effects that are not obvious, but hopefully testing should pick that up.

Ion-e (approved):
LGTM

cmeyer:
First, let me say that the workspace code is ancient code and in addition to the poorly named classes and files, it is not up to date on notifications, etc. So this was going to be a tricky change in the first place.

With that said, I'd like to take a different approach to more closely match the "reactive" style user interface code we use. "Reactive" meaning that the UI reacts to property changes on the underlying objects rather than "active" changes to make updates, which are considerably less flexible. Kevin correctly catches this in this PR that there may be other potential places where a workspace can be renamed and this code would not handle that properly.

I've made another PR which takes this more flexible, although more complicated, approach. I decided that the description of this problem and the solution was more concise in actual code rather than trying to explain the details via text. Hopefully the new PR is instructive.

So, I'm closing this PR in favor of the new one. The new PR also includes an integration test which is more or less mandatory for changes like this.

Superseding PR:
- https://github.com/nion-software/nionswift/pull/1765

Issue #1717, opened 2025-12-09
Spectrum interval text difficult to read for ZLP width measurement


jamesrussell216:
From Ben...

--

When we're doing high energy resolution EELS (which some would say is our flagship feature), we're constantly looking at a live spectrum with a "Live ZLP measurement" shown, which is an interval that moves around as a fit is done to the ZLP. The width of the interval is shown between the two interval edges. When the interval is narrow (which means you have high resolution yay!), the text telling you what the resolution is becomes obscured.

--

For illustrative purposes it ends up looking a bit like this...

<img width="1408" height="935" alt="Image" src="https://github.com/user-attachments/assets/d7770ee2-a4c3-4cf4-a8b1-131f358b5be3" />

Where it's difficult to easily read the "1.2" text because of the width of the interval. I'd guess maybe moving the text object a bit higher and putting a white background on it would do the job for instance.

PR #1764, opened 2026-02-12
improve interval readability

MattRoyle:
Addresses the initial part of issue #1717, to have the current values be more readable. This changes the display of the width label so it has is breaks the bounding lines. The label, left, right, and width text have been adjusted in position for legibility.

image

### DIFF
--- a/nion/swift/LineGraphCanvasItem.py
+++ b/nion/swift/LineGraphCanvasItem.py
@@ -15,6 +15,7 @@ import gettext
 import math
 import re
 import typing
+import warnings
 
 # third party libraries
 import numpy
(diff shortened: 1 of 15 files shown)


nion/swift/LineGraphCanvasItem.py
```diff
@@ -953,6 +955,39 @@ def convert_coordinate_to_pixel(canvas_size: Geometry.IntSize, c: float, data_sc
                 px = c * data_scale
                 return canvas_size.width * (px - data_left) / (data_right - data_left)
 
+            @dataclasses.dataclass
+            class TextRectangle:
```
KRLango: You should use `Geometry.FloatRect` class instead of creating a new class for this.

MattRoyle:
This edits the same files as #1771, I have checked the edits and the two PRs are changing different parts of the file, there are no shared lines so merging shouldn't be too bad

cmeyer (changes requested):
Overall, looks good. But can you calculate the width by substituting '0' for each of the numbers? The problem is going from something like 109.9 to 111.1 to 122.2 will split the interval lines (109.9), then won't (111.1) since '1' is narrowing in some fonts, then will (122.2). This can create a strange on/off flickering during dragging.

cmeyer:
See also https://github.com/nion-software/nionswift/discussions/1780

nion/swift/LineGraphCanvasItem.py
```diff
@@ -955,6 +954,61 @@ def convert_coordinate_to_pixel(canvas_size: Geometry.IntSize, c: float, data_sc
...
+
+            def _padded_number_rectangle(text: str, x: float, y: float, text_baseline: str) -> Geometry.FloatRect:
+                """Gets the text rectangle for numbers
+
+                Replaces the digits with 0 so the size doesn't flicker between numbers
+                """
+                replaced_text: str = ''
+                for letter in text:
+                    replaced_text += letter if not letter.isdigit() else '0'
+                return _get_text_rectangle(replaced_text, x, y, text_baseline, 2)
```
cmeyer: I actually think this looks better with `margin=-2` (yes `-2`). Doing that avoids the visually unappealing unnecessary breaking of the lines below:

---

With 2:

Image

---

With -2:

Image

---

And a narrower interval with -2:

Image

PR #1765, opened 2026-02-13
Fix title updating when renaming workspace

### LINKED ISSUE #627: Renaming workspace does not update name in title bar

cmeyer:
Fixes #627

- **Trigger property changed events on WorkspaceLayout.**
- **Observe workspace controller and name changes to update window title. Test.**

General approach is to add property changed events to the workspace layout, add workspace layout changed events to workspace controller, listen for workspace layout changes, connect property changed event to workspace layouts, and watch for name changes. Adds a test, too.

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -278,7 +278,6 @@ class DocumentController(Window.Window):
         if project_reference:
             self.window_file_path = project_reference.path
         self.__workspace_controller: typing.Optional[Workspace.Workspace] = None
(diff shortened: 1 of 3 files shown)


Ion-e (approved):
[Issue](https://github.com/nion-software/nionswift/issues/627) verified as fixed.

It looks good now, cloning a workspace with the same name creates the workspace with the same name and new date, which seems correct in my opinion.

I can't think of any other way a user could have access to worksapce name-changing tools.

PR #1770, opened 2026-02-16
Long computation stop and progress capability

cmeyer:
- **Add sync_mode/auto_update flag to computation. No UI.**
- **Add persistent last computed status. Make last computed time and duration persistent.**
- **Introduce computation executor context/parameters for progress, stop capabilities.**
- **Add computation stop/progress to edit computation dialog.**

Please review. This effectively enables long computations, although there are only a few operations which currently support it, such as Processing > Reduce > Mapped Sum.

This only implements a basic UI in the Processing > Edit Computation dialog. The UI includes a auto-update checkbox, an update button, progress/stop buttons for longer computations.

Please review the PR. Once this is merged, I'll submit follow-up PR's to change most processing to utilize these new capabilities.

### DIFF
--- a/nion/swift/ComputationPanel.py
+++ b/nion/swift/ComputationPanel.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 # standard libraries
+import asyncio
 import datetime
 import functools
 import gettext
@@ -8,6 +9,7 @@ import operator
 import random
 import string
 import sys
+import time
 import typing
 import uuid
 import weakref
@@ -44,6 +46,7 @@ from nion.utils import Model
 from nion.utils import Observable
 from nion.utils import ReferenceCounting
 from nion.utils import Registry
+from nion.utils import Stream
 
 if typing.TYPE_CHECKING:
     from nion.swift import DocumentController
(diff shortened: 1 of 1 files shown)


nion/swift/model/Symbolic.py
```diff
@@ -2928,14 +2992,119 @@ def get_source_metadata(self) -> typing.Optional[Persistence.PersistentDictType]
...
+
+    def sync_execution(self) -> None:
+        # executors should call this continuously to allow sync, scheduling, and pausing.
+        if self.__is_canceled:
+            raise ComputationCanceledException()
+
+    def __handle_stop(self) -> None:
+        self.__is_canceled = True
+
+    def set_progress(self, progress: float) -> None:
```
Brow71189: Why is there a method for `set_progress` instead of a `@progress.setter` for the `@progress` attribute?

cmeyer: I've changed it to use a property setter. I only had it that way for code migration and I agree it's better to have it as a property.

cmeyer:
> I added some comments inline, mostly about making sure we don't loose functionality compared to the old computations.

Another note: We will most definitely _not_ lose functionality with the upcoming changes.

PR #1771, opened 2026-02-17
IE2 maintain singular calculate_y_axis

lisham2000:


KRLango (commented):
I'm not sure if this is ready for a review or not yet, but I took a quick look anyway and thought I would raise a few points

cmeyer:
@lisham2000 This PR will have to be redone once display computations are in place. Can we close it in favor of a future PR?

PR #1773, opened 2026-02-18
Add ability to unlock aspect ratio on 1d svg exports

### LINKED ISSUE #1766: Export to SVG should allow export to aspect ratios other than 5:3
Needs UI design.

lisham2000:
Fixes #1766

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -417,7 +417,7 @@ class ExportSizeModel(Observable.Observable):
             if unit_description_.unit_id == unit_id_model.value:
                 unit_description = unit_description_
                 break
-        self.unlock_aspect_ratio = Model.PropertyModel(True)
+        self.aspect_ratio_locked = Model.PropertyModel(True)
         self.is_line_plot = (display_item.used_display_type == "line_plot")
         self.__width_quantity = get_quantity_from_pixels(display_size.width, unit_description)
         self.__height_quantity = get_quantity_from_pixels(display_size.height, unit_description)
@@ -449,20 +449,14 @@ class ExportSizeModel(Observable.Observable):
 
         If the primary field is width or aspect ratio unlocked, return the last quantity value directly.
 
-        Otherwise, calculated the width from the last quantity value and the aspect ratio.
+        Otherwise, calculate the width from the last quantity value and the aspect ratio.
 
         If the last quantity is in pixels, round the value.
         """
-        if not self.unlock_aspect_ratio.value:
-            return round(self.__width_quantity.value) if self.__width_quantity.is_pixels else self.__width_quantity.value
-
-        if self.__primary_field == 'width':
-            width_quant = self.__width_quantity
-            return round(width_quant.value) if width_quant.is_pixels else width_quant.value
-        else:
-            width_quant = self.__height_quantity
-            width_aspect = width_quant.value * self.__aspect_ratio
-            return round(width_aspect) if width_quant.is_pixels else width_aspect
+        width_quant = self.__width_quantity
+        if self.aspect_ratio_locked.value and self.__primary_field == 'height':
+            width_quant = self.__width_quantity.with_value(self.__height_quantity.value * self.__aspect_ratio)
+        return round(width_quant.value) if width_quant.is_pixels else width_quant.value
 
     @property
     def _height(self) -> float:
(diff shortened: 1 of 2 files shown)


nion/swift/ExportDialog.py
```diff
@@ -344,9 +344,10 @@ def __init__(self, display_item: DisplayItem.DisplayItem, unit_id_model: Model.P
             if unit_description_.unit_id == unit_id_model.value:
                 unit_description = unit_description_
                 break
-        min_pixels = inch_unit_description.convert_value_to_pixels(3.0)
-        max_pixels = inch_unit_description.convert_value_to_pixels(12.0)
-        self.__last_quantity = get_quantity_from_pixels(max(min_pixels, min(max_pixels, display_size.width)), unit_description)
+        self.unlock_aspect_ratio = Model.PropertyModel(True)
```
KRLango: I would invert this rather than have the negative "not locked" to makle the logic easier to follow. I would also change the name to `apect_ratio_locked` since `unlock_aspect_ratio` sounds like a command not a variable.

Also would it be a good idea to default to "unlocked" for line_plots, and "locked" for everything else?

nion/swift/ExportDialog.py
```diff
@@ -373,61 +374,76 @@ def calibrated_shape_str(self) -> typing.Optional[str]:
...
+            width_quant = self.__width_quantity
+            return round(width_quant.value) if width_quant.is_pixels else width_quant.value
         else:
-            if self.__primary_field == 'width':
-                return self.__last_quantity.value
-            else:
-                return self.__last_quantity.value * self.__aspect_ratio
+            width_quant = self.__height_quantity
+            width_aspect = width_quant.value * self.__aspect_ratio
+            return round(width_aspect) if width_quant.is_pixels else width_aspect
```
KRLango: This line has been duplicated 3 times in this function. Would a better approach be something like:

```
quant=self.__width_quantity
if not self.unlock_aspect_ratio and self._primary_field == 'height':
    quant = Quantity(self.__height_quantity.value * self.__aspect_ratio, self.__height_quantity.unit_description)
return round(quant.value) if quant.is_pixels else quant.value
```

Also - is there ever a reason not to do the rounding in the Quantity class itself? That would simplify the code here a lot as well.

nion/swift/ExportDialog.py
```diff
@@ -373,61 +374,76 @@ def calibrated_shape_str(self) -> typing.Optional[str]:
...
         Otherwise, calculated the height from the last quantity value and the aspect ratio.
 
         If the last quantity is in pixels, round the value.
         """
-        if self.__last_quantity.is_pixels:
-            if self.__primary_field == 'height':
-                return round(self.__last_quantity.value)
-            else:
-                return round(self.__last_quantity.value / self.__aspect_ratio)
+        if not self.unlock_aspect_ratio.value:
```
KRLango: Same comments as with width property - this if statement seems to be the wrong way round and there is a lot of duplication in this function.

nion/swift/ExportDialog.py
```diff
@@ -440,21 +456,36 @@ def placeholder_width_text(self) -> str | None:
...
-        if value_str:
-            value = self.__float_to_string_converter.convert_back(value_str) or 0.0
-            self.__last_quantity = self.__last_quantity.with_value(value)
+        if value_str is None:
+            return
+        value = self.__float_to_string_converter.convert_back(value_str) or 0.0
+
+        self.__primary_field = 'height'
+
+        if not self.unlock_aspect_ratio.value:
```
KRLango: The height gets updated in the same way in both the if and the else - it should be moved outside of the if statement and then the if statement can get simplified.

KRLango (changes requested):
The logic seems off to me in a few places, but I think that may just be a badly named variable - `unlock_aspect_ratio`. Where I have suggested alternative code I have kept the current name and usage.

lisham2000:
> The checkbox is nor persisten between exports.

Non persistant between which exports?

cmeyer (changes requested):
A few notes:
- when deselecting the 'linked' checkbox, can you populate the missing field text? i.e. if 'width' is 20, height is 12 (but dimmed), checking 'linked' checkbox should fill in '12' as the value.
- I'm indifferent as to the terminology "linked" or "locked" (but definitely NOT "unlocked"). can you make the code and the UI agree in naming? as Kevin says, the code should probably be either "linked" or "locked", again NOT "unlocked" - and it should match what the UI is using.
- the model in this case is getting complicated. can you include unit tests for the model. particularly, enabling the field disabled when you disable the 'link/lock' state should be testable. you can build up additional tests as you find bugs or as the dialog changes in behavior due to UI design.

EDIT:

I think we should settle on the term "lock aspect ratio" as that seems to be standard. We may replace that by a chain/broken-chain icon in the future.

We might also consider eliminating the checkbox and providing a "Aspect Ratio: <combo-box-drop-down>" with some standard aspect ratios [16:9, 5:3, 1:1, Custom].

Tiomat85:
> How does this look for what we spoke to about rather than having the checkbox? Having a combo box instead and having a 'Custom' option for the unlocked option aswell as common aspect ratios? @cmeyer @KRLango @Tiomat85

I think that is an improvement, the standard options are good as well.  Something feels off with the greyed out Height box but I am struggling to put my finger on what/why.  It might be because you are disabling a previous UI element depending on a dropdown that comes after it.

Fear my paint skills:
image

cmeyer:
I'm not seeing the changes with the new combo box for aspect ratio. The PR as it is today still shows a dialog below. Are you planning to update this PR?

image

Also, can you move the tests to the newly available `ExportDialog_test.py` file?

Also, you may consider `git rebase master` and then force push your branch with `git push --force-with-lease`.

cmeyer (changes requested):
This is getting closer... I like the aspect combo box. But...

The 'original' aspect ratio item in the combo box when rendering line plots is confusing. Maybe it should be 'Default (5:3)' or eliminated altogether?

Also, do we want to allow differing aspect ratios on images? I think it would be reasonable to always go with the image's aspect ratio. ...maybe with a limit of something like 1:10 or 10:1 to handle the case of someone trying to export 1024x2 data?

Ion-e:
Bug1: when adding any multiple of 10 in the edit box whilst the width/height is linked, both edit boxes will default to 0.

cmeyer:
Squashed/merged 0d18991cc676d34fb1e4ffe5a7b72c65006439f3. Thanks!

PR #1791, opened 2026-03-03
Use one as lower y-axis value when display limit negative and log display.

### LINKED ISSUE #1790: Log display has incorrect upper display limit if lower display limit is negative
Use `show(numpy.random.randn(100) + 5 * 1e6)`. Drag the y-axis on the resulting line plot upwards to establish a negative lower display limit and a valid upper display limit. Now switch to Log. Notice the displayed limits are -1 to 1, which is incorrect.

cmeyer:
Fixes #1790

The value of one is a compromise. This usually results from the user adjusting
the y-axis on non-log display upward to be able to see the lower valued data
and then enabling log display. The value of one will be values less than one
and negative values at the bottom of the display.

Example display limits

image

Example of original data

image

Example of incorrect log display (displaying -1 to 1) (existing code)

image

Example of "correct" log display (display 1 to non-log display limit 8e7) (new code, this PR)

image


cmeyer:
This is part of an ongoing set of changes to the line plot and fixes only one minor bug.

PR #1804, opened 2026-03-13
Add Workspace from selection

### LINKED ISSUE #1805: Create a new workspace from selected items
### Discussed in https://github.com/nion-software/nionswift/discussions/1793
Create a new workspace from selected items. The function requires being able to split a workspace and fill it with selected items which seems like a useful standalone function to have separate as well.

MattRoyle:
Fixes #1805
New Workspace from Selection - Creates a new workspace and populates the display layout with the selected items
The action is in the Workspace menu drop down as well as when you right click in the data panel or workspace display panel. 
The splitting of the workspace uses the number of selected items and finds the closest split to a 5:3 ratio. If the number of selected items is not exact the result will have some empty display panels at the end.
The maximum allowed split selection is 60 creating a 10x6 split.
The drop down menu name contains the outcome of the split, i.e. 2x2 as well as the number of items in the selection.

Workspace Menu from Data Panel Selection of 5 items, New Workspace from Selection is available.
image

Workspace Menu when disabled.
Screenshot 2026-04-15 105042

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -2793,6 +2793,7 @@ class DocumentController(Window.Window):
         self.add_action_to_menu_if_enabled(menu, "display.reveal", action_context)
         self.add_action_to_menu_if_enabled(menu, "file.export", action_context)
         self.add_action_to_menu_if_enabled(menu, "file.export_batch", action_context)
+        self.add_action_to_menu_if_enabled(menu, "workspace.new_workspace_from_selection", action_context)
 
         def reveal_data_item(data_item: DataItem.DataItem) -> None:
             # same as selecting menu item for 'reveal'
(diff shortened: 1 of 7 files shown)


MattRoyle:
The bug #1809, where multiple Undoable commands are added to the stack when a Line plot display is selected will impact the undo function on the new workspace from selection.

nion/swift/DocumentController.py
```diff
@@ -3814,6 +3814,65 @@ def execute(self, context: Window.ActionContext) -> Window.ActionResult:
         return super().execute(context)
 
 
+class NewFromSelectionAction(WorkspaceNewAction):
+    action_id = "workspace.new_workspace_from_selection"
+    action_name = _("New Workspace From Selection")
+
+    @staticmethod
+    def _get_selected(context: DocumentController.ActionContext) -> list[DisplayItem.DisplayItem]:
```
cmeyer: 'selected' is too general and is also an adjective, which is a strange choice. Can you name it as a noun and make it descriptive: 'selected_display_items'? Also 'selection' -> 'selected_display_items'.

But... why is this needed at all? We discussed earlier not having a weird new selection strategy (the new strategy being selected display panels _combined_ with select items in the data panel). Let's just use `document_controller.selected_display_items` instead. Is that not sufficient?

nion/swift/DocumentController.py
```diff
@@ -3814,6 +3814,65 @@ def execute(self, context: Window.ActionContext) -> Window.ActionResult:
...
+
+    def is_enabled(self, context: Window.ActionContext) -> bool:
+        context = typing.cast(DocumentController.ActionContext, context)
+        selection = self._get_selected(context)
+        return bool(context.focus_widget) and bool(selection) and len(selection) < 101
+
+    def get_action_name(self, context: Window.ActionContext) -> str:
+        context = typing.cast(DocumentController.ActionContext, context)
+        selection = self._get_selected(context)
+        if not bool(context.focus_widget) or not bool(selection) or len(selection) > 100:
```
cmeyer: I think 100 is too large - let's make it considerably smaller, maybe something like 36.

And can you put that into a constant? `MAX_PANELS: typing.Final[int] = 36`?

And can you update the `< 101` to use the same constant: `<= MAX_PANELS`?

Better yet: change line 3859 to be `if self.is_enabled:` so it only does that check in one place?

nion/swift/Workspace.py
```diff
@@ -48,6 +49,129 @@ def create_splitter_desc(orientation: str, splits: typing.Sequence[float], child
...
+
+    def _set_modified_state(self, modified_state: typing.Any) -> None:
+        self.__workspace_controller._project.modified_state = modified_state
+
+    def _perform(self) -> None:
+        horizontal, vertical = self.__layout_shape
+        if not (horizontal == 1 and vertical == 1):
+            display_panels = self.__workspace_controller.apply_layouts(self.__selected_panel, self.__display_panels, horizontal, vertical)
+            self.__undo_command = self.__workspace_controller.document_controller.last_undo_command
+            self.__workspace_controller.document_controller.pop_undo_command()
```
cmeyer: Doing `pop_undo_command` is incorrect here.

The proper solution is to refactor `apply_layouts` into its underlying mutation of the workspace and the command handling; then call the underlying mutation code.

Undo commands, unfortunately, have a lot of variations. The one used (`ChangeWorkspaceContentsCommand`) in `apply_layouts` expects that it is constructed before the layout is changed so it can save the "before" layout. The same technique used within `apply_layouts` can be used here.

To summarize:

1. move the command handling in `apply_layouts` into `DocumentController.py`/`WorkspaceSplitAction`
2. move the bulk of this code (i.e. `_perform`) into a workspace method.
3. delete this class `SplitFromSelectionCommand`
4. put undo handling for this code into `NewWorkspaceFromSelectionCommand`
5. review `NewWorkspaceFromSelectionCommand` to see if the same technique can be applied there

nion/swift/Workspace.py
```diff
@@ -48,6 +49,129 @@ def create_splitter_desc(orientation: str, splits: typing.Sequence[float], child
...
+        ceil_diff = columns_ceil * rows_ceil - selection_count
+        floor_diff = columns_floor * rows_floor - selection_count
+
+        if floor_diff < ceil_diff:
+            return columns_floor, rows_floor
+
+        return columns_ceil, rows_ceil
+
+
+class NewWorkspaceFromSelectionCommand(Undo.UndoableCommand):
```
cmeyer: Ideally, this can just be eliminated. If you need additional functionality, perhaps expand `CreateWorkspaceCommand` command instead? The `_perform` just calls another command, which is strange; it should just call a workspace method instead.

nion/swift/Workspace.py
```diff
@@ -48,6 +49,129 @@ def create_splitter_desc(orientation: str, splits: typing.Sequence[float], child
...
+    def _undo(self) -> None:
+        assert self.__undo_command is not None
+        self.__undo_command.undo()
+
+    def _redo(self) -> None:
+        assert self.__new_layout is not None
+        self.__workspace_controller.reconstruct(self.__new_layout)
+
+    @classmethod
+    def get_split(cls, selection_count: int) -> typing.Tuple[int, int]:
```
cmeyer: The `get_split` method should either be standalone or in the Workspace class, not in a `UndoableCommand` class.

cmeyer (changes requested):
I might be repeating myself, but the tests are too complicated. We generally don't have large supporting functions for tests.

How about writing a one self-contained test (or a very small number of them) that is representative of what a user would do? Then if we encounter bugs going forward, write test cases for those bugs.

Generally, our approach to testing isn't to comprehensively test every possible combination of cases. Instead, it is to verify that basic functionality works and then test by exception after that. The tests are not only used to verify _basic_ functionality, but to demonstrate how the functions should be called. Building complex infrastructure for testing works directly against that goal. These tests, if left "as is", will be very difficult to modify and they're already virtually impossible to understand without delving into the complex supporting functions.

You might consider putting these tests into a personal branch so if you want to work on the code again, you have them available, but they probably aren't generally useful going forward due to their complex nature.

nion/swift/DocumentController.py
MattRoyle: I've changed it to be 60 since the maximum for w or h in the apply_layouts that will work is 10, so a w of 10 at 5:3 ratio would be 60. 
image

I think it looks reasonable but could go lower if you think there are too many.

MattRoyle:
> I might be repeating myself, but the tests are too complicated. We generally don't have large supporting functions for tests.
> 
> How about writing a one self-contained test (or a very small number of them) that is representative of what a user would do? Then if we encounter bugs going forward, write test cases for those bugs.
> 
> Generally, our approach to testing isn't to comprehensively test every possible combination of cases. Instead, it is to verify that basic functionality works and then test by exception after that. The tests are not only used to verify _basic_ functionality, but to demonstrate how the functions should be called. Building complex infrastructure for testing works directly against that goal. These tests, if left "as is", will be very difficult to modify and they're already virtually impossible to understand without delving into the complex supporting functions.
> 
> You might consider putting these tests into a personal branch so if you want to work on the code again, you have them available, but they probably aren't generally useful going forward due to their complex nature.

I've made changes to reduce the compexity and bring in some of your suggestions.
The selection now is determined by the existing selected_display_items rather than the get_selection function which reduced the test complexity.
I also removed the testing of the shape since that is only testing the functionality of apply\_layouts which should be tested separately.
The CreateWorkspaceFromSelctionCommand now sub-classes CreateWorkspaceCommand.
I removed the SplitFromSelectionCommand, moving the splitting logic into the CreateWorkspaceFromSelctionCommand perform function.
I've also moved the get split function from the action into the workspace.
The command that was previously added in apply_layout has been moved to the WorkspaceSplitAction in order to get rid of the popping of the undo stack.

MattRoyle:
@cmeyer During swiftcon we discussed how the UI should look for this. I believe you said to not make it a seperate menu item, instead having it be a checkbox within the new workspace popup dialog?

cmeyer:
Squashed and merged 7d088e22f435b351387bdb10495593bc6d93a684

PR #1808, opened 2026-03-16
Disable Edit Title for invalid selections

### LINKED ISSUE #1807: Edit Title fails when multiple selected
Separated from #1806. Edit Title is still in the context menu when you have multiple selected data items and will silently fail. It would make sense to disable it when there is more than one item in the selection.

MattRoyle:
Fixes #1807. Adds the is_enabled method to verify there is a display item and there is only one display item selected.

### DIFF
--- a/nion/swift/DocumentController.py
+++ b/nion/swift/DocumentController.py
@@ -3395,7 +3395,7 @@ class OpenTitleEditAction(Window.Action):
     def invoke(self, context: Window.ActionContext) -> Window.ActionResult:
         context = typing.cast(DocumentController.ActionContext, context)
         window = typing.cast(DocumentController, context.window)
-        display_item = typing.cast(DisplayItem.DisplayItem, context.display_item)
+        display_item = context.display_item
         if window and display_item:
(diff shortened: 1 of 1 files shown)


nion/swift/DocumentController.py
```diff
@@ -3411,6 +3411,10 @@ def invoke(self, context: Window.ActionContext) -> Window.ActionResult:
             return Window.ActionResult(Window.ActionStatus.FINISHED)
         return Window.ActionResult(Window.ActionStatus.PASS)
 
+    def is_enabled(self, context: Window.ActionContext) -> bool:
+        context = typing.cast(DocumentController.ActionContext, context)
+        return bool(context.display_item)  # context.display_item will be None when there's selection or the selection contains more than one item
```
cmeyer: Can you change this to `context.display_item is not None` for clarity? The `bool` case is not idiomatic and is prone to error in cases where it's argument has defined `__bool__`. We're really checking whether it is `None` - not whether it can be converted to `bool` and `bool` returns True.

PR #1835, opened 2026-04-19
Consolidate data calibration handling, particularly in inspector and histogram

cmeyer:
- **Remove unused DisplayItem methods.**
- **Move DisplayCalibrationInfoStream to DisplayItem for general use.**
- **Improve typing of CalibrationStyle and DisplayCalibrationInfo.**
- **Fix error is DisplayCalibrationInfo.__ne__.**
- **Update CalibrationStyleModel to utilize DisplayCalibrationInfo.**
- **Rework GraphicsSectionHandler to use CalibrationStyleModel.**
- **Change histogram cursor handling to use DisplayCalibrationInfo.**
- **Introduce a DisplayItem.display_calibration_info method. Use for 'data info'.**
- **Consolidate most calibration results into DisplayCalibrationInfo.**
- **Replace DisplayItemCalibratedValueModel with DisplayItemCalibratedDimensionValueModel.**
- **Refactor CalibratedBinding and subclasses to use display_calibration_info stream.**
- **Move some common code to display_calibration_info methods.**
- **Refactor CalibratedLengthBinding to use display_calibration_info stream.**
- **Refactor remaining inspector cases display_calibration_info.**
- **Remove unused display_property_changed_event and related code.**
- **Eliminate DisplayCalibrationInfoStream in favor of a one-use FollowStream in HistogramPanel.**
- **Eliminate DisplayTypeMonitor by merging it into DisplayTracker class.**
- **Consolidate DisplayTracker into DisplayPanel for simplification.**
- **Move frame rate handling inside thread.**

This is an internal change that should not have any significant behavior or performance chances. It is a ongoing project to ensure that any data access for display is done on a thread.

This change updates all access to `DisplayCalibrationInfo` to go through a small number of interfaces which will eventually be replaced by a single stream that is only updated on a thread.

This change touches much of the calibration UI in the inspector, the histogram, cursor handing, and both the line plot and image displays.

This change also consolidates the utility classes `DisplayTypeMonitor` and `DisplayTracker` into `DisplayPanel`. This simplifies the code and allows for a more unified changes going forward.

### DIFF
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -1027,90 +1027,6 @@ class MissingDisplayCanvasItem(DisplayCanvasItem.DisplayCanvasItem):
         return False
 
 
-class DisplayTracker:
-    """Tracks messages from a display and passes them to associated display canvas item."""
-
-    def __init__(self, display_item: DisplayItem.DisplayItem, ui_settings: UISettings.UISettings,
-                 delegate: DisplayCanvasItem.DisplayCanvasItemDelegate, event_loop: asyncio.AbstractEventLoop,
-                 draw_background: bool) -> None:
-        self.__display_item = display_item
-        self.__ui_settings = ui_settings
-        self.__delegate = delegate
-        self.__event_loop = event_loop
-        self.__draw_background = draw_background
-        self.__closing_lock = threading.RLock()
-
-        # callbacks
-        self.on_clear_display: typing.Optional[typing.Callable[[], None]] = None
-        self.on_title_changed: typing.Optional[typing.Callable[[str], None]] = None
-        self.on_replace_display_canvas_item: typing.Optional[typing.Callable[[DisplayCanvasItem.DisplayCanvasItem], None]] = None
-
-        def clear_display() -> None:
-            if callable(self.on_clear_display):
-                self.on_clear_display()
-
-        def display_item_property_changed(key: str) -> None:
-            if key == "displayed_title":
-                if callable(self.on_title_changed):
-                    self.on_title_changed(display_item.displayed_title)
-
-        self.__display_about_to_be_removed_event_listener = display_item.about_to_be_removed_event.listen(clear_display)
-        self.__display_property_changed_event_listener = display_item.property_changed_event.listen(display_item_property_changed)
-
-        # ensure data stays in memory while displayed
-        display_item.increment_display_ref_count()
-
-        # create a canvas item and add it to the container canvas item.
-
-        self.__display_canvas_item = create_display_canvas_item(display_item, ui_settings, delegate, event_loop, draw_background=self.__draw_background)
-
-        def handle_display_data_delta(display_data_delta: DisplayItem.DisplayDataDelta | None) -> None:
-            # This is called when the display data delta stream produces a new value. The value is passed on to the
-            # display canvas item.
-            with self.__closing_lock:
-                assert display_data_delta
-                self.__display_canvas_item.update_display_data_delta(display_data_delta)
-
-        self.__display_data_delta_stream = display_item.display_data_delta_stream
-        self.__display_data_delta_stream_action = Stream.ValueStreamAction(self.__display_data_delta_stream, handle_display_data_delta)
-        self.__display_type: str | None = None
-
-        def handle_display_changed(display_item: DisplayItem.DisplayItem) -> None:
-            # This is called in response to a display change event, which may indicate a change in the display type.
-            # If the display type changes, update the display canvas item and the display data delta stream listener.
-            display_type = display_item.used_display_type
-            if self.__display_type != display_type:
-                self.__display_type = display_type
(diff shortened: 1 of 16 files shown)


nion/swift/LinePlotCanvasItem.py
```diff
@@ -390,6 +390,21 @@ def legend_entries(self) -> typing.List[LineGraphCanvasItem.LegendEntry]:
...
+    @property
+    def frame_index(self) -> int:
+        for display_values in self.__display_values_list:
+            if display_values:
+                data_metadata = display_values.element_data_metadata
+                if data_metadata:
+                    # allow registered metadata_display components to populate a dictionary
+                    # the line plot canvas item will look at "frame_index"
+                    d: Persistence.PersistentDictType = dict()
+                    for component in Registry.get_components_by_type("metadata_display"):
```
Tiomat85, line 402: Question: Why is this in the registry, the method called is just a static converter from MetadataType to a dictionary?

KRLango (approved):
Looks OK to me

nion/swift/DisplayPanel.py
```diff
@@ -2490,73 +2411,107 @@ def set_display_item(self, display_item: typing.Optional[DisplayItem.DisplayItem
...
-            if display_item:
-                def clear_display() -> None:
-                    self.set_display_panel_display_item(None, True)
-
-                def handle_title_changed(title: str) -> None:
+        if display_item != self.__display_item:
+            document_controller = self.__document_controller
+            document_model = document_controller.document_model
+
+            def add_display_controls(display_canvas_item: DisplayCanvasItem.DisplayCanvasItem) -> None:
```
Tiomat85, line 2418: This is a lot of functions-inside-functions. While I don't think this is going to be called enough to cause performance issues, it makes it quite hard to follow. Is there a reason these could not be promoted to normal private methods?

nion/swift/LinePlotCanvasItem.py
cmeyer: - https://github.com/nion-software/nionswift-instrumentation-kit/blob/47c017b98398fb031a2bcc310284fac0bb232cf1/nion/instrumentation/HardwareSource.py#L2022
- https://github.com/nion-software/nionswift-instrumentation-kit/blob/47c017b98398fb031a2bcc310284fac0bb232cf1/nion/instrumentation/HardwareSource.py#L2193

cmeyer: It uses a registry so that hardware components can extract metadata such as 'high tension' without hard-coding at the `nionswift` level.

nion/swift/Inspector.py
```diff
@@ -2510,343 +2489,306 @@ def convert_back(self, value_str: typing.Optional[str]) -> typing.Optional[float
...
+                    new_value = calibration.convert_from_calibrated_size(value) / data_size / self.__factor
+                else:
+                    new_value = calibration.convert_from_calibrated_value(value) / data_size / self.__factor
+        value_tuple = self.__property_model.value
+        new_value_tuple = value_tuple
+        if value_tuple is not None and new_value is not None:
+            value_list = list(value_tuple)
+            value_list[self.__dimension_index] = new_value
+            new_value_tuple = tuple(value_list)
+        if new_value_tuple != value_tuple:
```
Tiomat85: This check only applies if the previous if statement on 2738 happened, so it is conditional on that.  Assuming this is the correct condition how about:

```
value_tuple = self.__property_model.value
if value_tuple is not None and new_value is not None:
    if value_tuple[self.__dimension_index] != new_value:
        value_list = list(value_tuple)
        value_list[self.__dimension_index] = new_value
        self.__property_model.value = tuple(value_list)
```

nion/swift/Inspector.py
```diff
@@ -2510,343 +2489,306 @@ def convert_back(self, value_str: typing.Optional[str]) -> typing.Optional[float
...
+            data_size = calibration_and_data_size.data_size
+            value = Converter.FloatToStringConverter().convert_back(value_str)
+            if value is not None:
+                if self.__is_size:
+                    new_value = calibration.convert_from_calibrated_size(value) / data_size / self.__factor
+                else:
+                    new_value = calibration.convert_from_calibrated_value(value) / data_size / self.__factor
+        value_tuple = self.__property_model.value
+        new_value_tuple = value_tuple
+        if value_tuple is not None and new_value is not None:
```
Tiomat85: We do this even if the value has not changed?

Tiomat85 (commented):
Couple of bits to look at.

nion/swift/DisplayPanel.py
cmeyer: No reason, I was mostly just trying to leave the existing structure the same.

nion/swift/Inspector.py
cmeyer: I'll revisit this logic.

cmeyer: It looks like these checks are for type checking, since they are obviously not doing anything. I'll rewrite and comment to make this more clear.

PR #1836, opened 2026-04-21
Restructure image canvas item state tracking.

cmeyer:
This change restructures the image canvas item so that there is a
single info object to represent the state instead of having multiple
copies of parts of the state. This simplifies updating so that it can
occur when compositing is done and enables threaded updates.


cmeyer:
Superseded by #1859

PR #1849, opened 2026-04-27
Add verify filename utility function

### LINKED ISSUE #1848: Function to give more detailed invalid filename errors
The errors displayed for invalid filenames are not specific which doesn't help the user know what is wrong with the name given.

In a number of places the filename is checked by comparing the filename to the result of Utility.simplify_filename
https://github.com/nion-software/nionswift/blob/2cde63c489a40de54dfab33034069abf4c281da8/nion/swift/Application.py#L1044
A function that gives more specific errors such as 'Filename contains "*"' or 'Filename cannot be "CON"' would be beneficial

MattRoyle:
Fixes #1848.
Adds a utility function verify_filename_is_legal which checks a filename and returns a relevant error message if it is illegal.
The function checks run are identical to simplify_filename but will give specific error messages if there are any.

The PR changes the usage of simplify filename in the New Project Dialog to instead use this new function.

### DIFF
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -1041,16 +1041,15 @@ class NewProjectAction(UIWindow.Action):
                     return True
 
                 def verify_project_name(text: str) -> None:
-                    valid = text == Utility.simplify_filename(text)
-                    if valid:
-                        create_project_button.enabled = True
+                    is_valid, errors = Utility.verify_filename_is_legal(text)
+                    create_project_button.enabled = is_valid
+                    if errors is None:
                         create_project_button.tool_tip = None
                         project_name_status_label.text = None
                     else:
-                        create_project_button.enabled = False
-                        invalid_chars_str = _("Invalid Characters in Project Name")
-                        create_project_button.tool_tip = invalid_chars_str
-                        project_name_status_label.text = invalid_chars_str
+                        error_str = "\n".join(errors)
+                        create_project_button.tool_tip = error_str
+                        project_name_status_label.text = error_str
                         project_name_status_label.text_color = "red"
                 column = self.ui.create_column_widget()
 
--- a/nion/swift/model/Utility.py
+++ b/nion/swift/model/Utility.py
@@ -7,6 +7,7 @@ import contextlib
 import dataclasses
 import datetime
 import functools
+import gettext
 import logging
 import os
 import pathlib
@@ -24,6 +25,8 @@ import numpy
 
 # local libraries
 from nion.utils import DateTime
+
+_ = gettext.gettext
 
 
 # datetimes are _local_ datetimes and must use this specific ISO 8601 format. 2013-11-17T08:43:21.389391
@@ -448,6 +451,42 @@ ILLEGAL_FILENAME_CHARS_REGEX = r'[<>:/\\|?*"\0-\31]'  # Capture illegal characte
 ILLEGAL_FILENAME_CHARS_AND_POSITION_REGEX = f'^[. ]|{ILLEGAL_FILENAME_CHARS_REGEX}|[. ]$'
 ILLEGAL_FILENAMES = ['CON', 'PRN', 'AUX', 'NUL', 'COM1', 'COM2', 'COM3', 'COM4', 'COM5', 'COM6', 'COM7', 'COM8', 'COM9',
                      'LPT1', 'LPT2', 'LPT3', 'LPT4', 'LPT5', 'LPT6', 'LPT7', 'LPT8', 'LPT9', 'COM¹', 'COM²', 'COM³', 'LPT¹', 'LPT²', 'LPT³']
+
+
+def verify_filename_is_legal(filename: str, maximum_length: int = 128) -> tuple[bool, typing.Sequence[str] | None]:
+    """Check if a filename is allowed on all platforms.
+
+    Returns a tuple with a bool indicating if the path is valid or not, and a list of error messages or None when there are no errors.
+    Checks the filename is not in ILLEGAL_FILENAMES and no matches in the ILLEGAL_FILENAME_CHARS_REGEX.
+    Even when the platform can support the filename if it is illegal on any platforms it will not be allowed so there can be cross-platform file compatibility.
+    """
+    assert maximum_length > 5
+    errors = []
+    if filename.startswith("."):
+        errors.append(_("Leading periods cause files to be hidden on some platforms"))
+
+    if len(filename) > maximum_length:
+        errors.append(_("Exceeds the allowed length of {max} characters").format(max=maximum_length))
+
+    if filename == "":
+        errors.append(_("Cannot be empty"))
+    if len(filename) > 1:  # A filename can be " " or "." but cannot end with those characters, the "." will be caught by the leading periods check and so shouldn't be duplicated
+        if filename.endswith(" "):
+            errors.append(_("Cannot end with a whitespace"))
+        if filename.endswith("."):
+            errors.append(_("Cannot end with a period"))
+
+    if filename.upper() in ILLEGAL_FILENAMES:
+        errors.append(_("\"{filename}\" is illegal as it is reserved on some platforms").format(filename=filename))
+    matches = re.findall(ILLEGAL_FILENAME_CHARS_REGEX, filename)
+    matches = sorted(set(matches))
+    if matches:
+        if len(matches) == 1:
+            errors.append(_("Contains illegal character '{character}'").format(character=matches[0]))
+        else:
+            errors.append(_("Contains illegal characters {characters}").format(characters=matches))
+
+    return len(errors) == 0, errors or None
 
 
 def simplify_filename(filename: str, replacement_char: str = '_', maximum_length: int = 128) -> str:
--- a/nion/swift/model/Utility.py
+++ b/nion/swift/model/Utility.py
@@ -443,23 +443,25 @@ def sample_stack_all(count: int = 10, interval: float = 0.1) -> None:
     threading.Thread(target=do_sample).start()
 
 
+# See https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+ILLEGAL_FILENAME_CHARS_REGEX = r'[<>:/\\|?*"\0-\31]'  # Capture illegal characters <, >, :, /, \, |, ?, *, and " as well as characters with integer representation between 0 and 31
+ILLEGAL_FILENAME_CHARS_AND_POSITION_REGEX = f'^[. ]|{ILLEGAL_FILENAME_CHARS_REGEX}|[. ]$'
+ILLEGAL_FILENAMES = ['CON', 'PRN', 'AUX', 'NUL', 'COM1', 'COM2', 'COM3', 'COM4', 'COM5', 'COM6', 'COM7', 'COM8', 'COM9',
+                     'LPT1', 'LPT2', 'LPT3', 'LPT4', 'LPT5', 'LPT6', 'LPT7', 'LPT8', 'LPT9', 'COM¹', 'COM²', 'COM³', 'LPT¹', 'LPT²', 'LPT³']
+
+
 def simplify_filename(filename: str, replacement_char: str = '_', maximum_length: int = 128) -> str:
     # This function replaces any illegal characters in a file name. Not a full path.
     # macOS illegal characters = \0 /
     # linux illegal characters = \0 /
     # Windows illegal characters = < > : / \ | ? * " ASCII values 0-31 (non-printable chars)
     # Windows illegal filenames = CON PRN AUX NUL COM1 COM2 COM3 COM4 COM5 COM6 COM7 COM8 COM9
-    #                             LPT1 LPT2 LPT3 LPT4 LPT5 LPT6 LPT7 LPT8 LPT9
+    #                             LPT1 LPT2 LPT3 LPT4 LPT5 LPT6 LPT7 LPT8 LPT9 COM¹ COM² COM³ LPT¹ LPT² LPT³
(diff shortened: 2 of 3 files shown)


MattRoyle:
Here are some examples of the new error messages:

Screenshot 2026-04-29 154456

Screenshot 2026-04-29 154510

Screenshot 2026-04-29 154519

nion/swift/model/Utility.py
```diff
@@ -443,24 +446,75 @@ def do_sample() -> None:
...
+    The second element is a list of error messages if the filename is invalid or warnings if it is valid.
+    Checks the filename is not in ILLEGAL_FILENAMES and no matches in the ILLEGAL_FILENAME_CHARS_REGEX.
+    error_prefix is appended to the start of the returned error, i.e: "Filename" + " cannot end with whitespace"
+    If the directory and suffix are provided then it checks if the total path length would exceed the 260-character path length limit on some platforms.
+    See https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+    """
+    errors = []
+    warnings = []
+    if directory is not None and suffix is not None:
+        total_path_length = len(directory) + 1 + len(filename) + len(suffix)  # Add 1 for the path separator
```
KRLango: Does the suffix always include the "." e.g. is "txt" an acceptable suffix? If it is acceptable, this will be off by 1 since it does not count the "."

nion/swift/model/Utility.py
```diff
@@ -443,24 +446,75 @@ def do_sample() -> None:
...
 
 
+ILLEGAL_FILENAME_CHARS_AND_POSITION_REGEX = r'^[. ]|[<>:/\\|?*\"]|[\0-\31]|[. ]$'
+ILLEGAL_FILENAME_CHARS_REGEX = r'[<>:/\\|?*"\0-\31]'
+ILLEGAL_FILENAMES = ['CON', 'PRN', 'AUX', 'NUL', 'COM1', 'COM2', 'COM3', 'COM4', 'COM5', 'COM6', 'COM7', 'COM8', 'COM9',
+                     'LPT1', 'LPT2', 'LPT3', 'LPT4', 'LPT5', 'LPT6', 'LPT7', 'LPT8', 'LPT9', 'COM¹', 'COM²', 'COM³', 'LPT¹', 'LPT²', 'LPT³']
+
+
+def verify_filename_is_legal(filename: str, suffix: str | None, directory: str | None = None,
+                             maximum_length: int = 128, error_prefix: str = "Filename") -> tuple[bool, list[str] | None]:
```
KRLango: Is there a benefit to having the suffix separate to the file name rather than included as part of the filename?

KRLango: Why do we need an `error_prefix`? If this function is to verify a filename, surely we would always want that to be "Filename"?

nion/swift/model/Utility.py
```diff
@@ -443,24 +446,75 @@ def do_sample() -> None:
...
+
+    assert maximum_length > 5
+    if len(filename) > maximum_length:
+        errors.append(error_prefix + _(" exceeds the allowed length of") + f" {maximum_length} characters")
+
+    if len(filename) > 1 and filename.endswith(" "):
+        errors.append(error_prefix + _(" cannot end with a whitespace"))
+    if len(filename) > 1 and filename.endswith("."):
+        errors.append(error_prefix + _(" cannot end with a period"))
+    if filename == "" or filename == " ":
```
KRLango: Why is " " being treated the same as "", but not "  " (2 spaces)? I would only do the "" check since the single space could be covered by other checks (e.g. ends with " "). Also I would put the empty string check first with an early return as all other checks are meaningless in that case.

nion/swift/model/Utility.py
```diff
@@ -443,24 +446,75 @@ def do_sample() -> None:
...
+    if matches:
+        if len(matches) == 1:
+            errors.append(error_prefix + _(" contained illegal character") + f" '{matches[0]}'")
+        else:
+            errors.append(error_prefix + _(" contained illegal characters") + f" {matches}")
+
+    if filename.startswith(" "):
+        warnings.append(_("Warning: Leading whitespaces are removed on some platforms\n") + error_prefix + _(" will be replaced with ") + f"\"_{filename[1:]}\"")
+    elif filename.startswith("."):
+        warnings.append(_("Warning: Leading periods cause files to be hidden on some platforms\n") + error_prefix + _(" will be replaced with ") + f"\"_{filename[1:]}\"")
```
KRLango: These warnings make assumptions about what will happen in other parts of the software. We cannot guarantee this behaviour for all future uses.

nion/swift/model/Utility.py
```diff
@@ -443,24 +446,75 @@ def do_sample() -> None:
...
+
+    if len(errors) > 0:
+        return False, errors
+    elif len(warnings) > 0:
+        return True, warnings
+    return True, None
```
KRLango: Given this logic, would a dataclass that encapsulates this logic be a good idea?

nion/swift/model/Utility.py
```diff
@@ -443,24 +446,75 @@ def do_sample() -> None:
...
+    if directory is not None and suffix is not None:
+        total_path_length = len(directory) + 1 + len(filename) + len(suffix)  # Add 1 for the path separator
+        if total_path_length > 260:
+            errors.append(error_prefix + _(" exceeds the maximum path of 260 characters on some platforms"))
+
+    assert maximum_length > 5
+    if len(filename) > maximum_length:
+        errors.append(error_prefix + _(" exceeds the allowed length of") + f" {maximum_length} characters")
+
+    if len(filename) > 1 and filename.endswith(" "):
```
KRLango: Why only check for ending with " " if the length is greater than 1? The error is still valid.

nion/swift/model/Utility.py
```diff
@@ -443,24 +446,75 @@ def do_sample() -> None:
...
+        errors.append(error_prefix + _(" cannot end with a whitespace"))
+    if len(filename) > 1 and filename.endswith("."):
```
KRLango: Why only check for ending with "." if the length is greater than 1? The error is still valid. And a filename of "." will currently be considered valid by this code

nion/swift/test/Utility_test.py
```diff
@@ -73,6 +73,54 @@ def test_simplify_filename(self):
             finally:
                 os.remove(file_path)
 
+    def test_verify_filename(self):
```
KRLango: This should be split into separate tests. If you get a failure that is AssertFalse fails - which "sub-test" has failed?

I would have 5 tests (with subtests where appropriate):
Errors
Successful
Long file name
Multiple errors
Warnings

nion/swift/test/Utility_test.py
```diff
@@ -73,6 +73,54 @@ def test_simplify_filename(self):
...
+        test_filenames = [("", "Filename cannot be empty"),
+                          (" ", "Filename cannot be empty"),
+                          ("file.", "Filename cannot end with a period"),
+                          ("file ", "Filename cannot end with a whitespace"),
+                          ("COM¹", "Filename \"COM¹\" is illegal as it is reserved on some platforms"),
+                          (r"5/3/2024", "Filename contained illegal character '/'"),
+                          ("1234567890" * 13, "Filename exceeds the allowed length of 128 characters"),
+                          ("NUL", "Filename \"NUL\" is illegal as it is reserved on some platforms"),
+                          ("abc\n\rdef", "Filename contained illegal characters ['\\n', '\\r']"),
+                          (" file.", "Filename cannot end with a period")]  # The warning doesn't show when there are errors
```
KRLango: I would add "." filename since I think that will currently fail.

nion/swift/Application.py
```diff
@@ -998,6 +998,37 @@ def create_handler(self, component_id: str, container: typing.Any = None, item:
 PreferencesDialog.PreferencesManager().register_preference_pane(FeaturesPreferencePanel())
 
 
+def verify_project_name(text: str, base_directory: str, profile: Profile.Profile) -> tuple[bool, list[str] | None]:
```
KRLango: This could use the same dataclass suggested for Utility.py

KRLango: `text` is not a good variable name. Should be `file_name`.
Why `base_directory` not just `directory`?

nion/swift/Application.py
```diff
@@ -998,6 +998,37 @@ def create_handler(self, component_id: str, container: typing.Any = None, item:
...
+    is_valid, errors = Utility.verify_filename_is_legal(text, suffix, base_directory, error_prefix="Project Name")
+    if not is_valid:
+        return False, errors
+    project_name = f"{text}{suffix}"
+    new_project_path = pathlib.Path(base_directory, project_name)
+    new_data_path = new_project_path.parent / f"{text} Data"
+    project_path_exists = new_project_path.exists()
+    data_path_exists = new_data_path.is_dir()
+
+    if project_path_exists or data_path_exists:
```
KRLango: Why not separate these into 2 simpler if statements and just generate the error for the project existing if both exist since that would be the default anyway.

nion/swift/Application.py
```diff
@@ -998,6 +998,37 @@ def create_handler(self, component_id: str, container: typing.Any = None, item:
...
+        # Either the project file or data folder already exist so name cannot be used.
+        error_parts: list[str] = [_("Project File ") + f"\"{project_name}\""] if project_path_exists else []
+        if data_path_exists:
+            error_parts.append(_("Data Folder ") + f"\"{text} " + _("Data\""))
+        error_str = _(" and ").join(error_parts) + _(" already exists")
+        return False, [error_str]
+
+    if profile.get_project_reference_by_path(new_project_path) is not None:
+        error_str = _("Project Reference already exists, remove it via Choose Project before proceeding")
```
KRLango: I think this needs a comment - I assume it is because an old project with the same name has been deleted outside of Swift. Why does that need the user to fix it rather than being handled automatically?

nion/swift/Application.py
```diff
@@ -1031,7 +1062,8 @@ def handle_close() -> bool:
                 # this is invoked by the button. the dialog will be closed by returning True. see Dialog.add_button.
                 # it should not explicitly close the dialog.
                 def handle_new() -> bool:
-                    app.create_project_reference(pathlib.Path(self.directory), self.__project_name_field.text or "untitled")
+                    filename = Utility.simplify_filename(self.__project_name_field.text or "untitled")
```
KRLango: If the code is checking the filename for issues - why are we then simplifying it?

nion/swift/ExportDialog.py
```diff
@@ -66,8 +66,9 @@ def update_export_button(__: typing.Any) -> None:
             invalid_reasons = list[str]()
 
             prefix_str = self.prefix.value or str()
-            if prefix_str != Utility.simplify_filename(prefix_str):
-                invalid_reasons.append(_("Prefix contains invalid characters"))
+            illegal_chars = re.findall(Utility.ILLEGAL_FILENAME_CHARS_REGEX, prefix_str)
+            if len(illegal_chars) != 0:
+                invalid_reasons.append(_("Prefix contains invalid characters: ") + f"{illegal_chars}")
```
KRLango: There is a regression here -> the old version would have counted "COM" as an invalid prefix. This will not.

KRLango (changes requested):

nion/swift/model/Utility.py
MattRoyle: Since I removed the warnings, the addition of the data class is no longer necessary.

nion/swift/Application.py
MattRoyle: I would say that handling that automatically is out of the scope of this work. The function is to verify the project can be created which should be a non destructive process.

nion/swift/ExportDialog.py
MattRoyle: I've updated it to now use verify filename is legal if the prefix is the only option included in the filename. It will use the ILLEGAL_FILENAME_CHARS_REGEX if the prefix is only part of the overall filename meaning a prefix of "COM" would work when there are other options making the overall filename only start with COM.

nion/swift/Application.py
MattRoyle: I think directory is a bit vague, base_directory implies the directory that the project will be contained in while directory to me could be misinterpreted as the data directory.

nion/swift/model/Utility.py
MattRoyle: The filename cannot end with a period or a whitespace but it can start with one, although a file named just "." will now be prevented by the "starts with a period error".

nion/swift/test/Application_test.py
```diff
@@ -25,6 +48,67 @@ def test_changes_json_file(self) -> None:
...
+                data_item = DataItem.DataItem(numpy.zeros((6, 6)), large_format=True)
+                document_model.append_data_item(data_item)  # This will cause the data folder to be created
+                project_reference = profile.project_references[0]
+                reference_path = project_reference.path
+                base_directory = reference_path.parent
+                data_folder_path = base_directory / "Project Data"
+                self.assertTrue(data_folder_path.is_dir())
+                # Get the data path using the data item
+                handler_attributes = FileStorageSystem.make_storage_handler_attributes(data_item)
+                data_path = FileStorageSystem.FileProjectStorageSystem.get_base_path(handler_attributes)
```
cmeyer: `get_base_path` is private for a reason. If you need to use it in a test, a better technique is to add a new method `_get_base_path_for_testing` that is quasi-private but the name makes it unlikely to be used outside of tests.

cmeyer (changes requested):

cmeyer:
I give in... squashed and merged 632cb48e63dbfccfc6f3bc9a4321a3b5b6fd635a

PR #1850, opened 2026-04-27
Fix the export multiple directory validation to update correctly

### LINKED ISSUE #1838: Exporting to non-existent directory fails
The default directory used in the export dialog may have been removed.

From @Ion-e:

> Test plan:
> 
> - Opening dialog with previously selected directory deleted will show a warning dialog and prevent user from clicking export.
> - Deleting chosen directory after opening dialog produces relevant error.
>  - A export in an invalid format to a non-existing directory should show the latter error in the completion window.

MattRoyle:
Fixes #1838.

There were some issues missed by fcf5ae47644f91c2f89b649e36005102fe054ab2 & 2cde63c489a40de54dfab33034069abf4c281da8:
The path was not checked when the export button was clicked which would lead to a traceback if the folder was deleted after the directory was selected but before the Export was clicked. The export button did not update when the select path was updated meaning if you have an invalid path then you would need to select a different path to even if the path was created in the select directory dialog.

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -60,7 +60,10 @@ class ExportDialogViewModel:
         def validate_directory(directory: str | None) -> bool:
             return directory is not None and pathlib.Path(directory).is_dir()
 
-        self.is_directory_valid = Model.StreamValueModel(Stream.MapStream(directory_stream, validate_directory))
+        self.__is_directory_valid = Model.StreamValueModel(Stream.MapStream(directory_stream, validate_directory))
+
+        if not self.is_directory_valid:
+            self.directory.value = None  # Clear the initial directory if it was invalid
 
         def update_export_button(__: typing.Any) -> None:
             invalid_reasons = list[str]()
@@ -69,9 +72,10 @@ class ExportDialogViewModel:
             if prefix_str != Utility.simplify_filename(prefix_str):
                 invalid_reasons.append(_("Prefix contains invalid characters"))
 
-            is_directory_valid = self.is_directory_valid.value or False
+            is_directory_valid = self.__is_directory_valid.value or False
             if not is_directory_valid:
                 invalid_reasons.append(_("Directory does not exist"))
+                self.directory.value = None  # Clear the directory if it is invalid
 
             if invalid_reasons:
                 self.export_button_enabled.value = False
@@ -82,7 +86,7 @@ class ExportDialogViewModel:
                 self.export_button_tool_tip.value = None
                 self.directory_warning.value = str()
 
-        self.__is_directory_valid_action = Stream.ValueStreamAction(Stream.PropertyChangedEventStream[str](self.is_directory_valid, "value"), update_export_button)
+        self.__is_directory_valid_action = Stream.ValueStreamAction(Stream.PropertyChangedEventStream[str](self.__is_directory_valid, "value"), update_export_button)
         self.__prefix_action = Stream.ValueStreamAction(prefix_stream, update_export_button)
 
         update_export_button(None)
@@ -90,6 +94,11 @@ class ExportDialogViewModel:
     @property
     def directory_path_object(self) -> pathlib.Path:
         return pathlib.Path(self.directory.value or str())
+
+    @property
+    def is_directory_valid(self) -> bool:
+        self.__is_directory_valid.value = self.directory.value is not None and pathlib.Path(self.directory.value).is_dir()
+        return self.__is_directory_valid.value or False
 
     def build_filepath(self, displayed_title: str, date: datetime.datetime, dimensional_shape: DataAndMetadata.ShapeType | None, index: int) -> pathlib.Path:
         filename_components = list()
@@ -173,7 +182,7 @@ class ExportDialog(Declarative.Handler):
 
         # perform export, but save the last used writer
         def handle_export_clicked() -> bool:
-            if not self.viewmodel.is_directory_valid.value:
+            if not self.viewmodel.is_directory_valid:
                 return False
             selected_writer = self.viewmodel.writer.value
             writer_id = selected_writer.io_handler_id if selected_writer else "png-io-handler"
@@ -280,7 +289,7 @@ class ExportDialog(Declarative.Handler):
         writer = viewmodel.writer.value
         if writer:
             export_results = list()
-            is_directory_valid = viewmodel.is_directory_valid.value or False
+            is_directory_valid = viewmodel.is_directory_valid
             for index, display_item in enumerate(display_items):
                 if not is_directory_valid:
                     error_message = _("Directory does not exist")


nion/swift/ExportDialog.py
```diff
@@ -87,6 +87,10 @@ def update_export_button(__: typing.Any) -> None:
 
         update_export_button(None)
 
+    def validate_directory(self) -> bool:
```
cmeyer: I'm not sure what your intention is here... "validate" is a verb, so I would expect this function to validate the directory, if anything.

Why was the `is_directory_valid` model changed above?

nion/swift/ExportDialog.py
```diff
@@ -173,7 +177,8 @@ def __init__(self, ui: UserInterface.UserInterface, document_controller: Documen
 
         # perform export, but save the last used writer
         def handle_export_clicked() -> bool:
-            if not self.viewmodel.is_directory_valid.value:
+            if not self.viewmodel.validate_directory():
```
cmeyer: The recent model changes were intended so that all logic would be moved to the model. This is "logic" and should be in the model.

Did you originally intend `validate_directory` to validate the directory and update it?

nion/swift/ExportDialog.py
```diff
@@ -208,6 +213,7 @@ def choose_directory(self, widget: Declarative.UIWidget) -> None:
         if selected_directory:
             self.viewmodel.directory.value = selected_directory
             self.ui.set_persistent_string("export_directory", selected_directory)
+            self.viewmodel.is_directory_valid.value = self.viewmodel.validate_directory()
```
cmeyer: Model logic should be in the model, not here.

cmeyer (changes requested):

nion/swift/ExportDialog.py
MattRoyle: I'll change swap the name with the update_is_directory_valid. 
The issue is that the is_directory_valid wasn't being updated when the directory was being set without changing the actual value.

The user saves some data to C:\Data, they then archive that data which deletes the C:\Data on the PC, they go to export more images and it shows C:\Data as an invalid directory, so they press select directory and in the explorer dialog they create C:\Data and confirm that folder as the selection. Because the path did not change the state is not updated and so they cannot export despite the user making it valid.

My solution to the problem is probably not the most elegant and could probably be improved but what is needed is some mechanism to cause a callback function when the directory value is set, not just when it changes.

nion/swift/ExportDialog.py
MattRoyle: The reason behind this change is that only is_directory_valid value being checked when the export button was clicked meant the directory value could have become stale as the directory itself could be changed after it was selected and the is_directory_valid was updated. 

Using validate_directory (which in this version just checks the directory rather than refreshing the the is_directory_valid value), is probably not the best way to do it. It probably would make more sense to change validate_directory back to how you had it, where it updates the is_directory_valid, but move that out to be a member function of the ExportDialogViewModel. 

If the changes were to move the logic into model, why not move the handle_export_clicked into the model?

nion/swift/ExportDialog.py
MattRoyle: Would  conditional updating of the directory in the ExportDialog not be considered "logic" as well?
https://github.com/nion-software/nionswift/blob/2cde63c489a40de54dfab33034069abf4c281da8/nion/swift/ExportDialog.py#L205-L210

Without some kind of "on_value_set" callback for the self.viewmodel.directory.value = selected_directory the only way to catch the case where the [directory doesn't change values](https://github.com/nion-software/nionswift/pull/1850#discussion_r3162334864) but the directory itself did become valid is to refresh the state explicitly when the directory was selected

nion/swift/ExportDialog.py
cmeyer: One definition of the model could be "anything you would reasonably want to do without the actual UI for automation or testing". I think `export_clicked` should be renamed (`export`?) and go into the model. `handle_export_clicked` is inherently a UI handler, so should not be in the model; but any logic that can be consolidated to the model should be moved there. We're getting into edge cases, but generally if the UI does two or more lines of model manipulation, we should consider whether it belongs in the model.

nion/swift/ExportDialog.py
cmeyer: `validate_directory` still isn't a good name as it is a verb a is returning a condition, not performing any action. But...

~What about making `is_directory_valid_stream` a `MapStream` that takes the `directory_stream` and applied the condition in `validate_directory` to it, producing a `bool`? Then it's always up to date with the `directory` and doesn't require any extra logic?~

I see that was the original implementation.

cmeyer (changes requested):

nion/swift/ExportDialog.py
cmeyer: In the pre-PR code, if the user opens the dialog and the directory is invalid, the `is_directory_valid_stream` will be `False` from the beginning, as it is updated when creating the `MapStream`.

How about just putting a check at the beginning of `__init__` which sets the directory to `None` if it doesn't exist?

`directory = directory if directory and pathlib.Path(directory).is_dir() else None`?

MattRoyle: It would still not fire when the value is updated. The streams do not have the functionality to fire an event when the value was set but not changed. I have pushed a change where now the validate_directory will update the is_directory_valid value and return the result.

cmeyer: Detecting directory changes on disk is _way_ out of scope for this.

At most, we could check just before we perform the export, although I think this is also out of scope and will fail for other reasons anyway.

cmeyer: Isn't checking if the directory exists when we enter the dialog is entirely sufficient? I don't think we should do additional checks after that unless the user changes the directory.

nion/swift/ExportDialog.py
```diff
@@ -213,7 +211,7 @@ def choose_directory(self, widget: Declarative.UIWidget) -> None:
         if selected_directory:
             self.viewmodel.directory.value = selected_directory
             self.ui.set_persistent_string("export_directory", selected_directory)
-            self.viewmodel.is_directory_valid.value = self.viewmodel.validate_directory()
+        self.viewmodel.is_directory_valid.value = self.viewmodel.validate_directory()  # Update the is_directory_valid whenever the choose directory dialog is closed, even if the no directory was selected.
```
cmeyer: I continue to dislike this logic. As before, `validate_directory` is a verb and has side effects but is being assigned to a value in the model.

What about adding `validated_directory` property to the model? Then rewrite this whole section as

```
self.viewmodel.set_selected_directory(selected_directory)
self.ui.set_persistent_string("export_directory", self.viewmodel.validated_directory)
```

As client code, this makes sense and allows all of the logic to go into the viewmodel.

MattRoyle:
> Fixes #1838.
> 
> There were some issues missed by [fcf5ae4](https://github.com/nion-software/nionswift/commit/fcf5ae47644f91c2f89b649e36005102fe054ab2) & [2cde63c](https://github.com/nion-software/nionswift/commit/2cde63c489a40de54dfab33034069abf4c281da8): The path was not checked when the export button was clicked which would lead to a traceback if the folder was deleted after the directory was selected but before the Export was clicked. The export button did not update when the select path was updated meaning if you have an invalid path then you would need to select a different path to even if the path was created in the select directory dialog.

Revised changes: Set the value to None if the directory is invalid. Use a property getter for is_directory_valid instead of accessing the model is_directory_valid.value itself. The value can be refreshed when the export button is clicked preventing any tracebacks.

nion/swift/ExportDialog.py
```diff
@@ -69,9 +72,10 @@ def update_export_button(__: typing.Any) -> None:
             if prefix_str != Utility.simplify_filename(prefix_str):
                 invalid_reasons.append(_("Prefix contains invalid characters"))
 
-            is_directory_valid = self.is_directory_valid.value or False
+            is_directory_valid = self.__is_directory_valid.value or False
             if not is_directory_valid:
                 invalid_reasons.append(_("Directory does not exist"))
+                self.directory.value = None  # Clear the directory if it is invalid
```
MattRoyle, line 78: When the export button is updated it will clear the directory.value to None. Since the __is_directory_value model is set before the update method is called the value doesn't change when the directory.value is changed which prevents the __is_directory_valid_action from being recalled by the update_export_button.

cmeyer:
I'm merging, but we should consider a follow-up PR to make all `viewmodel` properties exposed from `viewmodel` directly instead of via `viewmodel.property.value`. This goes for `writer`, `directory`, `prefix`, and `include_xyz` style properties.

Ion-e:
Verified as implemented.

Thanks

PR #1859, opened 2026-05-12
Improve display pipeline threading

cmeyer:
- **Restructure image canvas item state tracking.**
- **Remove spurious change to last display data delta.**
- **Clean up some DisplayDataDelta typing (use tuple/sequence instead of list).**
- **Rename 'data_info' to more specific 'export_data_info'.**
- **Simplify access to DisplayDataChannel.DisplayValues.**
- **Introduce DisplayDataInfo and use it in place of DisplayValues where possible.**
- **Split non-display parts of line plot/graph into LinePlotDisplay.py.**
- **Split non-display parts of image display into ImageDisplay.py.**
- **Introduce DisplayInfo base class for image and line plot display info.**
- **Move DisplayInfo updating into base DisplayCanvasItem.**
- **Simplify display info handling.**
- **Merge DisplayDataDelta into DisplayInfo and eliminate DisplayDataDelta.**
- **Consolidate DisplayItem access to display_calibration_info and display_data_info into display_info_stream.**
- **Restructure the display_info_stream as a MapStream with 'compute_display_info' applied to inputs.**
- **Restructure display_info_stream to work on-demand. Eliminate cached_display_info.**
- **Split display info computation into data and non-data inputs.**
- **Move cursor text rendering and thread handling from DisplayItem to DisplayPanel.**
- **Rework thumbnails to use streams (display_info, bitmap).**
- **Move display data computation into a thread.** <-- most important commit
- **Update typing (mypy 1.21).**
- [1] **fixup! Rework thumbnails to use streams (display_info, bitmap).**
- [2] **fixup! Move display data computation into a thread.**
- [-] **Change workspace actions to use Dialog pose functions anchored to the workspace rect**
- [-] **Remove pose methods from Workspace and update Facade and ProjectPanel to use Dialog pose functions**
- [-] **Improve SVG export aspect ratio handling. Fixed for images. Choices/custom for line plots.**
- [3] **fixup! Move display data computation into a thread.**
- **Update test_select_multiple to be self contained.**
- **Fix canvas item thread ownership warning in line plot canvas item.**
- [4] **Introduce GraphicRenderer to eliminate races around Graphic objects during drawing.**
- [5] **fixup! Restructure the display_info_stream as a MapStream with 'compute_display_info' applied to inputs.**

The goal of this PR is to standardize access to display data, calibrations, display properties, display layers, graphics, and the graphic selection state so that lengthy computations occur on a thread and displays and other UI are not blocked by them. This is a major change which touches many parts of the codebase, but it is an important step towards improving the performance and responsiveness of the software when working with large datasets and complex displays.

A key outcome of this change is that display computations will not occur on the main thread. Furthermore, they will not hold up graphics updates on the display items. The user should be able to adjust graphics without stuttering even when display data computations are occurring on a thread. This is a significant improvement in the user experience when working with large datasets and complex displays.

This PR touches display canvas items, display items, computations, the inspector, the histogram, thumbnails, and cursor handling.

This PR rearranges the display canvas items (image, line plot) so that they receive updates via a DisplayInfo object which contains precomputed display data, calibrations, display properties, display layers, graphics, and the graphic selection. The display canvas items are simplified by being able to assume there is a minimal compute cost for accessing the display info.

To do this, the DisplayInfo is further separated into model portion (ImageDisplayInfo, LinePlotDisplayInfo) and canvas item portion (ImageDisplayCanvasItem, LinePlotDisplayCanvasItem). The model portion is responsible for computing the display data and related information required for the canvas item, while the canvas item portion is responsible for rendering the display based on the provided information. This separation allows for better organization of the code and makes it easier to manage the threading of the display data computations.

An important change is standardizing DisplayInfo to be general and not specific to either image or line plot display. This eliminates the need for the display item to know too much about the display info required for its display type. Each display type has its own specific DisplayInfo wrapper, but it is created from within the specific display canvas item (UI) rather than in the display item (model). Note: both ImageDisplayInfo and LinePlotDisplayInfo are left in the model namespace for now even though they are only used in the UI level.

DisplayInfo is only computed when there are listeners to it. When there is a display canvas item that needs display info, the display canvas item will receive updated DisplayInfo on a thread and then an update will be triggered on the display canvas item. The canvas item system will also update the canvas item on a thread. DisplayInfo can also update graphics and other non-data related information without triggering the display data computation, allowing for faster updates when only graphics or display properties change. Graphics can be updated while the display data is being computed on a thread, allowing for smoother user interactions.

Many of the steps in this PR are intermediate steps to the overall goal of standardizing the access to display data and display info so that it can be computed on a thread.

I plan to add the description of this PR above to developer docs once this PR is merged. There are probably other places that need a bit more documentation, but I wanted to start this review process now.

---

Fixes forthcoming for these issues revealed in tests:

1. A race condition where the initial thumbnail bitmap is None can be triggered by running `test_data_item_display_thumbnail_source_produces_data_item_mime_data` test resulting in failed tests.
2. AI-assisted review spotted an issue with the lock being held while setting the value in `ComputedValueStream.__run`, which can lead to deadlock if the value setter tries to acquire the same lock. The solution is to release the lock before setting the value.
3. A display pipeline deadlock occurs when creating display_info_stream. The problem is reassigning self.__display_info_stream releases the previous stream which can be finalized while holding the display_info_stream_lock. Finalization requires the lock too. The solution is to keep reference to the previous stream while holding the lock. Seen during `test_select_multiple` and other tests.
4. The graphics can be deleted during display item canvas drawing, revealed in `test_computation_inspector_panel_handles_computation_being_removed_implicitly` and other tests. The problem is how the graphics are passed to the displays. I think the solution is to pass a light graphic copy to the display, not the heavy weight close-able object. Still working on this.
5. Composite line plots don't reload properly. This was the hack that was removed. Restored the hack.


nion/swift/ImageCanvasItem.py
```diff
@@ -1435,17 +1408,17 @@ def _update_image_canvas_position(self, widget_delta: Geometry.FloatSize) -> Geo
...
         if coord:
-            widget_mapping = ImageCanvasItemMapping.make(self.__data_shape, self.__composite_canvas_item.canvas_bounds,
-                                                         list())
+            data_shape = self.__last_image_display_info.data_shape.as_tuple() if self.__last_image_display_info.data_shape else None
+            widget_mapping = ImageCanvasItemMapping.make(data_shape, self.__composite_canvas_item.canvas_bounds, list())
             if widget_mapping:
                 mapped = self.map_widget_to_image(coord)
-                if (mapped is not None) and (self.__data_shape is not None):
-                    norm_coord = tuple(mapped_coord / shape_dim for mapped_coord, shape_dim in zip(iter(mapped), iter(self.__data_shape)))
+                if (mapped is not None) and (data_shape is not None):
```
MattRoyle, line 1263: Could these parenthesizes be removed?

nion/swift/ImageCanvasItem.py
```diff
@@ -1723,25 +1705,26 @@ def set_two_to_one_mode(self) -> None:
     def get_current_zoom(self) -> float:
```
MattRoyle, line 1548: In this function the only allowed values of image_canvas_mode are "custom", "fit", "fill", "1:1" and "2:1" this differs from the behavior of the calculate_origin_and_size. Would these benefit from making this an enum, although it may be a lot of work to change.

nion/swift/ImageCanvasItem.py
```diff
@@ -1273,115 +1253,114 @@ def _scale_marker_canvas_item_for_test(self) -> ScaleMarkerCanvasItem:
...
+            # update the display values
+            display_values = image_display_info.display_values
+            if display_values is not None:
+                if display_values != self.__last_image_display_info.display_values:
+                    display_data = display_values.adjusted_data_and_metadata
+                    if display_data:
+                        if display_data.data_dtype == numpy.float32:
+                            display_range = display_values.transformed_display_range
+                            color_map_data = display_values.color_map_data
+                            color_map_rgba: typing.Optional[DrawingContext.RGBA32Type]
```
MattRoyle, line 1157: Could change Typing.Optional to be a | None instead.

nion/swift/ImageCanvasItem.py
```diff
@@ -1273,115 +1253,114 @@ def _scale_marker_canvas_item_for_test(self) -> ScaleMarkerCanvasItem:
...
+
+            self.__frame_rate_canvas_item.frame_tick(frame_info.frame_index)
+
+            self.scroll_area_canvas_item._data_shape = data_shape
+            self.__composite_canvas_item._data_shape = data_shape
+
+            scroll_area_canvas_size = self.scroll_area_canvas_item.canvas_size
+            if scroll_area_canvas_size is not None:
+                # only update layout if the size/origin will change. it is slow.
+                image_canvas_rect: typing.Optional[Geometry.IntRect]
```
MattRoyle, line 1136: I think because of the indentation changes this is marked as new code even though it already existed but typing.Optional should be Geometry.IntRect | None

nion/swift/ImageCanvasItem.py
```diff
@@ -436,28 +431,27 @@ def _get_composer(self, composer_cache: CanvasItem.ComposerCache) -> typing.Opti
...
+        return Geometry.IntRect(image_canvas_origin, data_shape)
     elif image_canvas_mode == "2:1":
-        image_canvas_size = Geometry.IntSize(round(data_shape[0] * 0.5), round(data_shape[1] * 0.5))
-        image_canvas_origin = Geometry.IntPoint(round(canvas_size.height * 0.5 - data_shape[0] * 0.25),
-                                                round(canvas_size.width * 0.5 - data_shape[1] * 0.25))
+        image_canvas_size = Geometry.IntSize(round(data_shape.height * 0.5), round(data_shape.width * 0.5))
+        image_canvas_origin = Geometry.IntPoint(round(canvas_size.height * 0.5 - data_shape.height * 0.25),
+                                                round(canvas_size.width * 0.5 - data_shape.width * 0.25))
         return Geometry.IntRect(image_canvas_origin, image_canvas_size)
     else:
```
MattRoyle, line 460: What is this else representing in the image_canvas_mode? Elsewhere it seems it can be set to "custom" so is this case supposed to work for only "custom" or also other values?

nion/swift/ImageCanvasItem.py
```diff
@@ -1255,133 +1089,116 @@ def __init__(self, ui_settings: UISettings.UISettings,
...
+                    image_canvas_rect = None
+                if image_canvas_rect != self.__composite_canvas_item.canvas_rect:
+                    # layout. this makes sure that the info overlay gets updated too.
+                    self.update_layout(self.canvas_origin, self.canvas_size)
+                    # trigger updates
+                    self.__composite_canvas_item.update()
+
+            # update the display values
+            display_data_info = image_display_info.display_data_info
+            if display_data_info:
```
MattRoyle, line 1149: The truthfulness of display_data_info and display_data could be ambiguous, would "is not None" be more clear here?

MattRoyle, line 1149: The nested if blocks could be simplified to something like this:
``` python
if not display_data_info:
    self.__bitmap_canvas_item.set_rgba_bitmap_data(None)
    self.__timestamp_canvas_item.timestamp_ns = 0
elif display_data_info != self.__last_image_display_info.display_data_info:
    derived_display_values = display_data_info.derived_display_values
    display_data = derived_display_values.adjusted_data_and_metadata
    if not display_data:
        self.__bitmap_canvas_item.set_rgba_bitmap_data(None)
    elif display_data.data_dtype == numpy.float32:
            display_range = derived_display_values.transformed_display_range
            color_map_data = derived_display_values.color_map_data
            color_map_rgba: DrawingContext.RGBA32Type | None = None
            if color_map_data is not None:
                color_map_rgba = numpy.empty(color_map_data.shape[:-1] + (4,), numpy.uint8)
                color_map_rgba[..., 0:3] = color_map_data
                color_map_rgba[..., 3] = 255
                color_map_rgba = color_map_rgba.view(numpy.uint32).reshape(color_map_rgba.shape[:-1])
            # the canvas item expects a ndarray. since the data is being accessed directly here, it could
            # be an array compatible that might be deleted (e.g. a hdf5 dataset). so convert it to a numpy
            # if not already.
            display_data_data = display_data.data
            if not isinstance(display_data_data, numpy.ndarray):
                display_data_data = numpy.array(display_data_data)
            self.__bitmap_canvas_item.set_data(display_data_data, display_range, color_map_rgba)
    else:
        data_rgba = derived_display_values.display_rgba
        self.__bitmap_canvas_item.set_rgba_bitmap_data(data_rgba)
    data_metadata = display_data_info.data_metadata
    metadata_d = data_metadata.metadata if data_metadata else dict()
    timestamp_ns = metadata_d.get("hardware_source", dict()).get("system_time_ns", time.perf_counter_ns()) if self.__display_latency_model.value else 0
    self.__timestamp_canvas_item.timestamp_ns = timestamp_ns
```

MattRoyle (commented):
Mostly pedantic review of the changes in the first commit ffe9257, I looked at the final version to make sure the comments were still relevant.

nion/swift/ImageCanvasItem.py
cmeyer: Since this is such a large PR, I've tried to keep most of the formatting the same as before.

nion/swift/ImageCanvasItem.py
cmeyer: This is out of scope of this PR.

nion/swift/ImageCanvasItem.py
cmeyer: Since this is such a large PR, I've tried to keep most of the formatting the same as before.

nion/swift/ImageCanvasItem.py
cmeyer: Out of scope.

nion/swift/ImageCanvasItem.py
cmeyer: Since this is such a large PR, I've tried to keep most of the formatting the same as before.

cmeyer: Since this is such a large PR, I've tried to keep most of the formatting the same as before.

nion/swift/ImageCanvasItem.py
cmeyer: Since this is such a large PR, I've tried to keep most of the formatting the same as before.

cmeyer:
Tested on AppDev. Slightly faster overall latency/frame-rate; significantly reduced latency/frame-rate variation. Tested using Ctrl+Shift+L on both Ronchigram and SuperScan displays running simultaneously, neither display with keyboard focus, and cursor not over either one.

cmeyer:
I updated the top comment with several test failures and deadlocks. I will be providing patches in this PR to fix these issues and will update the top comment with the additional fixup commits when available.

lisham2000 (changes requested):
Problem on project load draw of composite line plots. When closing and reopening software any composite line plots will fail to draw all but one line and the legend will read unknown for all plots. Repro case is just create create composite line plot and close then reopen software

Ion-e:
Bugs:
1. Creating a data item appears as "No data" data panel but it does have data when dragging to workspace. Please see screenshot. (Edit: this seems to happen due to 5D items not being populated with data)
image

2. (Sometimes) when generating data, the item created doesn't automatically populate the workspace:
image

3. Traceback when generating data: 
[Traceback (most recent call last).txt](https://github.com/user-attachments/files/27757437/Traceback.most.recent.call.last.txt)

4. Generate data freezes software when atttempting to create item as shown in screenshot, after a while it unfreezes:
image
5. After using this PR, and swapping to Master, I got this logs (unsure if they are backwards compatibility issues,, but most likely happened when clicking the "My Data" collection to rename it):
[Launch time 2026-05-14 143213.99316.txt](https://github.com/user-attachments/files/27761376/Launch.time.2026-05-14.143213.99316.txt)

KRLango:
@Ion-e In the first screenshot you are trying to generate a 256GB data item. Generate data does this entirely in RAM, and you don't have that much RAM. In the second, you are trying to generate a 4TB data item - depending on where you are doing that, you probably don't even have the hard-drive space to save it. I don't think either of these relate to this PR.

cmeyer:
@lisham2000 Thanks. Will fix the composite line plot issue.

cmeyer:
This has been autosquashed, rebased, and merged.

`git log 0d18991cc676d34fb1e4ffe5a7b72c65006439f3..97bfeaf3fae33967b12b7ac6fde274674872e9f2`

Please open new issues for any bugs.

Issue #1854, opened 2026-05-05
Processing > FFT > Gaussian Window Doesn't Work *as expected*


Ion-e:
From @bentps, could you please elaborate on the issue?

The Gaussian window doesn't work in certain cases

Ion-e:
Edit: The core behavioural issue is that in most cases the default sigma is not very useful and users must change it after the computation is run.

cmeyer:
This is the raw code from `ProcessingGaussianWindow` in `Processing.py`. The 1D path ignores `signma`. The 2D path doesn't use `scipy.signal.windows.gaussian`. So, as a minimum effort, we probably need a different approach to make the two consistent.

```
    def process(self, parameters: Symbolic.ComputationParameters) -> typing.Mapping[str, _ProcessingResult]:
        src_xdata = parameters.get_data_and_metadata("src")
        sigma = parameters.get_float_value("sigma", 1.0)
        if src_xdata and src_xdata.datum_dimension_count == 1:
            w = src_xdata.datum_dimension_shape[0]
            return {"target": src_xdata * scipy.signal.windows.gaussian(src_xdata.datum_dimension_shape[0], std=w / 2)}
        elif src_xdata and src_xdata.datum_dimension_count == 2:
            # uses circularly rotated approach of generating 2D filter from 1D
            h, w = src_xdata.datum_dimension_shape
            y, x = numpy.meshgrid(numpy.linspace(-h / 2, h / 2, h), numpy.linspace(-w / 2, w / 2, w), indexing='ij')
            s = 1 / (min(w, h) * sigma)
            r = numpy.sqrt(y * y + x * x) * s
            return {"target": src_xdata * numpy.exp(-0.5 * r * r)}
        return dict()
```

Ion-e:
LGTM.

cmeyer:
I reopened this. The old code ignored sigma in the 1d case. The revised code was not self consistent.

I think we want to specify the "sigma" parameter as a "window fraction" or something similar, somehow specifying what fraction of the image is windowed.

cmeyer:
Note: I have a preliminary PR which implements the window generation functions in `niondata`:
- https://github.com/nion-software/niondata/pull/78

Once we settle on how/if to handle "window fraction", I'll expand that PR with window _apply_ functions which take the 'window fraction' as the parameter and translate the 'window fraction' to the low level 'sigma' in the window generation functions.

Ion-e:
[...]

PR #1864, opened 2026-05-15
Fix Sigma default value for Gaussian Window computation

Ion-e:
From @bentps, it is more likely that the user will want a sigma closer to 0.3 values than 1 when running a Gaussian Window Computation, the resulting computation will also give a better visual cue that the computation was run.


Ion-e:
Fixes https://github.com/nion-software/nionswift/issues/1854

cmeyer:
Thanks for the quick fix!

PR #1872, opened 2026-05-21
Prevent export when no options are selected

### LINKED ISSUE #1871: Export traceback when all options are disabled
When all the options in the export multiple dialog are disabled the export button can still be clicked which causes a traceback due to the filename being an empty string.

MattRoyle:
Fixes #1871. Refactor the export dialog to use CombineLatestStream to call the update_export_button whenever there is a change to any of the options in order to make sure when all the options are deselected the export button becomes disabled.

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -46,7 +46,6 @@ class ExportDialogViewModel:
         self.include_date = Model.PropertyModel(date)
         self.include_dimensions = Model.PropertyModel(dimensions)
         self.include_sequence = Model.PropertyModel(sequence)
-        self.include_prefix = Model.PropertyModel(prefix is not None)
         self.prefix = Model.PropertyModel[str](prefix)
         self.directory = Model.PropertyModel(directory)
         self.directory_warning = Model.PropertyModel(str())
@@ -65,12 +64,17 @@ class ExportDialogViewModel:
         if not self.is_directory_valid:
             self.directory.value = None  # Clear the initial directory if it was invalid
 
-        def update_export_button(__: typing.Any) -> None:
+        def update_export_button(*__: typing.Any) -> None:
             invalid_reasons = list[str]()
 
             prefix_str = self.prefix.value or str()
-            if prefix_str != Utility.simplify_filename(prefix_str):
-                invalid_reasons.append(_("Prefix contains invalid characters"))
+            if prefix_str != "":
+                if prefix_str != Utility.simplify_filename(prefix_str):
+                    invalid_reasons.append(_("Prefix contains invalid characters"))
+            elif not (self.include_title.value or self.include_date.value or
+                      self.include_dimensions.value or self.include_sequence.value):
+                # There must be at least one option enabled otherwise the filenames will be empty
+                invalid_reasons.append(_("Filename requires at least one option to be selected"))
 
             is_directory_valid = self.__is_directory_valid.value or False
             if not is_directory_valid:
@@ -86,9 +90,18 @@ class ExportDialogViewModel:
                 self.export_button_tool_tip.value = None
                 self.directory_warning.value = str()
 
-        self.__is_directory_valid_action = Stream.ValueStreamAction(Stream.PropertyChangedEventStream[str](self.__is_directory_valid, "value"), update_export_button)
-        self.__prefix_action = Stream.ValueStreamAction(prefix_stream, update_export_button)
+        input_streams: typing.Sequence[Stream.PropertyChangedEventStream[typing.Any]] = [
+            # Update the button when one of the options changes
+            prefix_stream,
+            Stream.PropertyChangedEventStream(self.include_title, "value"),
+            Stream.PropertyChangedEventStream(self.include_date, "value"),
+            Stream.PropertyChangedEventStream(self.include_dimensions, "value"),
+            Stream.PropertyChangedEventStream(self.include_sequence, "value"),
+            # Also update the button when the directory changes validity
+            Stream.PropertyChangedEventStream(self.__is_directory_valid, "value")
+        ]
 
+        self.__export_button_update_stream: Stream.CombineLatestStream[typing.Any, typing.Any] = Stream.CombineLatestStream(input_streams, update_export_button)
         update_export_button(None)
 
     @property
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -63,18 +63,20 @@ class ExportDialogViewModel:
         if not self.is_directory_valid:
             self.directory.value = None  # Clear the initial directory if it was invalid
 
-        def update_export_button(*__: typing.Any) -> None:
+        def update_export_button(prefix_value: str | None, include_title: bool | None,
+                                 include_date: bool | None, include_dimensions: bool | None,
+                                 include_sequence: bool | None, is_directory_valid: bool | None) -> None:
+            """Update the export button enabled state and tooltip based on the current options selected and directory validity."""
             invalid_reasons = list[str]()
 
-            prefix_str = self.prefix.value or str()
+            prefix_str = prefix_value or str()
             if prefix_str != Utility.simplify_filename(prefix_str):
                 invalid_reasons.append(_("Prefix contains invalid characters"))
 
-            if prefix_str == "" and not (self.include_title.value or self.include_date.value or self.include_dimensions.value or self.include_sequence.value):
+            if not (prefix_str or include_title or include_date or include_dimensions or include_sequence):
                 # There must be at least one option enabled otherwise the filenames will be empty
                 invalid_reasons.append(_("Filename requires at least one option to be selected"))
 
-            is_directory_valid = self.__is_directory_valid.value or False
             if not is_directory_valid:
                 invalid_reasons.append(_("Directory does not exist"))
                 self.directory.value = None  # Clear the directory if it is invalid
@@ -88,7 +90,7 @@ class ExportDialogViewModel:
                 self.export_button_tool_tip.value = None
                 self.directory_warning.value = str()
 
-        export_filename_option_streams: typing.Sequence[Stream.PropertyChangedEventStream[typing.Any]] = [
+        export_filename_option_streams: typing.Sequence[Stream.PropertyChangedEventStream[str | bool]] = [
             # Update the button when one of the options changes
             Stream.PropertyChangedEventStream(self.prefix, "value"),
             Stream.PropertyChangedEventStream(self.include_title, "value"),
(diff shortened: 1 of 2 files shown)


nion/swift/test/ExportDialog_test.py
```diff
@@ -47,6 +47,29 @@ def test_model(self):
         self.assertFalse(model.export_button_enabled.value)
         self.assertIsNotNone(model.export_button_tool_tip.value)
 
+    def test_model_prevents_no_options_selection(self) -> None:
+        model = ExportDialog.ExportDialogViewModel(
+            title=False,
+            date=False,
+            dimensions=False,
+            sequence=False,
+            writer=None,
```
KRLango, line 59: `writer=None` feels like something that could also disable the export button. Should this be set to isolate that as a potential problem within this test?

nion/swift/test/ExportDialog_test.py
```diff
@@ -47,6 +47,29 @@ def test_model(self):
         self.assertFalse(model.export_button_enabled.value)
         self.assertIsNotNone(model.export_button_tool_tip.value)
 
+    def test_model_prevents_no_options_selection(self) -> None:
```
KRLango: This name doesn't represent what this test does. Currently it tests that no options selected disables the export button, and that when you select an option it enables it, and that adding a prefix enables it. Currently this name would suggest that you are testing the model prevents you from having no selections which is not true.

I would split this into 3 distinct tests.

nion/swift/ExportDialog.py
```diff
@@ -65,12 +64,17 @@ def validate_directory(directory: str | None) -> bool:
...
             self.directory.value = None  # Clear the initial directory if it was invalid
 
-        def update_export_button(__: typing.Any) -> None:
+        def update_export_button(*__: typing.Any) -> None:
             invalid_reasons = list[str]()
 
             prefix_str = self.prefix.value or str()
-            if prefix_str != Utility.simplify_filename(prefix_str):
-                invalid_reasons.append(_("Prefix contains invalid characters"))
+            if prefix_str != "":
```
KRLango: I would change this logic to something a little simpler since the nested if's make it hard to follow at the moment:

```python
if prefix_str != Utility.simplify_filename(prefix_str):
    invalid_reasons.append(_("Prefix contains invalid characters"))

if prefix_str  == "" and not (self.include_title.value or self.include_date.value or
                              self.include_dimensions.value or self.include_sequence.value):
    invalid_reasons.append(_("Filename requires at least one option to be selected"))
```

nion/swift/ExportDialog.py
```diff
@@ -86,9 +90,18 @@ def update_export_button(__: typing.Any) -> None:
                 self.export_button_tool_tip.value = None
                 self.directory_warning.value = str()
 
-        self.__is_directory_valid_action = Stream.ValueStreamAction(Stream.PropertyChangedEventStream[str](self.__is_directory_valid, "value"), update_export_button)
-        self.__prefix_action = Stream.ValueStreamAction(prefix_stream, update_export_button)
+        input_streams: typing.Sequence[Stream.PropertyChangedEventStream[typing.Any]] = [
```
KRLango: `input_streams` does not seem like a very descriptive name, something like `export_name_option_streams` would better describe this variable.

nion/swift/ExportDialog.py
```diff
@@ -86,9 +90,18 @@ def update_export_button(__: typing.Any) -> None:
...
+            # Update the button when one of the options changes
+            prefix_stream,
```
KRLango: This looks to be the only use of `prefix_stream` - I would declare it in this list creation to match the format of the others below

KRLango (changes requested):

nion/swift/test/ExportDialog_test.py
MattRoyle: Since the tests only testing the UI and aren't actually writing any data the writer is not required. It shouldn't be possible to not have the writer set as it is accessed though a combo box.

nion/swift/test/ExportDialog_test.py
MattRoyle: I have split it into two tests: test_model_updates_button_status_when_prefix_changes, test_model_updates_button_status_when_options_change.
The tests start disabled then update the prefix and an option respectively to check that the button becomes enabled then tests that when they become invalid again the button status updates accordingly. 
The no options selected is tested as a consequence of this, but maybe that should be it's own test as well.

nion/swift/test/ExportDialog_test.py
```diff
@@ -33,19 +33,64 @@ def test_model(self):
             prefix=None,
             directory=None)
         # initial case - no directory
-        self.assertIsNotNone(model.directory_warning.value)
+        self.assertTrue(model.directory_warning.value)  # The warning and tool tip can be '' or None if the button is in an invalid state
```
cmeyer: Why is this test being changed? And if there is a reason to change it, can you explain why `assertIsNotNone` is insufficient?

nion/swift/ExportDialog.py
```diff
@@ -65,13 +63,17 @@ def validate_directory(directory: str | None) -> bool:
         if not self.is_directory_valid:
             self.directory.value = None  # Clear the initial directory if it was invalid
 
-        def update_export_button(__: typing.Any) -> None:
+        def update_export_button(*__: typing.Any) -> None:
```
cmeyer: `*__` does not look like valid syntax.

nion/swift/ExportDialog.py
```diff
@@ -65,13 +63,17 @@ def validate_directory(directory: str | None) -> bool:
...
             invalid_reasons = list[str]()
 
             prefix_str = self.prefix.value or str()
             if prefix_str != Utility.simplify_filename(prefix_str):
                 invalid_reasons.append(_("Prefix contains invalid characters"))
 
+            if prefix_str == "" and not (self.include_title.value or self.include_date.value or self.include_dimensions.value or self.include_sequence.value):
```
cmeyer: How about `if not prefix_str and ...`? At the very minimum, either use `str()` (line 69) or `""` (line 73) but be consistent.

nion/swift/ExportDialog.py
cmeyer: The combine function should take inputs (the stream value for each input stream) and stand on its own, i.e. not access `self`. Using `typing.Any` hides the problem.

nion/swift/ExportDialog.py
```diff
@@ -86,9 +88,18 @@ def update_export_button(__: typing.Any) -> None:
...
+            Stream.PropertyChangedEventStream(self.prefix, "value"),
+            Stream.PropertyChangedEventStream(self.include_title, "value"),
+            Stream.PropertyChangedEventStream(self.include_date, "value"),
+            Stream.PropertyChangedEventStream(self.include_dimensions, "value"),
+            Stream.PropertyChangedEventStream(self.include_sequence, "value"),
+            # Also update the button when the directory changes validity
+            Stream.PropertyChangedEventStream(self.__is_directory_valid, "value")
+        ]
 
+        self.__export_button_update_stream: Stream.CombineLatestStream[typing.Any, typing.Any] = Stream.CombineLatestStream(export_filename_option_streams , update_export_button)
```
cmeyer: Why does `typing.Any` need to be used here? It obscures the intention. At the very minimum, it should not use `typing.Any` for the _output_  of the stream. In fact, the output should match the output type of `update_export_button`. See the note above about `update_export_button`. Streams should not be used for their side effects - if you want a side effect, you can use `Stream.ValueStreamAction` explicitly, but I don't think that's a good approach here. Define and utilize the return value from `update_export_button` instead.

cmeyer (changes requested):

nion/swift/test/ExportDialog_test.py
MattRoyle:
```
# The warning and tool tip can be '' or None if the button is in an invalid state
```

The reason is that some of the assertIsNotNone were 'dead code', as when the model.directory_warning is cleared the value is set to an empty string rather than None, so the assertion would always pass even if the value was an empty string. 
https://github.com/nion-software/nionswift/blob/9753485bd58c3e8e40cfcd0fafe16dd0e0c08d77/nion/swift/ExportDialog.py#L52-L55
https://github.com/nion-software/nionswift/blob/9753485bd58c3e8e40cfcd0fafe16dd0e0c08d77/nion/swift/ExportDialog.py#L80-L87
Changing to assertTrue means that the warning tests will be testing the truthfulness, which covers both cases of being initialized to None and being set to an empty string.

MattRoyle: Having read the comment on line 36 again it should say "is in a *valid* state" rather than "is in an *invalid* state"
I have updated it to better explain why the truthfulness check is required:
```
# The warning and tool tip can be '' or None when the button is in a valid state 
# So we check that the truthfulness when there is expected to be an invalid state to ensure there is a warning and tool-tip present
```
https://github.com/nion-software/nionswift/pull/1872/changes#diff-559292718b3c190b4ace3c41909026e346178b0ec50b87f8232b642a1c57b424R36-R38

Ion-e:
LGTM

nion/swift/test/ExportDialog_test.py
cmeyer: Further note: these are the only two places in the code that have this problem (using a source wide search with regexp `isnotnone.+tool_tip`).

nion/swift/ExportDialog.py
```diff
@@ -65,14 +63,20 @@ def validate_directory(directory: str | None) -> bool:
         if not self.is_directory_valid:
             self.directory.value = None  # Clear the initial directory if it was invalid
 
-        def update_export_button(__: typing.Any) -> None:
+        def update_export_button(prefix_value: str | None, include_title: bool | None,
+                                 include_date: bool | None, include_dimensions: bool | None,
+                                 include_sequence: bool | None, is_directory_valid: bool | None) -> None:
```
cmeyer: Instead of returning `None` from the update_export_button combine function, we should keep our stream combination functions side-effect-free (meaning they don't reference `self` or captured local variables).

To handle this cleanly, let's have the combine function output a `tuple` or a `dataclass` containing the `export_button_enabled`, `export_button_tool_tip`, and `directory_warning` fields. From there, we can attach an action to that combined stream's output to handle the updates. Since actions are meant for side effects, that's the perfect place to update self with the new values.

cmeyer (changes requested):

cmeyer:
squashed, merged, minor typing edit, updated commit message. feaf331aed187d12876959a598cd51dee65eb855

PR #1874, opened 2026-05-21
Renaming projects on disk

### LINKED ISSUE #504: Add ability to rename projects on disk
Project chooser should give option to remove and rename projects. Or some other UI.

These are basic functions expected by users.

Rename might also be something to be applied to the currently open project.

MattRoyle:
Fixes #504. Adds the ability to rename the currently open project via the File menu, as well as rename other projects through the Choose Project menu.
When current open project is renamed it is closed, then renamed then opened. 
Similar to the opening of projects, if the user renames a project while computations, acquisitions or other live updates, there will be problems caused.

### DIFF
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -968,7 +968,7 @@ class Application(UIApplication.BaseApplication):
         new_project_reference: Profile.ProjectReference | None = None
         try:
             new_project_reference, rename_result = self.profile.rename_project(project_reference, name)
-            error_message = rename_result.error_message
+            error_message = "\n".join(rename_result.error_messages)
         except Exception as e:
             error_message = _("Exception during renaming project")
             logging.exception(f"{error_message}:\n{e}")
@@ -1269,7 +1269,8 @@ class NameProjectViewModel:
         self.profile = profile
 
     @classmethod
-    def verify_project_name(cls, project_name: str, base_directory: str, profile: Profile.Profile) -> ProjectNameVerificationResult:
+    def verify_project_name(cls, project_name: str, base_directory: str, profile: Profile.Profile,
+                            project_name_available_fn: typing.Callable[[str, str], FileStorageSystem.ProjectNameResult] | None = None) -> ProjectNameVerificationResult:
         """Verify that a project name is a valid filename and doesn't already exist.
 
         Returns a tuple with the first element being whether the project name is valid and the second being a sequence of error messages.
@@ -1278,26 +1279,25 @@ class NameProjectViewModel:
             return ProjectNameVerificationResult(False, ["Project name cannot be empty"])
         if project_name != Utility.simplify_filename(project_name):
             return ProjectNameVerificationResult(False, ["Project Name contains illegal characters"])
-        new_project_path = pathlib.Path(base_directory, project_name).with_suffix(".nsproj")
-        errors = []
-        new_data_path = new_project_path.parent / f"{project_name} Data"
-        new_project_path_exists = new_project_path.exists()
-        data_path_exists = new_data_path.is_dir()
-        if new_project_path_exists:
-            errors.append(_("Project Name") + f" \"{new_project_path.name}\" " + _("already exists"))
-        if data_path_exists:
-            errors.append(_("Data Folder") + f" \"{new_data_path.stem}\" " + _("already exists"))
 
-        # Check for an orphaned project reference exists, it would need to be removed manually
-        project_reference = profile.get_project_reference_by_path(new_project_path)
-        if not new_project_path_exists and not data_path_exists and project_reference is not None:
-            errors.append(_("Project Reference") + f" \"{project_reference.title}\" " + _("already exists, remove it via Choose Project before proceeding"))
+        errors: list[str] = []
+        if project_name_available_fn is None:
+            return ProjectNameVerificationResult(True, errors)
+
+        name_available_result = project_name_available_fn(project_name, base_directory)
+        errors.extend(name_available_result.error_messages)
+        if not errors and name_available_result.project_path:
+            # Check for an orphaned project reference exists, it would need to be removed manually
+            project_reference = profile.get_project_reference_by_path(name_available_result.project_path)
+            if project_reference is not None:
+                errors.append(_("Project Reference") + f" \"{project_reference.title}\" " + _("already exists, remove it via Choose Project before proceeding"))
 
         is_valid = len(errors) == 0
         return ProjectNameVerificationResult(is_valid, errors)
 
     def update_project_status_label(self, text: str) -> None:
-        project_name_verification_result = self.verify_project_name(text, self.directory.value or str(), self.profile)
+        project_name_verification_result = self.verify_project_name(text, self.directory.value or str(), self.profile,
+                                                                    FileStorageSystem.FileProjectStorageSystem.check_project_name_is_available)
         self.accept_button_enabled.value = project_name_verification_result.is_valid
         if not project_name_verification_result.error_messages:  # The name is valid and doesn't already exist.
             self.project_name_status_label.value = str()
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -1325,7 +1325,7 @@ class NameProjectDialog(Declarative.Handler):
         self.ui = ui
         self.__application = application
         self.viewmodel = NameProjectViewModel(project_name, directory, self.__application.profile)
-
+        self._project_name_line_edit: UserInterface.LineEditWidget | None = None
         # build the UI
         u = Declarative.DeclarativeUI()
 
@@ -1364,7 +1364,7 @@ class NameProjectDialog(Declarative.Handler):
 
         project_name_row = u.create_row(
             u.create_spacing(26),
-            u.create_line_edit(text="@binding(viewmodel.filename.value)", width=400, on_text_edited="handle_project_name_changed"),
+            u.create_line_edit(name="_project_name_line_edit", text="@binding(viewmodel.filename.value)", width=400, on_text_edited="handle_project_name_changed"),
             u.create_stretch(),
             u.create_spacing(13)
         )
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -1280,7 +1280,7 @@ class NameProjectViewModel:
             -> FileStorageSystem.ProjectNameResult:
         """Verify that a project name is a valid filename and doesn't already exist.
 
-        Returns a ProjectNameVerificationResult with is_valid, and a sequence of error messages.
+        Returns a ProjectNameResult with a sequence of error messages if there are any.
         """
         if project_name == "":
             return FileStorageSystem.ProjectNameResult(["Project name cannot be empty"], None)
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -747,7 +747,9 @@ class Application(UIApplication.BaseApplication):
                             action = RenameProjectAction(project_reference)
                             document_controller = self.__application.document_controllers[0]
                             action.invoke(UIWindow.ActionContext(self.__application, document_controller, None))
+
                         menu.add_menu_item(_("Rename Project"), functools.partial(rename_selected_project, project_reference_item.project_reference))
+
                         menu.popup(gx, gy)
                     return True
 
@@ -950,33 +952,35 @@ class Application(UIApplication.BaseApplication):
 
         If the project reference is the current project, it will be closed and reopened after renaming to update the title.
         """
-        self._enter_prevent_close_state()
-        old_title = project_reference.title
-        renaming_current_project = self.profile.last_project_reference == project_reference.uuid
-        document_controller = self.document_controllers[0]
-        assert document_controller is not None
-        if renaming_current_project:
-            document_controller.request_close()
-        error_message: str | None = None
-        new_project_reference: Profile.ProjectReference | None = None
-        try:
-            new_project_reference, rename_result = self.profile.rename_project(project_reference, name)
-            error_message = "\n".join(rename_result.error_messages)
-        except Exception as e:
-            error_message = _("Exception during renaming project")
-            logging.exception(f"{error_message}:\n{e}")
+        with self.prevent_close():
+            old_title = project_reference.title
+            renaming_current_project = self.profile.last_project_reference == project_reference.uuid
+            document_controller = self.document_controllers[0]
+            assert document_controller is not None
+            if renaming_current_project:
+                document_controller.request_close()
+            error_message: str | None = None
+            new_project_reference: Profile.ProjectReference | None = None
+            try:
+                new_project_reference, rename_result = self.profile.rename_project(project_reference, name)
+                error_message = "\n".join(rename_result.error_messages)
+            except Exception as e:
+                # Catch and log any exceptions during the rename process
+                error_message = _("Exception during renaming project")
+                logging.exception(e)
+                new_project_reference = project_reference
+            finally:
+                # Reopen the project only when renaming the current project and when the project reference is returned
+                if renaming_current_project and new_project_reference is not None:
+                    self.switch_project_reference(new_project_reference)
+                    assert new_project_reference.project is not None
+                    new_project_reference.project.title = name
+                    self.profile.last_project_reference = new_project_reference.uuid
 
-        if renaming_current_project and new_project_reference is not None:  # Reopen the project only when renaming the current project and when the project reference is returned
-            self.switch_project_reference(new_project_reference)
-            assert new_project_reference.project is not None
-            new_project_reference.project.title = name
-            self.profile.last_project_reference = new_project_reference.uuid
-
-        if error_message:
-            self.__show_project_error_dialog(_("Error Renaming Project"), error_message)
-        else:
-            logging.info(f"Renamed Project \"{old_title}\" to \"{name}\"")
-        self._exit_prevent_close_state()
+                if error_message:
+                    self.__show_project_error_dialog(_("Error Renaming Project"), error_message)
+                else:
+                    logging.info(f"Renamed Project \"{old_title}\" to \"{name}\"")
 
 
 def get_root_dir() -> str:
@@ -1335,8 +1339,6 @@ class NameProjectDialog(Declarative.Handler):
         self.accept_fn = accept_fn
         # build the UI
         u = Declarative.DeclarativeUI()
-
-
 
         title = dialog_name
         directory_header = u.create_row(
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -742,6 +742,19 @@ class Application(UIApplication.BaseApplication):
                             profile.remove_project_reference(project_reference_item.project_reference)
 
                         menu.add_menu_item(_(f"Remove Project from List"), functools.partial(remove_project, index))
+
+                        def rename_selected_project(project_reference: Profile.ProjectReference) -> None:
+                            document_controller = self.__application.document_controllers[0]
+                            if document_controller is not None:
+                                directory = project_reference.path.parent.as_posix()
+
+                                def handle_rename_clicked(new_name: str, _: str) -> bool:
+                                    self.__application.rename_project(project_reference, new_name)
+                                    return False  # Don't close the dialog since it will have already been closed when reopening the project
+
+                                NameProjectDialog(self.__application.ui, document_controller, self.__application, directory, project_reference.title, handle_rename_clicked, "Rename Project", False)
+
+                        menu.add_menu_item(_("Rename Project"), functools.partial(rename_selected_project, project_reference_item.project_reference))
                         menu.popup(gx, gy)
                     return True
 
@@ -1209,6 +1222,35 @@ class ChooseProjectAction(UIWindow.Action):
         return UIWindow.ActionResult(UIWindow.ActionStatus.FINISHED)
 
 
+class RenameCurrentProjectAction(UIWindow.Action):
+    action_id = "project.rename_current_project"
+    action_name = _("Rename Project")
+
+    def execute(self, context: UIWindow.ActionContext) -> UIWindow.ActionResult:
+        raise NotImplementedError()
+
+    def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
+        context = typing.cast(DocumentController.DocumentController.ActionContext, context_)
+        document_controller = typing.cast(DocumentController.DocumentController, context.window)
+        application = typing.cast(Application, context.application)
+
+        current_project_reference = None
+        last_project_uuid = application.profile.last_project_reference
+        if last_project_uuid is not None:
+            current_project_reference = application.profile.get_project_reference(last_project_uuid)
+
+        if current_project_reference is None:
+            return UIWindow.ActionResult(UIWindow.ActionStatus.CANCELLED)
+
+        def handle_rename_clicked(new_name: str, _: str) -> bool:
+            assert current_project_reference is not None
+            application.rename_project(current_project_reference, new_name)
+            return False  # Don't close the dialog since it will have already been closed when reopening the project
+
+        directory = current_project_reference.path.parent.as_posix()
+        NameProjectDialog(application.ui, document_controller, application, directory, current_project_reference.title, handle_rename_clicked, "Rename Project", False)
+        return UIWindow.ActionResult(UIWindow.ActionStatus.FINISHED)
+
 
 @dataclasses.dataclass(frozen=True)
 class ProjectNameVerificationResult:
@@ -1372,3 +1414,4 @@ class NameProjectDialog(Declarative.Handler):
 UIWindow.register_action(NewProjectAction())
 UIWindow.register_action(OpenProjectAction())
 UIWindow.register_action(ChooseProjectAction())
+UIWindow.register_action(RenameCurrentProjectAction())
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -745,14 +745,13 @@ class Application(UIApplication.BaseApplication):
 
                         def rename_selected_project(project_reference: Profile.ProjectReference) -> None:
                             document_controller = self.__application.document_controllers[0]
-                            if document_controller is not None:
-                                directory = project_reference.path.parent.as_posix()
+                            directory = project_reference.path.parent.as_posix()
 
-                                def handle_rename_clicked(new_name: str, _: str) -> bool:
-                                    self.__application.rename_project(project_reference, new_name)
-                                    return False  # Don't close the dialog since it will have already been closed when reopening the project
+                            def handle_rename_clicked(new_name: str, _: str) -> bool:
+                                self.__application.rename_project(project_reference, new_name)
+                                return False  # Don't close the dialog since it will have already been closed when reopening the project
 
-                                NameProjectDialog(self.__application.ui, document_controller, self.__application, directory, project_reference.title, handle_rename_clicked, "Rename Project", False)
+                            NameProjectDialog(self.__application.ui, document_controller, self.__application, directory, project_reference.title, handle_rename_clicked, "Rename Project", False)
 
                         menu.add_menu_item(_("Rename Project"), functools.partial(rename_selected_project, project_reference_item.project_reference))
                         menu.popup(gx, gy)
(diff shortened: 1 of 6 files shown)


nion/swift/test/Project_test.py
```diff
@@ -3,19 +3,24 @@
...
 from nion.swift import Facade
 from nion.swift.model import Connection
 from nion.swift.model import DataItem
 from nion.swift.model import DisplayItem
 from nion.swift.model import Profile
 from nion.swift.test import TestContext
+from nion.swift.test import Storage_test
 from nion.ui import TestUI
+import pathlib
+from nion.ui import UserInterface
```
cmeyer: Can you run your aggregate change (commits squashed) through the auditor, which would catch these out-of-order import issues.

nion/swift/test/Project_test.py
```diff
@@ -82,3 +100,182 @@ def test_adding_same_project_raises_error_during_append(self):
                 project_reference.close()
 
     # do not import same project (by uuid) twice
+
+    def setup_project_from_profile(self, profile_context: TestContext.MemoryProfileContext) -> tuple[Profile.Profile, Application.Application, DocumentController.DocumentController]:
```
cmeyer: Having complicated tests is an anti-pattern. These tests are far too complicated.

cmeyer (changes requested):
Here are a few suggestions to streamline this PR:

Architecture: I recommend keeping the command simple rather than undoable to avoid unnecessary error risks. Ideally, consolidate the rename function into a single Application method for tests to invoke.

Testing: It would be cleaner to move these tests directly into Storage_test to avoid cross-imports, and skip the mock setup to keep things straightforward. What was the main use case you had in mind for the mocks?

UI & Validation: Let's update RenameProjectDialog to use declarative widgets. Also, let's omit the filename validation from this PR entirely—we can implement the simplified version from the discussion post later if desired.

cmeyer:
Additional design comments:

- Renaming a project will be dependent on the project storage; let's add a new method on `ProjectStorageSystem` to rename a project to a different name (passing the name only, not the directory). This allows a future HDF5-backed filesystem to rename itself properly.
- We will also need a method in `nion.swift.model` (likely in `Profile`) that calls the `ProjectStorageSystem` method and manages the `ProjectReference` cleanly. It should probably take the target `ProjectReference` as a parameter so we can rename already-closed projects in the future.
- There should be a UI-level function in the `nion.swift` namespace that closes the current project if it is the one being renamed, calls the project-reference-aware function in `Profile` to rename it, and then re-opens the project.
- Include a declarative dialog that calls this UI-level function. Separate into model/UI so model might be tested.
- We should ensure there are tests for each of these function levels.

nion/swift/model/FileStorageSystem.py
```diff
@@ -43,6 +46,18 @@
 _g_large_format_size = 16 * 1024 * 1024
 
 
+@dataclasses.dataclass(frozen=True)
+class ProjectNameResult:
```
KRLango, line 50: This would benefit from a `success` property - not passed in but derived from the error messages list. Something like:

```python
@property
def success(self)->bool:
    return not self.error_messages
```

nion/swift/Application.py
```diff
@@ -1176,6 +1222,202 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+            project_reference = profile.get_project_reference_by_path(name_available_result.project_path)
+            if project_reference is not None:
+                errors.append(_("Project Reference") + f" \"{project_reference.title}\" " + _("already exists, remove it via Choose Project before proceeding"))
+
+        is_valid = len(errors) == 0
+        return ProjectNameVerificationResult(is_valid, errors)
+
+    def update_project_status_label(self, text: str) -> None:
+        project_name_verification_result = self.verify_project_name(text, self.directory.value or str(), self.profile,
+                                                                    FileStorageSystem.FileProjectStorageSystem.check_project_name_is_available)
```
KRLango: This locks the dialog into using the FileProjectStorageSystem - are we sure that we want to do that?

nion/swift/model/FileStorageSystem.py
```diff
@@ -1115,6 +1150,65 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
...
+        old_title = old_project_path.stem
+        new_project_path = old_project_path.with_name(f"{name}.nsproj")
+        error_messages = []
+        project_path: pathlib.Path | None = None
+        try:
+            project_path = old_project_path.rename(new_project_path)
+        except Exception as e:
+            logging.exception(_("Failed to rename project. Renaming Project File gave an exception:\n") + str(e))
+            error_messages.append(_("Exception while renaming the Project File"))
+            project_path = None
```
KRLango: An early return on exception here would avoid an extra level of nesting in the next part of this function.

nion/swift/model/FileStorageSystem.py
```diff
@@ -1115,6 +1150,65 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
...
+
+        if not error_messages:  # Only rename the data folder if the project file was renamed
+            try:
+                if old_data_path is not None:
+                    old_data_path.rename(old_data_path.parent / f"{name} Data")
+            except Exception as e:
+                error_messages.append(_("Exception while renaming the Data Folder"))
+                logging.exception(_("Failed to rename project. Renaming Data Folder gave an exception:\n") + str(e))
+                try:  # Try to undo the rename of the project file when the data folder failed to rename
+                    if new_project_path.exists():
```
KRLango: Why would the new_project_path not exist - we just made it earlier in this function?

nion/swift/model/FileStorageSystem.py
```diff
@@ -1115,6 +1150,65 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
...
+                        new_project_path.rename(new_project_path.with_name(f"{old_title}.nsproj"))
```
KRLango: `new_project_path.rename(old_project_path)` should be the same and is simpler to read.

nion/swift/model/FileStorageSystem.py
```diff
@@ -1115,6 +1150,65 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
...
+                        project_path = None
+                except Exception as e2:
+                    error_messages.append(_("WARNING: Project File and Data Folder names have diverged"))
+                    logging.exception(_("Failed to undo Data Folder rename. Renaming project file gave exception:\n") + str(e2))
+                    project_path = self.__project_path.with_name(f"{name}.nsproj")  # The undo failed meaning the project path remains as the new one
```
KRLango: is this not just `new_project_path`?

nion/swift/model/FileStorageSystem.py
```diff
@@ -1115,6 +1150,65 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
...
+            project_path = old_project_path.rename(new_project_path)
+        except Exception as e:
+            logging.exception(_("Failed to rename project. Renaming Project File gave an exception:\n") + str(e))
+            error_messages.append(_("Exception while renaming the Project File"))
+            project_path = None
+
+        if not error_messages:  # Only rename the data folder if the project file was renamed
+            try:
+                if old_data_path is not None:
+                    old_data_path.rename(old_data_path.parent / f"{name} Data")
```
KRLango: I would define `new_data_path` at the start of the function - that way these parts of the logic would be easier to follow.

nion/swift/model/FileStorageSystem.py
```diff
@@ -607,6 +622,26 @@ def _migrate_library_properties(self, library_properties: PersistentDictType, re
     @abc.abstractmethod
     def _read_library_properties(self, migration_stage: ProjectStorageSystemMigrationStage) -> PersistentDictType: ...
 
+    @classmethod
+    def check_project_name_is_available(cls, name: str, directory: str) -> ProjectNameResult:
```
KRLango: Why are we splitting the directory and the filename here? Surely a path would be the best option and will include both.

nion/swift/Application.py
```diff
@@ -1176,6 +1222,202 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+
+    The invoke method must be used to call the rename function, calling execute will raise a NotImplementedError.
+    """
+    action_id = "project.rename_current_project"
+    action_name = _("Rename Project")
+
+    def execute(self, context: UIWindow.ActionContext) -> UIWindow.ActionResult:
+        raise NotImplementedError()
+
+    def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
```
KRLango, line 1245: why `context_`?

nion/swift/Application.py
```diff
@@ -1176,6 +1222,202 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+            application.rename_project(current_project_reference, new_name)
+            return False  # Don't close the dialog since it will have already been closed when reopening the project
+
+        directory = current_project_reference.path.parent.as_posix()
+        NameProjectDialog(application.ui, document_controller, application, directory, current_project_reference.title, handle_rename_clicked, "Rename Project", False)
+        return UIWindow.ActionResult(UIWindow.ActionStatus.FINISHED)
+
+
+@dataclasses.dataclass(frozen=True)
+class ProjectNameVerificationResult:
```
KRLango: Should `is_valid` be an input here, or is it a derived property?

nion/swift/Application.py
```diff
@@ -1176,6 +1222,202 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+    def verify_project_name(cls, project_name: str, base_directory: str, profile: Profile.Profile,
+                            project_name_available_fn: typing.Callable[[str, str], FileStorageSystem.ProjectNameResult] | None = None) \
+            -> ProjectNameVerificationResult:
+        """Verify that a project name is a valid filename and doesn't already exist.
+
+        Returns a ProjectNameVerificationResult with is_valid, and a sequence of error messages.
+        """
+        if project_name == "":
+            return ProjectNameVerificationResult(False, ["Project name cannot be empty"])
+        if project_name != Utility.simplify_filename(project_name):
```
KRLango: Is this check something that belongs here? Would it be better as part of the `project_name_available` functions that understand how the project is being saved?

MattRoyle:
> Additional design comments:
> 
>     * Renaming a project will be dependent on the project storage; let's add a new method on `ProjectStorageSystem` to rename a project to a different name (passing the name only, not the directory). This allows a future HDF5-backed filesystem to rename itself properly.
> 
>     * We will also need a method in `nion.swift.model` (likely in `Profile`) that calls the `ProjectStorageSystem` method and manages the `ProjectReference` cleanly. It should probably take the target `ProjectReference` as a parameter so we can rename already-closed projects in the future.
> 
>     * There should be a UI-level function in the `nion.swift` namespace that closes the current project if it is the one being renamed, calls the project-reference-aware function in `Profile` to rename it, and then re-opens the project.
> 
>     * Include a declarative dialog that calls this UI-level function. Separate into model/UI so model might be tested.
> 
>     * We should ensure there are tests for each of these function levels.

I have implemented the suggested changes, and moved the tests across. Now there is a declarative Dialog in the Application.py that calls a method in the application to manage closing and reopening the project if the project is open calling the profile rename method. The profile method handles removing the existing project_reference, calling the project_storage_system rename method then adding the new project reference. The project storage system level handles the actual renaming the files on the disk, which is specific to the project storage system, right now only the FileProjectStorageSystem implements the rename method renaming a ".nsproj" project file and a data folder if one exists. 

The dialog takes a method to check the project name is available which is added to the ProjectStorageSystem with an implementation in the FileProjectStorageSystem. The reason the dialog takes a method is to keep the project_storage_system level outside of the dialog/application.

There are four specific tests in the Storage_test for the FileProjectStorageSystem which start with "test_file_project_storage_". There are four tests of the application function to rename the project in Storage_test as they use the create_temp_profile_context, these begin with "test_rename_current_project_ ". 
There is a test for the NameProjectViewmodel in project_test test_project_name_viewmodel_is_invalid_with_existing_reference. There is another test for the NameProjectDialog which tests the connection between the viewmodel and the dialog handler, although I am not sure if that is necessary.

KRLango (commented):
There are a few dataclasses here that are of the form:
```python
@dataclasses.dataclass
class SomethingResult:
    is_valid: bool
    errors: typing.Sequence[str]
    result
```

If we are going to use this pattern we should formalise it similar to how TryValue works and have just a single definition with a generic type for the result.

nion/swift/Application.py
MattRoyle: I think it made sense to have it at this level as the simplify_filename is also at this level. Maybe both should be moved to the project_name_available or something.

nion/swift/Application.py
MattRoyle: I removed this dataclass to change it to use the ProjectNameResult instead which has success as a derived property

nion/swift/model/FileStorageSystem.py
MattRoyle: Because then the caller would need to know how to construct the path which is the responsibility of the project_storage_system. If it assumes that the project file will just be the directory / project name then this function could just use the stem of the path and use add a suffix, but that is an assumption that I think shouldn't be made.

nion/swift/Application.py
MattRoyle: I copied the methods from ChooseProjectAction which uses context_, I'll leave it for now

MattRoyle:
> There are a few dataclasses here that are of the form:
> 
> ```python
> @dataclasses.dataclass
> class SomethingResult:
>     is_valid: bool
>     errors: typing.Sequence[str]
>     result
> ```
> 
> If we are going to use this pattern we should formalise it similar to how TryValue works and have just a single definition with a generic type for the result.

I think that is outside of the scope here. The simple dataclass NameProjectResult that has a project path and errror messages suits the current implementation. But it could be readdressed to have some generic Result dataclass.

Ion-e (approved):
LGTM

cmeyer:
Overall this code looks ok, but a few things upon testing:
1. rename dialog escape key doesn't work as expected
2. if I type into the field and hit 'rename' button without hitting return to validate the text field, it will not rename to the right name. you can look at how other dialogs handle this. only tested on macOS.
3. RenameCurrentProjectAction.execute should implement the guts of the command without the dialog. RenameCurrentProjectAction.invoke should present the dialog then potentially call 'execute'. `handle_rename_clicked` should essentially be what the `execute` method does; and that could also possibly go into the `NameProjectViewModel` instead.
4. search for `from nion.swift.model.FileStorageSystem import FileProjectStorageSystem` and change usage to fully qualified name
5. search for `from nion.swift.model import DataItem, FileStorageSystem` and separate onto their own lines.
6. shouldn't test_name_project_dialog_button_validity_updates be testing the _model_ not the dialog (which is UI and we don't really test UI stuff...)
7. can `rename_selected_project` use the `RenameCurrentProjectAction`?
8. instead of passing `check_project_name_is_available` can you just pass the `project_storage_system` to the dialog? (see next item too)
9. can you separate the model and dialog so that the model can be tested independently of the dialog?
10. the uses of `unittest.mock` are code smells. why can't the tests create real files? this goes towards my earlier comment that `Application.NameProjectViewModel` should not appear in the tests - it's a dialog and can't be tested. We can test the model instead, right?
11. stopping here...

MattRoyle:
>     11. stopping here...

1. I have implemented the closing on pressing the escape key and accepting on pressing return. This only works when the line edit is focused, which is the same with how the New Project Dialog version works.
2. On windows as soon as you type anything the accept button refreshes, I've added a guard to validate when the rename button is clicked before it renames which should prevent this from happening.
3. I've moved the logic around so now it is split across invoke, which creates the dialog and on success calls the execute which calls the rename_project in the application.
4. refactored
5. refactored
6. I have removed the test_name_project_dialog_button_validity_updates UI test
7. I refactored the action to take an optional project_reference and use it in the invoke when provided or use the current project when not provided.
8. It now passes type[project_storage_system] | project_storage_system, this way if the dialog is used for the new project then it can pass a ProjectStorageSystem class rather than an instance so the class function can be called.
9. The model and dialog are already separate, I based the structure off the ExportDialog and ExportDialogViewModel where the dialog is handled by the former (NameProjectDialog), and the model logic is handled by the latter (NameProjectViewModel)
10. For the test in Project_test.py: The reason it can't test real files is that the Model being tested here NameProjectViewModel is meant to be independent of the underlying ProjectStorageSystem so instead of using actual files the mocks are used facilitate the specific part being tested, checking the project reference doesn't already exist. #
For the two uses in Storage_test.py this was done to simplify the tests as otherwise it would require making a temp profile context and a project.

nion/swift/Application.py
```diff
@@ -938,6 +945,39 @@ def execute_next_command(
...
+        if renaming_current_project:
+            document_controller.request_close()
+        error_message: str | None = None
+        new_project_reference: Profile.ProjectReference | None = None
+        try:
+            new_project_reference, rename_result = self.profile.rename_project(project_reference, name)
+            error_message = "\n".join(rename_result.error_messages)
+        except Exception as e:
+            error_message = _("Exception during renaming project")
+            logging.exception(f"{error_message}:\n{e}")
```
cmeyer: Is this intended to fall through on exception? The code here is not clear. If it's intentional, can you add a comment. If not, it's a bug. Overall, if intentional, it's confusing logic.

Related: can you use `with self.prevent_close():` to handle the 'close state' so it is guaranteed to exit cleanly?

nion/swift/Application.py
```diff
@@ -1175,6 +1215,233 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+        Returns a ProjectNameResult with a sequence of error messages if there are any.
+        """
+        if project_name == "":
+            return FileStorageSystem.ProjectNameResult(["Project name cannot be empty"], None)
+        is_valid, error_messages = Utility.verify_filename_is_legal(project_name)
+        if not is_valid and error_messages:
+            return FileStorageSystem.ProjectNameResult(error_messages, None)
+
+        errors: list[str] = []
+        name_available_result = project_storage_system.check_project_name_is_available(project_name, base_directory)
```
cmeyer: Is this meant to be a virtual method call to `check_project_name_is_available`? As written it's confusing. If it's meant to be static, can you make it a `@staticmethod` so it can't be called as a regular method; otherwise make it a regular method. Either way, `@classmethod` is wrong here — it gives access to `cls`, which isn't used.

nion/swift/model/FileStorageSystem.py
```diff
@@ -607,6 +626,26 @@ def _migrate_library_properties(self, library_properties: PersistentDictType, re
     @abc.abstractmethod
     def _read_library_properties(self, migration_stage: ProjectStorageSystemMigrationStage) -> PersistentDictType: ...
 
+    @classmethod
```
cmeyer: This shouldn't be a @classmethod since it doesn't use `cls`. Depending on intent, this is probably either a @staticmethod or a regular instance method — could you clarify which and adjust accordingly?

nion/swift/model/FileStorageSystem.py
```diff
@@ -1115,6 +1154,59 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
...
+                if new_project_path.exists():
+                    self.__project_path = new_project_path.rename(old_project_path)
+            except Exception as e2:
+                error_messages.append(_("WARNING: Project File and Data Folder names have diverged"))
+                logging.exception(_("Failed to undo Data Folder rename. Renaming project file gave exception:\n") + str(e2))
+
+        return ProjectNameResult(error_messages, self.__project_path)
+
+    @classmethod
+    def check_project_name_is_available(cls, name: str, directory: str) -> ProjectNameResult:
```
cmeyer: This shouldn't be a @classmethod since it doesn't use cls. Depending on intent, this is probably either a @staticmethod or a regular instance method — could you clarify which and adjust accordingly?

nion/swift/model/FileStorageSystem.py
```diff
@@ -43,6 +46,22 @@
 _g_large_format_size = 16 * 1024 * 1024
 
 
+@dataclasses.dataclass(frozen=True)
+class ProjectNameResult:
```
cmeyer, line 50: This looks like it's violating single responsibility — it's acting as a result for both a "rename" operation and a "check" (status) at once. Can you split into two classes/functions?

nion/swift/model/FileStorageSystem.py
```diff
@@ -1115,6 +1154,59 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
                         raise
         return storage_handlers
 
+    def rename_project(self, name: str) -> ProjectNameResult:
```
cmeyer, line 1187: Does `project_data_folders` entry in the JSON get updated here? If not, where does it get updated? Can you document how `project_data_folders` gets updated (or ensure it's updated if not)? And that should get a test by itself, too.

nion/swift/test/Project_test.py
```diff
@@ -82,3 +81,23 @@ def test_adding_same_project_raises_error_during_append(self):
                 project_reference.close()
 
     # do not import same project (by uuid) twice
+
+    def test_project_name_viewmodel_is_invalid_with_existing_reference(self) -> None:
```
cmeyer, line 85: > For the two uses in Storage_test.py this was done to simplify the tests as otherwise it would require making a temp profile context and a project.

This is critical functionality. Can you get rid of the mocks and create real files with the tests instead of using mocks?

nion/swift/test/Storage_test.py
```diff
@@ -4645,6 +4646,189 @@ def disabled_test_data_item_leaks_no_memory(self):
...
+                self.assertEqual(old_workspace_layout, document_controller.workspace_controller._workspace_layout)
+            finally:
+                app._set_profile_for_test(None)
+                app.exit()
+                app.deinitialize()
+                if document_controller.count > 0:
+                    document_controller.request_close()
+
+    def test_file_project_storage_system_check_project_name_is_unavailable_with_existing_project(self) -> None:
+        with unittest.mock.patch.object(pathlib.Path, 'exists', lambda _self: True):
```
cmeyer: Why is a `mock` needed here? Can you just create real files?

nion/swift/test/Storage_test.py
```diff
@@ -4645,6 +4646,189 @@ def disabled_test_data_item_leaks_no_memory(self):
...
+            current_working_directory = pathlib.Path.cwd()
+            check_name_result = FileStorageSystem.FileProjectStorageSystem.check_project_name_is_available("ExistingProject", str(current_working_directory))
+            self.assertEqual(check_name_result.project_path, (current_working_directory / "ExistingProject").with_suffix(".nsproj"))
+            self.assertEqual(check_name_result.error_messages, ["Project Name \"ExistingProject.nsproj\" already exists"])
+
+    def test_file_project_storage_system_check_project_name_is_unavailable_with_existing_data_folder(self) -> None:
+        with unittest.mock.patch.object(pathlib.Path, 'is_dir', lambda _self: True):
```
cmeyer: Why is a `mock` needed here? Can you just create real files?

nion/swift/Application.py
```diff
@@ -741,6 +742,12 @@ def remove_project(index: int) -> None:
                             profile.remove_project_reference(project_reference_item.project_reference)
 
                         menu.add_menu_item(_(f"Remove Project from List"), functools.partial(remove_project, index))
+
+                        def rename_selected_project(project_reference: Profile.ProjectReference) -> None:
+                            action = RenameProjectAction(project_reference)
+                            document_controller = self.__application.document_controllers[0]
+                            action.invoke(UIWindow.ActionContext(self.__application, document_controller, None))
+                        menu.add_menu_item(_("Rename Project"), functools.partial(rename_selected_project, project_reference_item.project_reference))
```
cmeyer, line 751: Can you add line breaks so this is similar to the previous version? i.e. line break after the function; line break after the `add_menu_item`?

nion/swift/Application.py
```diff
@@ -1175,6 +1215,233 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+        if self.project_reference is None:  # If the project reference is not set then use the current project reference
+            last_project_uuid = application.profile.last_project_reference
+            if last_project_uuid is not None:
+                self.project_reference = application.profile.get_project_reference(last_project_uuid)
+
+        if self.project_reference is None:
+            return UIWindow.ActionResult(UIWindow.ActionStatus.CANCELLED)
+
+        if self.project_reference.project is None:
+            project_storage_system = self.project_reference.make_storage(application.profile.profile_context)
```
cmeyer, line 1259: Shouldn't `make_storage` be in the `execute` portion of the code? It has to handle unwinding its operation if something fails afterwards, too.

nion/swift/Application.py
```diff
@@ -1175,6 +1215,233 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+        self.ui = ui
+        self.__application = application
+        self.viewmodel = NameProjectViewModel(project_name, directory, self.__application.profile, project_storage_system)
+        self._project_name_line_edit: UserInterface.LineEditWidget | None = None
+        self.accept_fn = accept_fn
+        # build the UI
+        u = Declarative.DeclarativeUI()
+
+
+
```
cmeyer: This is a lot of line breaks.

cmeyer (changes requested):
This is getting closer. Rough summary of the review:

- A few confusing bits in the code (see inline comments)
- I still think we shouldn't use mock anywhere in these tests
- project_data_folders in the JSON needs to be updated somewhere

nion/swift/Application.py
MattRoyle: The reason it is here is that the NameProjectDialog needs to take in a type of ProjectStorageSystem so it can use the check_project_name_is_available method. If the project_reference is not loaded then the .project does not exist and there is currently no way to determine the type that the project_reference will create without actually using the make_storage. Since load_properties is not being called on make_storage I assumed that there should be no side effects of creating a project storage system, however if there are then we may need to create a way to get the type of project storage system that a project reference will make so it can be passed. This is also happening in the profile rename_project as the same is true of when the rename is needed on a project reference that is not currently loaded with the project_storage_system.rename_project being used. 
https://github.com/nion-software/nionswift/blob/9f3957e0ed76cd4ecb7f6f00007e4e17cfff6263/nion/swift/model/Profile.py#L648-L652

nion/swift/model/FileStorageSystem.py
MattRoyle: I overlooked updating that, I thought something would trigger the write properties but upon testing that is not the case. Any suggestions on how/where the triggering of writing should be?

nion/swift/Application.py
cmeyer: I'm not evaluating this carefully, but since this code is more-or-less duplicated in `Profile.rename_project` as you point out, I'm guessing there is a cleaner way of handling this so that this is only done in that method. The solution probably involves passing `profile_context` and `project` around to `Application.rename_project` and subsequently `Profile.rename_project`, but, again, I didn't evaluate this in detail.

MattRoyle: So the profile is already at the level that in profile.rename_project. The NameProjectDialog does not need a intance of the project storage system type only the type so it can use the static method check_if_project_name_is_available. Maybe it makes sense for the project refernence to have a property that returns the type of Project storage system it will uses, or some mapping between the two. So then the NameProjectDialog can just take the type and call the check.
The profile.rename_project requires the project storage system so it can call down to the rename project of said storage system. If the project_reference's project is None, that means the project is not currently loaded so there is no associated project storage system so one needs to be created using make_storage even if I add a type getter to the project reference.

nion/swift/model/FileStorageSystem.py
MattRoyle: Both check and rename need to return the project path and a list of error messages hence why they return the ProjectNameResult. I can still split them up but the code would be duplicated.

nion/swift/test/Storage_test.py
```diff
@@ -12,6 +12,7 @@
 import threading
 import typing
 import unittest
+import unittest.mock
```
MattRoyle, line 15: This import is unused and should be removed

MattRoyle:
Closing in favor of splitting the UI #1926 and the renaming #1927.

PR #1875, opened 2026-05-27
Add initial data model and persistent object model design docs.

cmeyer:
I created a `design` directory (similar to PR #1484). The documents in design describe implementation details and are intended to be read by both humans and used in AI-assisted workflows. They are intended to apply to new code; existing code may not adhere completely to the documents.

### DIFF
--- a/design/data_model.md
+++ b/design/data_model.md
@@ -1,0 +1,81 @@
+# Nion Data Model
+
+## Dataset
+
+A **Dataset** is one complete unit of data: array, descriptor, dimensions, and associated metadata. This section describes the Dataset model in two parts: the current implementation state and the planned state.
+
+### Current Implementation
+
+This section uses dataset and axis-set terminology where possible, although the current implementation calls dataset a `DataAndMetadata` or `xdata`.
+
+A Dataset is organized into three ordered sections: an optional sequence axis of rank 1, an optional collection axis set of rank 1 or 2, and a datum axis set of rank 0, 1, 2.
+
+A Dataset holds values of type scalar, complex, RGB, or RGBA.
+
+Each axis has a single calibration associated with it. The calibration may be an identity calibration.
+
+### Planned Implementation
+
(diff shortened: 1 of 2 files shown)


design/data_model.md
```diff
@@ -0,0 +1,81 @@
+# Nion Data Model
+
+## Dataset
+
+A **Dataset** is one complete unit of data: array, descriptor, dimensions, and associated metadata. This section describes the Dataset model in two parts: the current implementation state and the planned state.
```
MattRoyle, line 5: The use of Dataset here is a bit confusing, in this first paragraph it describes what is the currently encapsulated in the DataAndMetadata class, line 9 then confirms this. In the code-base the term dataset is mostly used to when talking about hdf5 datasets which is the arraylike object of the raw data itself without metadata, descriptor, dimensions etc. This conflicting terminology sticks out, and may be confusing for AI. I haven't found other places where the term dataset is used to describe DataAndMetadata.

design/data_model.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Current Implementation
+
+This section uses dataset and axis-set terminology where possible, although the current implementation calls dataset a `DataAndMetadata` or `xdata`.
```
MattRoyle, line 9: This seems to imply that the intention is to change DataAndMetadata to be called Dataset or Dataset Model? If that is not the intention, I would say an AI agent may easily misinterpret that as being the intention. If a naming schema is needed we should base it off existing naming or standard naming which is usually that datasets are the raw numpy / hdf5 data, while Data or DataObject/Container (both bad names and definitely worse than the current DataAndMetadata) refers to the package of the dataset and auxiliary infomation

design/data_model.md
cmeyer: `Dataset` here is used in mostly the same sense as hdf5.

design/data_model.md
cmeyer: This may end up to be confusing and we may have to clarify instructions if this document is used for that.

PR #1876, opened 2026-06-08
Drag display panel split moves the content

### LINKED ISSUE #1797: Dragging display panel to split should remove old display panel content
1. create new workspace
2. drag image into workspace
3. drag title to form split
4. only one panel should be populated
5. observe how both panels are populated with the same item

Requested during software meeting.

MattRoyle:
Fixes #1797. 
Dragging a display panel over a split region now moves the display panels content to the newly created split rather than duplicating the content in both. 
The first commit improves the readability of the code in charge of the dropping handle_drop, which was initially modified although now the changes happen outside that function.


Tiomat85 (approved):
Liking the clearer variable names, as for the actual fix it looks a little weird but if it works and that is how it is supposed to work..

Ion-e:
LGTM

cmeyer:
Squashed and merged a9f12f10eba43906f3d4dff653c071d3be3b6744

PR #1880, opened 2026-06-17
Delay computation writes when inputs are write delayed.

### LINKED ISSUE #1881: Performance regression: live computations write to disk, affecting performance
The 'needs_update' value is written out.

cmeyer:
Fixes #1881

Delay computation writes when inputs are write delayed.

Implements 'transaction state' for computations, which is enabled when
one of their inputs goes into transaction state. This state prevents
writing of the computation and avoids writing to disk when 'needs
recompute' (or any other property) changes during live acquisition.

This was probably a regression when 'needs recompute' was made
persistent.

### DIFF
--- a/nion/swift/model/Symbolic.py
+++ b/nion/swift/model/Symbolic.py
@@ -2049,6 +2049,8 @@ class Computation(Persistence.PersistentObject):
         self.__result_base_item_removed_event_listeners: typing.List[Event.EventListener] = list()
         self.__processor: typing.Optional[ComputationProcessor] = None
         self.last_evaluate_data_time = 0.0
+        self.__in_transaction_state = False
+        self.__write_delay_modified_count = 0
         self.__recompute_enabled = True  # used for disabling recompute machinery while close document
         self.__needs_update = True
         self.computation_mutated_event = Event.Event()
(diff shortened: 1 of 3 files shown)


nion/swift/model/Symbolic.py
```diff
@@ -2100,6 +2102,39 @@ def soft_delete(self) -> None:
...
+            if self.modified_count > self.__write_delay_modified_count:
+                self.rewrite()
+
+    def _transaction_state_entered(self) -> None:
+        self.__in_transaction_state = True
+        # first enter the write delay state.
+        self.__enter_write_delay_state()
+
+    def _transaction_state_exited(self) -> None:
+        self.__in_transaction_state = False
```
Tiomat85, line 2126: Minor thing, but for sake of it being the inverse I would put the setting of the flag after the __exit_write_delay_state has completed.

nion/swift/model/Symbolic.py
```diff
@@ -2100,6 +2102,39 @@ def soft_delete(self) -> None:
...
+
+    def __enter_write_delay_state(self) -> None:
+        self.__write_delay_modified_count = self.modified_count
+        if self.persistent_object_context:
+            self.enter_write_delay()
+
+    def __exit_write_delay_state(self) -> None:
+        if self.persistent_object_context:
+            self.exit_write_delay()
+            if self.modified_count > self.__write_delay_modified_count:
```
Tiomat85, line 2117: Do we care about write order?  Exiting write_delay state before we flush the in flight writes could mean _new_ writes get in before the old ones if they are on other threads?

Related: If we change the order so we exit afterwards, can we have other threads add more edits to the item after rewrite but before we clear the flag? Should exit_write_delay do a flush if needed directly?

Tiomat85 (commented):
Questions/comments mostly based around the threading side and how bullet-proof we want this to be.

nion/swift/model/Symbolic.py
cmeyer: In both cases, I copied the code from another class (`DisplayItem`), so I'd prefer to leave it alone just for this PR. The whole transaction thing needs reworking - the existing implementation is error prone and tedious.

cmeyer: > Do we care about write order? Exiting write_delay state before we flush the in flight writes could mean new writes get in before the old ones if they are on other threads?

This is all UI thread code, which is why it's needed in the first place. Persistent objects can't be modified outside the main thread.

nion/swift/model/Symbolic.py
cmeyer: I'm going to rework the transaction code soon; so this is a valid style issue, but I'd like to leave it the same as the other classes using transactions (`DisplayItem`, `DataItem`) for now.

PR #1883, opened 2026-06-23
Text background color and visibility properties

MattRoyle:
Fixes https://github.com/nion-software/nion-internal/issues/304.
Requires #1884 to fix the type check failure introduced by numpy updating.
Adds two properties to the graphics for text background color and text visibility. Text Visibility set to "always" will always show the text, "selection" when the graphic is selected, "focus" when the graphic is in focus. The default is "always" which is overridden by the Interval graphic to "selection" so it only shows the text when the graphic is selected. One issue that was not discussed in the specification is that since the name label and the width, left, and right texts are all driven by the same text visibility the behavior of the name label changes from always showing to only showing on selection. This is a side effect of the implementation that differs from the current behavior and from all other graphics where when there is a name label it will always show.

### DIFF
--- a/nion/swift/LineGraphCanvasItem.py
+++ b/nion/swift/LineGraphCanvasItem.py
@@ -659,16 +659,16 @@ class LineGraphRegionsCanvasItemComposer(CanvasItem.BaseComposer):
                     else:
                         draw_marker(drawing_context, Geometry.FloatPoint(level, mid_x), stroke=selection_color)
                     drawing_context.font = self.font
-                    if region.middle_text and region.style != "tag" and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("width"), region_selected):
-                        _draw_label_with_background(region.middle_text, mid_x, level - self.font_size_metric.height, "center", "bottom", region.get_text_background_color_fn("width") or self.text_background_color)
+                    if region.middle_text and region.style != "tag" and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("width_text"), region_selected):
+                        _draw_label_with_background(region.middle_text, mid_x, level - self.font_size_metric.height, "center", "bottom", region.get_text_background_color_fn("width_text") or self.text_background_color)
                     drawing_context.fill_style = region_color
-                    if region.left_text and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("left"), region_selected):
-                        _draw_label_with_background(region.left_text, left - 5, level, "right", "middle", region.get_text_background_color_fn("left") or self.text_background_color)
-                    if region.right_text and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("right"), region_selected):
-                        _draw_label_with_background(region.right_text, right + 5, level, "left", "middle", region.get_text_background_color_fn("right") or self.text_background_color)
+                    if region.left_text and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("left_text"), region_selected):
+                        _draw_label_with_background(region.left_text, left - 5, level, "right", "middle", region.get_text_background_color_fn("left_text") or self.text_background_color)
+                    if region.right_text and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("right_text"), region_selected):
+                        _draw_label_with_background(region.right_text, right + 5, level, "left", "middle", region.get_text_background_color_fn("right_text") or self.text_background_color)
                     label = region.label
-                    if label and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("label"), region_selected):
-                        _draw_label_with_background(label, mid_x, level + self.font_size_metric.height, "center", "top", region.get_text_background_color_fn("label") or self.text_background_color)
+                    if label and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("label_text"), region_selected):
+                        _draw_label_with_background(label, mid_x, level + self.font_size_metric.height, "center", "top", region.get_text_background_color_fn("label_text") or self.text_background_color)
                     drawing_context.close_path()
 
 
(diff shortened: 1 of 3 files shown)


cmeyer:
This is a good initial PR. A few comments/requests:

The 'used_<property>' are the values that have been resolved as a function of the 'role' property. Let's stick with that idea.

The 'used_<property>' are defined are properties but they really need to resolve according to the "style context" which consists of the graphic type, role, part, selection/focus state, etc.

Let's introduce a new role of 'measurement' and then use the same mechanism 'used_<property>' but make it a function. Something like:

```
def used_text_visibility(self, role: str | None, part: str | None) -> bool: ...
def used_text_background_color(self, role: str | None, part: str | None) -> str: ...
```

Then those can be overridden for `IntervalGraphic` to provide sensible defaults for the 'measurement' role.

Other comments about the PR:
- eliminate the extra line before `Graphic`
- `text_background_color` and `text_visibility` property methods should take `str | none` with `None` indicating 'auto' behavior. note: the `used..` functions are read-only and should return `str` since they should always resolve I think.
- by using `role` and `part` I think you can eliminate most of the `is_selected`, `is_focused` modifications for now.
- there are some line length problems when constructing `RegionInfo`, but those may be eliminated anyway

cmeyer:
Are the `text_background_color` and `text_visibility` fields required at this moment? Can this be done as a 'role' only with no file format modifications? You'd still need `used_background_color` and `used_text_visibility`, but can we add the overrides that require file format changes later when we do the full styling system?

nion/swift/model/Graphics.py
```diff
@@ -925,6 +929,12 @@ def used_fill_style(self) -> typing.Optional[str]:
     def used_stroke_width(self) -> float:
         return self.stroke_width if self.stroke_width is not None else 1.0
 
+    def used_text_visibility(self, part: str) -> str:
+        return self.text_visibility if self.text_visibility is not None else "always"
```
Tiomat85: Assuming "" is not valid, this might be cleaner as
`return self.text_visibility or "always"`

I know this deviates from the way the other functions are doing it, but stylistically it removes scope for one bug (checking the wrong parameter is not None).

nion/swift/LineGraphCanvasItem.py
```diff
@@ -655,23 +656,19 @@ def _draw_label_with_background(label_text: str, label_x: float, label_y: float,
...
-                        if right_text:
-                            _draw_label_with_background(right_text, right + 5, level, "left", "middle")
                     else:
                         draw_marker(drawing_context, Geometry.FloatPoint(level, mid_x), stroke=selection_color)
-
+                    drawing_context.font = self.font
+                    if region.middle_text and region.style != "tag" and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("width"), region_selected):
+                        _draw_label_with_background(region.middle_text, mid_x, level - self.font_size_metric.height, "center", "bottom", region.get_text_background_color_fn("width") or self.text_background_color)
+                    drawing_context.fill_style = region_color
+                    if region.left_text and Graphics.GraphicRenderer.is_label_visible(region.get_text_visibility_fn("left"), region_selected):
```
Tiomat85: These names, "left", "right", "width" and "label" - would we be better having them match the attribute names in the region so when drawing the "right_text" we check the visibility of "right_text".

Tiomat85 (commented):
Couple of comments, but generally it looks like it roughly matches existing patterns and extends in a small way like we agreed.  It highlights the need for a bigger rethink on styling but that is massively out-of-scope for this.

nion/swift/model/LinePlotDisplay.py
```diff
@@ -33,6 +33,8 @@ class RegionInfo:
     label: typing.Optional[str]
     style: typing.Optional[str]
     color: typing.Optional[str]
+    get_text_visibility_fn: typing.Callable[[str], str | None]  # Returns the visibility of the passed in text part
```
MattRoyle: From @cmeyer;
> used_text_visibility_fn goes directly against the dataclass nature of RegionInfo 
> RegionInfo is a data only class, so adding a function callback is a sneaky way of changing its intrinsic nature. > So as written, I would reject this PR on that point only.

> Some implications of the dataclass and immutable nature of RegionInfo: the parts must be resolved at 
RegionInfo __init__ time. You can put a dict with a mapping of part to style; but it has to be static. This is required for the rendering system.  One approach might be that the Graphic declares its parts and they get resolved when creating RegionInfo . Another approach (my preferred approach) is slightly more complicated: introduce a Stylesheetdataclass which is immutable and store it in RegionInfo instead of any styles. Then the Stylesheetbecomes a stand-in for the callbacks. But it is immutable (a frozen dataclass) and has methods to resolve the style for a specific part. This is essentially moving used_text_visibility into that class. In the first pass, you can just resolve the new fields; but eventually we'll want to expand this to handle all the cases. Stylesheet can be specific for now to text visibility, etc., but eventually will hold the stylesheet for all graphics. (this is kind of vague, so there might be things I'm not thinking about, but this seems like a good general approach; it will probably take a few passes to get it just right).

If this is the primary issue with the PR then I have a solution that aligns with how the proposed version #1900 does of passing a class that has a resolve method that can be called to get a property value. Fundamentally though if adding a function callback into the dataclass is an issue then the same is true of passing a class with a method.

The Stylesheetdataclass seems unnecessary, why not just pass in the entire GraphicRenderer into the region info which is an more or less what is being described. https://github.com/nion-software/nionswift/blob/8ed2817a1e1ec94206af7cc0cac58773f01c212e/nion/swift/model/Graphics.py#L1015

cmeyer: If you have a quick fix, submit the PR and I'll re-review. I think we want the stylesheets in the long run, but it's fairly easy for me to rebase/adapt that PR to whatever solution comes out of this PR. If we're close enough on general stylesheets, we can try to get that merged and then put this PR on top of that one. But that may drag on a bit.

MattRoyle: In order to get rid of the function callbacks in the dataclass I have changed the used_ properties for visibility and background color to instead return a look up dictionary that can be resolved in the drawing of the text. Obviously this is a step closer to a CSS like version than previously discussed but without going full in with a stylesheet for multiple properties. This is not intended to be how the CSS version will look like I just wanted to clarify this is very much the quick fix that can go in before properly refactoring with astylesheet version.

cmeyer:
Merged 134a516f5391727e001f275feaa3d3efda556d17. Thanks!

PR #1886, opened 2026-06-24
Use the verify filename utility function in the export dialog

### LINKED ISSUE #1885: Use verify filename in the export dialog
In the export dialog the prefix is checked using simplify filename and if the result makes any changes then a generic error message "Prefix contains invalid characters" is displayed. 
This means prefixing with illegal filenames (CON, NUL, AUX, etc) is prevented even though the prefix may not be the whole file when other options are enabled. 
The check should use the verify filename utility function #1849 so the error messages for the files are more explicit. Also only the illegal characters should be checked when there are other options.

MattRoyle:
Fixes #1885.
Changes the check of the prefix to use the verify filename utility function when it is the only option enabled otherwise only the illegal characters are checked.

### DIFF
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -81,11 +81,12 @@ class ExportDialogViewModel:
             non_prefix_options_enabled = include_title or include_date or include_dimensions or include_sequence
             if prefix_enabled and non_prefix_options_enabled:
                 # The prefix will only be part of the filename so only check that it is made up of valid characters
-                illegal_chars = re.findall(Utility.ILLEGAL_FILENAME_CHARS_REGEX, prefix_str)
-                if len(illegal_chars) == 1:
-                    invalid_reasons.append(_("Prefix contains illegal character") + f" \"{illegal_chars[0]}\"")
-                elif len(illegal_chars) > 1:
-                    invalid_reasons.append(_("Prefix contains illegal characters") + f" {illegal_chars}")
+                matches = re.findall(Utility.ILLEGAL_FILENAME_CHARS_REGEX, prefix_str)
+                matches = sorted(set(matches))
+                if len(matches) == 1:
+                    invalid_reasons.append(_("Prefix contains illegal character") + f" {matches[0]}")
+                elif len(matches) > 1:
+                    invalid_reasons.append(_("Prefix contains illegal characters {characters}").format(characters="".join(matches)))
             elif prefix_enabled and not non_prefix_options_enabled:
                 # The prefix is the full filename so check it is allowed using the verify filename utility
                 is_valid, errors = Utility.verify_filename_is_legal(prefix_str)
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -70,53 +70,6 @@ class ExportDialogViewModel:
         if not self.is_directory_valid:
             self.directory.value = None  # Clear the initial directory if it was invalid
 
-        def compute_export_options_validity(prefix_value: str | None, include_title: bool | None,
-                                            include_date: bool | None, include_dimensions: bool | None,
-                                            include_sequence: bool | None, is_directory_valid: bool | None) -> ExportOptionsValidity:
-            """Use the combined values of the export filename options and the directory validity to get a list of invalid reasons."""
-            invalid_reasons: list[str] = []
-
-            prefix_str = prefix_value or str()
-            prefix_enabled = prefix_str != str()
-            non_prefix_options_enabled = include_title or include_date or include_dimensions or include_sequence
-            if prefix_enabled and non_prefix_options_enabled:
-                # The prefix will only be part of the filename so only check that it is made up of valid characters
-                matches = re.findall(Utility.ILLEGAL_FILENAME_CHARS_REGEX, prefix_str)
-                matches = sorted(set(matches))
-                if len(matches) == 1:
-                    invalid_reasons.append(_("Prefix contains illegal character") + f" {matches[0]}")
-                elif len(matches) > 1:
-                    invalid_reasons.append(_("Prefix contains illegal characters {characters}").format(characters="".join(matches)))
-            elif prefix_enabled and not non_prefix_options_enabled:
-                # The prefix is the full filename so check it is allowed using the verify filename utility
-                is_valid, errors = Utility.verify_filename_is_legal(prefix_str)
-                if not is_valid and errors is not None:
-                    invalid_reasons.append(_("Prefix was not a valid filename:"))
-                    invalid_reasons.extend(errors)
-            elif not prefix_enabled and not non_prefix_options_enabled:
-                # There must be at least one option enabled otherwise the filenames will be empty
-                invalid_reasons.append(_("Filename requires at least one option to be selected"))
-
-            if not is_directory_valid:
-                invalid_reasons.append(_("Directory does not exist"))
-
-            return ExportOptionsValidity(bool(is_directory_valid), tuple(invalid_reasons))
-
-        def handle_export_options_validity_changed(export_validity: ExportOptionsValidity | None) -> None:
-            """Update the export button's enabled state and tooltip based on the computed invalid reasons and directory validity."""
-            export_validity = export_validity or ExportOptionsValidity(False, tuple())
-            if not export_validity.is_directory_valid:
-                self.directory.value = None  # Clear the directory if it is invalid
-
-            if export_validity.invalid_reasons:
-                self.export_button_enabled.value = False
-                self.export_button_tool_tip.value = ", ".join(export_validity.invalid_reasons)
-                self.directory_warning.value = "\n".join(export_validity.invalid_reasons)
-            else:
-                self.export_button_enabled.value = True
-                self.export_button_tool_tip.value = None
-                self.directory_warning.value = str()
-
         export_filename_option_streams: typing.Sequence[Stream.PropertyChangedEventStream[str | bool]] = [
             # Update the button when one of the options changes
             Stream.PropertyChangedEventStream(self.prefix, "value"),
@@ -129,13 +82,64 @@ class ExportDialogViewModel:
         ]
 
         self.__export_button_update_action = Stream.ValueStreamAction(
-            Stream.CombineLatestStream(export_filename_option_streams, compute_export_options_validity),
-            handle_export_options_validity_changed
+            Stream.CombineLatestStream(export_filename_option_streams, self.__compute_export_options_validity),
+            self.__handle_export_options_validity_changed
         )
+        self.update_options()
 
-        handle_export_options_validity_changed(compute_export_options_validity(self.prefix.value, self.include_title.value,
-                                                                               self.include_date.value, self.include_dimensions.value,
-                                                                               self.include_sequence.value, self.__is_directory_valid.value))
+    @staticmethod
+    def __compute_export_options_validity(prefix_value: str | None, include_title: bool | None,
+                                        include_date: bool | None, include_dimensions: bool | None,
+                                        include_sequence: bool | None, is_directory_valid: bool | None) -> ExportOptionsValidity:
+        """Use the combined values of the export filename options and the directory validity to get a list of invalid reasons."""
+        invalid_reasons: list[str] = []
+
+        prefix_str = prefix_value or str()
+        prefix_enabled = prefix_str != str()
+        non_prefix_options_enabled = include_title or include_date or include_dimensions or include_sequence
+        if prefix_enabled and non_prefix_options_enabled:
+            # The prefix will only be part of the filename so only check that it is made up of valid characters
+            is_valid, error = Utility.verify_filename_has_no_illegal_characters(prefix_str)
+            if not is_valid and error is not None:
+                invalid_reasons.append(_("Prefix ") + error[0].lower() + error[1:])
+        elif prefix_enabled and not non_prefix_options_enabled:
+            # The prefix is the full filename so check it is allowed using the verify filename utility
+            is_valid, errors = Utility.verify_filename_is_legal(prefix_str)
+            if not is_valid and errors is not None:
+                invalid_reasons.append(_("Prefix was not a valid filename:"))
+                invalid_reasons.extend(errors)
+        elif not prefix_enabled and not non_prefix_options_enabled:
+            # There must be at least one option enabled otherwise the filenames will be empty
+            invalid_reasons.append(_("Filename requires at least one option to be selected"))
+
+        if not is_directory_valid:
+            invalid_reasons.append(_("Directory does not exist"))
+
+        return ExportOptionsValidity(bool(is_directory_valid), tuple(invalid_reasons))
+
+    def __handle_export_options_validity_changed(self, export_validity: ExportOptionsValidity | None) -> None:
+        """Update the export button's enabled state and tooltip based on the computed invalid reasons and directory validity."""
+        export_validity = export_validity or ExportOptionsValidity(False, tuple())
+        if not export_validity.is_directory_valid:
+            self.directory.value = None  # Clear the directory if it is invalid
+
+        if export_validity.invalid_reasons:
+            self.export_button_enabled.value = False
+            self.export_button_tool_tip.value = ", ".join(export_validity.invalid_reasons)
+            self.directory_warning.value = "\n".join(export_validity.invalid_reasons)
+        else:
+            self.export_button_enabled.value = True
+            self.export_button_tool_tip.value = None
+            self.directory_warning.value = str()
+
+    def update_options(self, prefix_value: str | None = None) -> None:
+        """Directly refresh the validity of the export button's enabled state and tooltip."""
+        self.__handle_export_options_validity_changed(
+            self.__compute_export_options_validity(
+                prefix_value or self.prefix.value, self.include_title.value, self.include_date.value,
+                self.include_dimensions.value, self.include_sequence.value, self.__is_directory_valid.value
+            )
+        )
 
     @property
     def directory_path_object(self) -> pathlib.Path:
@@ -268,6 +272,9 @@ class ExportDialog(Declarative.Handler):
         writer = self.__writers[current_index]
         self.viewmodel.writer.value = writer
 
+    def handle_prefix_edited(self, widget: UserInterface.LineEditWidget, text: str) -> None:
+        self.viewmodel.update_options(text)
+
     def __build_ui(self, u: Declarative.DeclarativeUI) -> None:
         """Build the UI and store it in self.ui_view. This is called from __init__."""
 
@@ -303,7 +310,7 @@ class ExportDialog(Declarative.Handler):
 
         # Prefix. Include a messy callback so that prefix gets updated during typing.
         prefix_label = u.create_label(text=_("Prefix:"))
-        prefix_textbox = u.create_line_edit(text="@binding(viewmodel.prefix.value)", placeholder_text=_("None"), width=230)
+        prefix_textbox = u.create_line_edit(text="@binding(viewmodel.prefix.value)", placeholder_text=_("None"), width=230, on_text_edited="handle_prefix_edited")
         prefix_row = u.create_row(prefix_label, prefix_textbox, u.create_stretch(), spacing=10)
 
         # File Type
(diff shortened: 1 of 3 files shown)


nion/swift/ExportDialog.py
```diff
@@ -77,10 +77,22 @@ def compute_export_options_validity(prefix_value: str | None, include_title: boo
...
-            if not (prefix_str or include_title or include_date or include_dimensions or include_sequence):
+            prefix_enabled = prefix_str != str()
+            non_prefix_options_enabled = include_title or include_date or include_dimensions or include_sequence
+            if prefix_enabled and non_prefix_options_enabled:
+                # The prefix will only be part of the filename so only check that it is made up of valid characters
+                illegal_chars = re.findall(Utility.ILLEGAL_FILENAME_CHARS_REGEX, prefix_str)
+                if len(illegal_chars) == 1:
+                    invalid_reasons.append(_("Prefix contains illegal character") + f" \"{illegal_chars[0]}\"")
+                elif len(illegal_chars) > 1:
+                    invalid_reasons.append(_("Prefix contains illegal characters") + f" {illegal_chars}")
```
KRLango: This will print the list of illegal characters used as `"["*", "/"]"` This is not really readable. I would suggest something like:

```python
illegal_chars = set(re.findall(Utility.ILLEGAL_FILENAME_CHARS_REGEX, prefix_str))
invalid_reasons.append(_("Prefix contains illegal characters") + " " +  "".join(illegal_chars")
```

Ion-e:
LGTM

nion/swift/test/Utility_test.py
```diff
@@ -78,11 +78,11 @@ def test_verify_path_is_legal_catches_illegal_names(self) -> None:
                           ("file.", "Cannot end with a period"),
                           ("file ", "Cannot end with a whitespace"),
                           ("COM¹", "\"COM¹\" is illegal as it is reserved on some platforms"),
-                          ("5>3>2024", "Contains illegal character '>'"),
+                          ("5>3>2024", "Contains illegal character >"),
                           ("1234567890" * 13, "Exceeds the allowed length of 128 characters"),
                           ("NUL", "\"NUL\" is illegal as it is reserved on some platforms"),
-                          ("abc\n\rdef", "Contains illegal characters ['\\n', '\\r']"),
-                          ("\26", "Contains illegal character '\x16'")]
+                          ("abc\n\rdef", "Contains illegal characters \n\r"),
```
cmeyer: This doesn't look right. Was this intentional? Now \n and \r are directly embedded in the strings and you're not checking printable characters. The LLM audit caught this too.

cmeyer (changes requested):
Why are the characters being unquoted?

Also, \n and \r are now directly embedded (mishandling backslash).

nion/swift/test/Utility_test.py
cmeyer: To be clear, the problematic part is `"Contains illegal characters \n\r"`.

MattRoyle: Yeah, that was a mistake, I was trying to change the output from a printed list. 
I have updated it so the repr is manually called for each element (before printing the list did that anyway). I added special case for \ as I don't want it to print \\\\. Also the \r and \n have Carriage Return and New Line prepended respectively since this is a fairly common case when copying and pasting and I think the names are more informative to a user than the representation.
I have changed the verify_filename utility so there is a separate function that checks the illegal characters rather than duplicating the code in the export dialog. I also updated the viewmodel so that the validity can be refreshed when the prefix text is edited using the on_text_edited callback.

cmeyer (changes requested):
Overall looks good, but I'm going to use a slightly different merge process here and push the final minor edit responsibility back to you.

Issues and actions (most of these are also identified by audit, btw):
- the indentation on the call to `__handle_export_options_validity_changed` is wonky
- `handle_options_changed` should be named something else like `update_options` or something more descriptive. it's not "handling" anything
- `prefix_edited` should be renamed to `handle_prefix_edited`. it _is_ "handling" something
- there is no blank line before `directory_path_object` property def
- can you squash/fixup/rebase into a small number of commits (or a single one)? then `git push --force-with-lease`? each commit should be a self-contained reviewable change with proper comment.
- if you want to keep commit history "just in case", back up your local branch before squashing/fixup/rebase. i often compare the original vs. the cleaned up branch as a double-check before force pushing.

```
git rebase origin/master
git push --force-with-lease
```

nion/swift/model/Utility.py
```diff
@@ -478,26 +506,24 @@ def verify_filename_is_legal(filename: str, maximum_length: int = 128) -> tuple[
...
 def simplify_filename(filename: str, replacement_char: str = '_', maximum_length: int = 128) -> str:
-    # This function replaces any illegal characters in a file name. Not a full path.
-    # macOS illegal characters = \0 /
-    # linux illegal characters = \0 /
-    # Windows illegal characters = < > : / \ | ? * " ASCII values 0-31 (non-printable chars)
-    # Windows illegal filenames = CON PRN AUX NUL COM1 COM2 COM3 COM4 COM5 COM6 COM7 COM8 COM9
-    #                             LPT1 LPT2 LPT3 LPT4 LPT5 LPT6 LPT7 LPT8 LPT9 COM¹ COM² COM³ LPT¹ LPT² LPT³
-    # those files names are illegal with or without a suffix, upper and lower case
-    # Windows file names cannot start or end with periods or spaces
+    r"""This function replaces any illegal characters in a file name. Not a full path.
```
cmeyer: Let's switch this to a non-raw string comment. As it is now, it doesn't seem like it'll play well with tooling.

If needed, you can spell out the characters that need escaping, or use the named Unicode characters if that's easier.

Also worth reconsidering: does this level of detail need to be in the comment? It's already clear from the code, so it might not add much for the reader.

MattRoyle: I've removed the raw string. I've left the comment as is since I think it helps give the context of how the function works, but I can simplify if it is necessary.

MattRoyle:
> Overall looks good, but I'm going to use a slightly different merge process here and push the final minor edit responsibility back to you.
> 
> Issues and actions (most of these are also identified by audit, btw):
> 
>     * the indentation on the call to `__handle_export_options_validity_changed` is wonky
> 
>     * `handle_options_changed` should be named something else like `update_options` or something more descriptive. it's not "handling" anything
> 
>     * `prefix_edited` should be renamed to `handle_prefix_edited`. it _is_ "handling" something
> 
>     * there is no blank line before `directory_path_object` property def
> 
>     * can you squash/fixup/rebase into a small number of commits (or a single one)? then `git push --force-with-lease`? each commit should be a self-contained reviewable change with proper comment.
> 
>     * if you want to keep commit history "just in case", back up your local branch before squashing/fixup/rebase. i often compare the original vs. the cleaned up branch as a double-check before force pushing.
> 
> 
> ```
> git rebase origin/master
> git push --force-with-lease
> ```

I've reduced the changes to three commits that I think show the different logical changes in the PR.

KRLango (approved):
LGTM

nion/swift/ExportDialog.py
```diff
@@ -116,13 +82,64 @@ def handle_export_options_validity_changed(export_validity: ExportOptionsValidit
...
+        self.update_options()
 
-        handle_export_options_validity_changed(compute_export_options_validity(self.prefix.value, self.include_title.value,
-                                                                               self.include_date.value, self.include_dimensions.value,
-                                                                               self.include_sequence.value, self.__is_directory_valid.value))
+    @staticmethod
+    def __compute_export_options_validity(prefix_value: str | None, include_title: bool | None,
+                                        include_date: bool | None, include_dimensions: bool | None,
+                                        include_sequence: bool | None, is_directory_valid: bool | None) -> ExportOptionsValidity:
+        """Use the combined values of the export filename options and the directory validity to get a list of invalid reasons."""
```
cmeyer: Indentation is still wrong in `__compute_export_options_validity`.

Also, can you either (a) move this function outside of the class, (b) make it a regular method, or (c) keep it static but remove the underscores and call it using the class name `ExportDialogViewModel.compute_export_options_validity`? The way it reads now and is called is confusing.

nion/swift/ExportDialog.py
```diff
@@ -116,13 +82,64 @@ def handle_export_options_validity_changed(export_validity: ExportOptionsValidit
...
+        invalid_reasons: list[str] = []
+
+        prefix_str = prefix_value or str()
+        prefix_enabled = prefix_str != str()
```
cmeyer: `prefix_enabled = bool(prefix_str)`

nion/swift/model/Utility.py
```diff
@@ -453,10 +453,38 @@ def do_sample() -> None:
                      'LPT1', 'LPT2', 'LPT3', 'LPT4', 'LPT5', 'LPT6', 'LPT7', 'LPT8', 'LPT9', 'COM¹', 'COM²', 'COM³', 'LPT¹', 'LPT²', 'LPT³']
 
 
+def verify_filename_has_no_illegal_characters(filename: str) -> tuple[bool, str | None]:
```
cmeyer: This method is returning redundant information and makes the caller's job more difficult. `True` is only returned when the string is `None` and `False` is only returned when the string is `not None`.

How about renaming it to `get_filename_validation_error` and just returning the error string or None. That makes the callers life much easier.

nion/swift/model/Utility.py
```diff
@@ -453,10 +453,38 @@ def do_sample() -> None:
...
+        else:
+            match = repr(match)  # repr ensures the character is printable and has quotations
+        if match not in illegal_chars:
+            illegal_chars.append(match)
+
+    if illegal_chars:
+        if len(illegal_chars) == 1:
+            return False, _("Contains illegal character {character}").format(character=illegal_chars[0])
+        else:
+            return False, _("Contains illegal characters {characters}").format(characters=", ".join(illegal_chars))
```
cmeyer: How about eliminating this duplicated code by using: `"Filename contains illegal characters: {characters}"`?

nion/swift/model/Utility.py
```diff
@@ -453,10 +453,38 @@ def do_sample() -> None:
...
+    return True, None
+
+
 def verify_filename_is_legal(filename: str, maximum_length: int = 128) -> tuple[bool, typing.Sequence[str] | None]:
```
cmeyer, line 482: `verify_filename_is_legal` should probably also be changed in a future PR to `get_filename_errors` and just return the errors. The `bool` is redundant and makes the calling code more difficult.

nion/swift/model/Utility.py
```diff
@@ -453,10 +453,38 @@ def do_sample() -> None:
...
+def verify_filename_has_no_illegal_characters(filename: str) -> tuple[bool, str | None]:
+    """Check if a filename contains illegal characters.
+
+    Returns a tuple of a bool indicating if the filename is valid and an error message or None when it is valid.
+    The case of a \r or \n are likely when pasting but using those representations are not clear so use the name of the characters instead.
+    """
+    matches = re.findall(ILLEGAL_FILENAME_CHARS_REGEX, filename)
+    illegal_chars = []
+    for match in matches:
+        if match == "\\":
```
cmeyer: Is showing users quoted Python strings of illegal characters really what we want to do here? My answer is a resounding "NO!". As before, I question this whole approach. I think it is sufficient for a project like this to just say "Filename contains illegal characters." and be on with it.

nion/swift/ExportDialog.py
```diff
@@ -116,13 +82,64 @@ def handle_export_options_validity_changed(export_validity: ExportOptionsValidit
...
+            self.directory_warning.value = str()
+
+    def update_options(self, prefix_value: str | None = None) -> None:
+        """Directly refresh the validity of the export button's enabled state and tooltip."""
+        self.__handle_export_options_validity_changed(
+            self.__compute_export_options_validity(
+                prefix_value or self.prefix.value, self.include_title.value, self.include_date.value,
+                self.include_dimensions.value, self.include_sequence.value, self.__is_directory_valid.value
+            )
+        )
```
cmeyer, line 141: Can you put the result of the `__compute_export_options_validity` function call into a `export_validity` local variable and pass that to `__handle_export_options_validity_changed`? The way it's written now is difficult to read and understand without extra mental effort.

cmeyer (changes requested):
Can you improve the two commit messages so someone reading the commit log knows what they refer to? Suggestions:

- "Refactor the printing of illegal characters" -> "Refactor the printing of illegal characters in filename verification."
- "Refactor so the button updates when the prefix is edited" -> "Refactor export dialog so the button updates when the prefix is edited."

nion/swift/model/Utility.py
MattRoyle: I think previously that having the success be determined when the error list was None had been noted as a weird way of doing it by @KRLango for the verify filename function. I kept the same principle here of returning the is_valid seperately from the error message so it is more explicit for the caller

cmeyer: It's weird if the function isn't named properly, but `get_filename_validation_error` makes clear what it is doing.

nion/swift/model/Utility.py
MattRoyle: The first is when there is a single illegal character the second is when there are character*s* so is a plural. There could be a way to do this in a single line like "Contains illegal character(s) {characters}" if that is better?

nion/swift/model/Utility.py
MattRoyle: On the first point the reason the single quotes are added is to keep consistency with the output of the repr method which returns the character as printable with single quotes around it. In the previous version when printing the entire list it would do this anyway since printing a list in python calls the __repr__ for each item i.e. "['\\n', '\\r']". The case with one item however did not use the repr meaning that a carriage return character would render as a new line in the output. Annoyingly __repr__ on a backslash \\ will cause it to be '\\\\' rather than '\\' and It is far more user friendly to use Carriage return and new line than the \\r \\n

> "Filename contains illegal characters." and be on with it.

The whole point of the function is to inform a user on why what they have typed is invalid, if they cannot see what character is invalid and they don't know what the illegal characters present are then it fails the purpose. 
Try using a password input that does not tell you which character you have typed is an illegal character vs one that does.

nion/swift/model/Utility.py
cmeyer: This is just more code that necessary for an minor feature. Any given special case is arguable one way or another, but the aggregate of many special cases like this makes the code difficult to work on. Tell the user the filename is being rejected for "illegal characters" is sufficient in my opinion. It would be confusing to tell them that "\n" as a character is not allowed. That is a programmer thing, not an end user thing.

nion/swift/model/Utility.py
cmeyer: As before, I don't think we should do this. I think this code should be minimal or non-existent. This is why. It's difficult to get right. This is not the focus of the application and adds little value. So the code should reflect it's low value by being as minimal as possible. At the minimum it shouldn't convey "programmer info" to the end user. End users do not know what "\r" means.

nion/swift/ExportDialog.py
Tiomat85: Is that indicator even needed?

You can check `if prefix_str` instead of the enabled flag, passing None (or "") to the functions can be handled inside those with `if prefix` blocks.

If you want to ensure a string is passed, make those pass `prefix_value or ""` instead.

nion/swift/model/Utility.py
Tiomat85: If you are wanting to add a bit more help to the user focussed error message (not a bad thing to try and do) then the simplest thing would be to add a list of acceptable characters (non-exhaustive!) that cover most cases.

Something like
"Acceptable characters are alphanumeric, ' ' and _".  That will cover 99% of bases, not add weird characters that users may not understand and is fixed.

MattRoyle: > . At the minimum it shouldn't convey "programmer info" to the end user. End users do not know what "\r" means.

That is exactly what the change is doing, it is changing it from being "Contains invalid character '\r'" to being "Contains invalid character '\r' (Carriage Return)". The code could for writing this is as minimal as possible to enable this while remaining readable.

^ Specifically it first changes it so that the singular actually prints the '\r' instead of a new line, although the plural would print '\r' as previously mentioned in regarding the consistancy

MattRoyle: > Something like "Acceptable characters are alphanumeric, ' ' and _". That will cover 99% of bases, not add weird characters that users may not understand and is fixed.

The enabled characters is a blacklist not a whitelist. There are very few characters that are not allowed so changing this would be needlessly restrictive. Besides all the unicode non-latin characters that are available most symbols are fine ^&!%@ all work in filenames. So I don't think that would work.

The other way of conveying to the end user what is allowed would be to detect the invalid character and print a standard "Contained Invalid Character(s) <, >, :, /, \, |, ?, *, ", Carriage Return or New Line." That would be a much longer message and I would say not as user friendly as they have to read the whole list rather than it just telling you what character is incorrect but is better than not telling at all. This would also take up multiple lines which isn't terrible but just a slight nitpick. The unprintable characters (integer representations 0 and 31) also obviously could not be put in this message.

Tiomat85: Looking at Microsoft styleguides:

Something along the lines of:
The file name contains non-printable characters.
Please remove any hidden or control characters and try again.

Might help get across the problem for some of the weirder characters, and the word description maybe for CR\NL and Tabs.  Obvious concern though is getting caught up in a spiral of trying to find a perfect solution to a problem that does not have one, when a solution that captures the 99% case is better than what we currently have.

cmeyer: My objection is mainly putting in things like "\n" for an end user, which they won't understand. The Microsoft wording is fine with me. My secondary objection is singular/plural special case code which just make maintenance difficult and can be avoided by choosing simplified wording.

nion/swift/ExportDialog.py
MattRoyle: I think it the enabled flag was requested last week when you and Kev had a look at the code for readability

nion/swift/model/Utility.py
MattRoyle: It is not doing the full validation so get_filename_validation_error isn't correct. It could be get_filename_illegal_chars_error, but then it would be inconsistent with the verify_filename_is_legal and I don't think changing the name resolves the weirdness of the positive case being None.

cmeyer (changes requested):
Is this final? I would continue to reject this PR "as is" for multiple reasons - namely the redundant return types of the functions, the anti-pattern of displaying backslash characters to end users, and the other issue I raised earlier today about flashing error messages while typing valid filenames like "My Export.jpeg" (flashes errors for spaces and periods at the end).

cmeyer:
The dialog no longer works:

image

```
diff --git a/nion/swift/ExportDialog.py b/nion/swift/ExportDialog.py
index 77a79527..339d3cdc 100644
--- a/nion/swift/ExportDialog.py
+++ b/nion/swift/ExportDialog.py
@@ -100,13 +100,13 @@ class ExportDialogViewModel:
         if prefix_enabled and non_prefix_options_enabled:
             # The prefix will only be part of the filename so only check that it is made up of valid characters
             errors = Utility.get_filename_illegal_chars_error(prefix_str)
-            if errors is not None:
+            if errors:
                 invalid_reasons.append(_("Prefix was invalid:"))
                 invalid_reasons.extend(errors)
         elif prefix_enabled and not non_prefix_options_enabled:
             # The prefix is the full filename so check it is allowed using the verify filename utility
             is_valid, errors = Utility.verify_filename_is_legal(prefix_str)
-            if not is_valid and errors is not None:
+            if not is_valid and errors:
                 invalid_reasons.append(_("Prefix was not a valid filename:"))
                 invalid_reasons.extend(errors)
         elif not prefix_enabled and not non_prefix_options_enabled:
```

Also, again I ask if we can just show the message "Prefix not valid." rather than including a reason, which is very difficult to maintain.

On macOS, filenames cannot start with `.`.

cmeyer:
Also, the dialog changes height depending on if the specific error is displayed or not. At the very minimum, how about just a one line error that doesn't change the size of the dialog while typing?

cmeyer (changes requested):
Any value in 'prefix' shows an error.

MattRoyle:
> The dialog no longer works:

I have fixed the prefix so it now works correctly.

> On macOS, filenames cannot start with `.`.

Files can start with periods on macos and linux but become hidden files.

cmeyer:
> Files can start with periods on macos and linux but become hidden files.

That's the point. If we're preventing user errors - this is an obvious one.

cmeyer:
Merged 7ddd1ac8bd8a647add4653d1a5238033614dd2b1. Thanks! Open issue is still clean up two line error so that dialog doesn't change size when typing.

PR #1887, opened 2026-06-24
Make calls to auto-legend more explicit. Fixes SVG export issue.

### LINKED ISSUE #1776: SVG ignores legend position when exporting
When exporting to SVG if you set the legend to be in a certain position in the inspector this position is not propagated through to the export. 

Repro case:
Create a composite line plot with 2 line graphs
Set the legend to be outside left on the inspector
Export to SVG
Open the SVG and the legend will be in the top right

This is the same with trying to remove the legend by setting it to None it returns to the top right corner

cmeyer:
Fixes #1776

The legend gets enabled/disabled automatically when a line plot display goes from one item to two or two to one.
This change makes that more explicit and adds a test. This wasn't a good solution in the first place and this is
only a band-aid on top of that original solution. This fixes the issue with the legend reverting to top-right
when exporting to SVG. In that case, the legend got reset during snapshot. Moving the 'auto_legend' call out of
the low level fixes that issue. Snapshot was probably also failing in the same way SVG was failing, too.

### DIFF
--- a/nion/swift/model/DisplayItem.py
+++ b/nion/swift/model/DisplayItem.py
@@ -2218,7 +2218,7 @@ class DisplayItem(Persistence.PersistentObject):
         self.define_property("session_id", hidden=True, changed=self.__property_changed)
         self.define_property("calibration_style_id", "calibrated", hidden=True, changed=self.__property_changed)
         self.define_property("intensity_calibration_style_id", "calibrated", hidden=True, changed=self.__property_changed)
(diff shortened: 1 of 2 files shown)


lisham2000 (approved):
Checked and solves problem in a much more elegant way than the patch I had for it. This is the approach I have been looking for but never found. Useful for learning where this stuff is done thank you.

MattRoyle (approved):
LGTM

PR #1889, opened 2026-06-30
Dragging to split panel from selection

MattRoyle:
Fixes https://github.com/nion-software/nion-internal/issues/312. 
Requires https://github.com/nion-software/nionui/pull/137.
Adds the ability to split a display panel by dragging a selection of multiple items. This PR can be summarized in three parts: 
1. Fix the multi-drag code. 
2. Add a function in the workspace to split and insert items into a workspace adding this to the drag handler.
3. Add UI to draw a grid preview to show the user what the display panel split will look like after the split action.

### DIFF
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -2140,11 +2140,14 @@ class DisplayPanel(CanvasItem.LayerCanvasItem):
                         thumbnail_data = Image.get_rgba_data_from_rgba(
                             Image.scaled(Image.get_rgba_view_from_rgba_data(thumbnail_data), tuple(Geometry.IntSize(w=80, h=80))))
                 elif len(display_items) > 1:
+                    # Thumbnail preference order: The display item the drag started from, then the anchor item, then the first display item in the list
+                    if display_item is None and self.__selection.anchor_index is not None:
+                        display_item = document_controller.display_items_model.items[self.__selection.anchor_index]
+                    else:
+                        display_item = display_items[0]
                     mime_data = document_controller.ui.create_mime_data()
                     MimeTypes.mime_data_put_display_items(mime_data, display_items)
-                    anchor_index = self.__selection.anchor_index or 0
-                    thumbnail_display_item = display_items[anchor_index]
-                    thumbnail_source = Thumbnails.ThumbnailManager().thumbnail_source_for_display_item(self.__ui, thumbnail_display_item)
+                    thumbnail_source = Thumbnails.ThumbnailManager().thumbnail_source_for_display_item(self.__ui, display_item)
                     thumbnail_data = thumbnail_source.thumbnail_data
                 return mime_data, thumbnail_data
 
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -30,6 +30,7 @@ from nion.swift import MimeTypes
 from nion.swift import Panel
 from nion.swift import Thumbnails
 from nion.swift import Undo
+from nion.swift import Workspace
 from nion.swift.model import Changes
 from nion.swift.model import DataItem
 from nion.swift.model import DisplayInfo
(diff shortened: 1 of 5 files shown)


cmeyer (commented):
👀

I didn't review closely or even try it, but no big red flags in the code upon initial read through. However, it would be useful to add tests where warranted. That may involve refactoring some of the functions to make them easier to test.

nion/swift/DisplayPanel.py
```diff
@@ -197,6 +200,16 @@ def _repaint(self, drawing_context: DrawingContext.DrawingContext, canvas_bounds
                         drawing_context.rect(0, 0, canvas_bounds.width, int(canvas_bounds.height * 0.10))
                     elif drop_region == "bottom":
                         drawing_context.rect(0, int(canvas_bounds.height * 0.90), canvas_bounds.width, int(canvas_bounds.height - canvas_bounds.height * 0.90))
+                    elif self.__drag_items_split and self.__drag_items_split != (1, 1):
+                        # Draw a grid of panels to preview what the split will look like
+                        gap = 4  # Each panel is separated by a gap of 4px
+                        horizontal, vertical = self.__drag_items_split
+                        # The panel width and height can be calculated as: n + 1 gaps + n panel widths equals the total canvas size which is rearranged to get the panel width/height
```
KRLango: I assume the n+1 gaps are the n-1 gaps between the panels, + 2 for a border. If that is the case, it should be made more clear in the comment.

nion/swift/Workspace.py
```diff
@@ -793,42 +791,43 @@ def handle_drop(self, destination_display_panel: DisplayPanel.DisplayPanel, mime
...
+            return_type = "move"
+        elif mime_data.has_format(MimeTypes.DISPLAY_ITEMS_MIME_TYPE):
+            source_display_items = MimeTypes.mime_data_get_display_items(mime_data, self.document_model)
+            return_type = "copy"
+        elif mime_data.has_format(MimeTypes.DISPLAY_ITEM_MIME_TYPE):
+            source_display_item = MimeTypes.mime_data_get_display_item(mime_data, self.document_model)
+            return_type = "copy"
+        elif mime_data.has_format("text/uri-list"):
+            index = len(self.document_model.data_items)
+            display_items = self.document_controller.receive_files(list(reversed(mime_data.file_paths)), None, index)
```
KRLango: I think this line needs a comment to say why you are reversing the list

KRLango (commented):
A couple of comments that I think need some work. Otherwise LGTM

nion/swift/Workspace.py
MattRoyle: Not sure why I was reversing it. I have removed that now.

Ion-e:
Verified as failed:
Issue 1:
Click and drag 1d data from data panel into workspace 1d data without dropping it
Select that 1D data and another 1D data from data panel to have 2 data items in total
Click and drag them over the previous workspace 1D data
Actual: It will show the pink overlay that the 1D data in the workspace will merge them to combine into a compound line plot
Expected: the pink overlay should show that the panel will split.

Issue2:
Click and drag multiple into the side of a display panel in the workspace.
Actual: It will show a pink overlay that it will split and add one item
Expected:TBC

Ion-e (changes requested):
^

MattRoyle:
> Verified as failed: Issue 1:

I have fixed the displaying of the composite plot so now it will not show the smaller composite region when you have a selection of multiple items, it will instead just show the main replace with split display.

> Issue2: Click and drag multiple into the side of a display panel in the workspace. Actual: It will show a pink overlay that it will split and add one item Expected:TBC

The split logic and display logic both now follow the expected behavior path. When you drop with multiple items on the side of an existing panel the panel will first be split midway as would happen when dragging a single item to the side. The newly created panel will have the items inserted in the grid panel, leaving the original panel as half of its initial size. See the example images demonstrating this idea below.

<details><summary>Example images</summary>
image

image
</details>

MattRoyle:
I'm moving this back to In Progress while I write the tests.

PR #1890, opened 2026-07-01
Add missing line-plot case for nudge slice left/right to key_config.

### LINKED ISSUE #1888: Control + Arrow keys doesn't change the sequence slice for 1D spectra
It works for 2D rasters.

There needs to be testing happening on how this behaves with a combination of collection ranks and datum ranks.

cmeyer:
Fixes #1888

### DIFF
--- a/nion/swift/resources/key_config.json
+++ b/nion/swift/resources/key_config.json
@@ -75,10 +75,12 @@
         "raster_display": ["Shift+down_arrow", "Shift+M"]
     },
     "raster_display.nudge_slice_left": {
-        "raster_display": ["Ctrl+left_arrow", "move_to_start_of_line"]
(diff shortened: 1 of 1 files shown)


cmeyer:
@Ion-e This is a trivial fix - if you give it the go-ahead, I'll merge. My only concern is if the key overlaps with other functionality, but it seems pretty safe to me.

Ion-e:
Will test, thank you. The only 2 functionalities that it should interact are:
1. Arrow Keys Nudge graphics in corresponding direction.
2. Shift + Arrow Keys Move graphics in corresponding direction.

However if there is an awkward overlapping we should look into it and the behaviour would expected to be consistent between 1-2D display

Ion-e:
Issue verified as fixed.
Looks great, thanks for implemnenting.

PR #1898, opened 2026-07-07
Item delegate refactor

MattRoyle:
This is a work in progress/example of a refactor to Item Delegates as mentioned in https://github.com/nion-software/nion-internal/issues/322

### DIFF
--- a/nion/swift/DataPanel.py
+++ b/nion/swift/DataPanel.py
@@ -338,9 +338,13 @@ class DataPanel(Panel.Panel):
                 elif len(display_items) > 1:
                     mime_data = document_controller.ui.create_mime_data()
                     MimeTypes.mime_data_put_display_items(mime_data, display_items)
-                    anchor_index = self.__selection.anchor_index or 0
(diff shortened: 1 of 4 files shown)


cmeyer:
Should we close this? As discussed, I think this is the right direction, but wrong solution.

MattRoyle:
> Should we close this? As discussed, I think this is the right direction, but wrong solution.

I'll close the issue since the branch has diverged now anyway. Would you be able to put the reason you think this is the wrong approach on the original issue?

cmeyer:
Quick version: the new delegate file contains two specific implementations that differ only slightly for the data and display panels. It also ties the delegate file to _both_ use cases in other files. I'd want to go the other way: a common delegate (probably in the display panel - but needs to be evaluated when implemented) with minimal or no overrides for the data/display panels. Instead the data/display panels should conform to what the delegate needs so that both could be used interchangeably. This is a rough outline, not a worked out plan. I don't think this work is needed at the moment, but discuss with me if you think differently.

PR #1900, opened 2026-07-09
stylesheet for label visibility

cmeyer:
- **Introduce GraphicStylesheet, default stylesheet, and tests.**
- **Integrate GraphicStylesheet into graphic rendering.**
- **Add properties for the text background and visibility.**

This PR introduces a stylesheet system for graphics. It is as close to possible to a CSS style system, but customized for use in this application.

How to review:
- The first commit (introduce stylesheet) stands alone and should be reviewed to see the gist of the system.
- The second commit builds on the first commit and uses it in limited cases in the existing code. No functionality change at this point.
- The third commit is https://github.com/nion-software/nionswift/pull/1883 modified to run on top of stylesheets.

The benefit here is an extendible system of styles that can be used for most properties of most graphics. There is a selector system (CSS) that can select based on graphic type, graphic category, part, some graphic properties, and some user state (focused, selected, etc.). The second commit makes the modifications to `draw` methods and similar to make sure the ability to resolve styles is available to rendering/composer code while still being immutable.

I expect the first two commits to be merged with this PR, but the third commit may be merged as part of #1883 if we decide to go this direction.

This PR implements a portion of
- https://github.com/nion-software/nionswift/issues/1673.

### DIFF
--- a/nion/swift/model/Graphics.py
+++ b/nion/swift/model/Graphics.py
@@ -946,9 +946,14 @@ class Graphic(Persistence.PersistentObject):
             constraints.add("bounds")
         return constraints
 
-    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, is_selected: bool, is_focused: bool) -> None:
+    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, graphic_state: GraphicInteractionState) -> None:
         if not self._closed:
-            self.get_renderer().draw(ctx, ui_settings, mapping, is_selected, is_focused)
+            graphic_attributes = GraphicAttributes(
+                position_locked=self.is_position_locked,
+                shape_locked=self.is_shape_locked,
+                rotation_locked=self.is_rotation_locked,
+            )
+            self.get_renderer().draw(ctx, ui_settings, mapping, graphic_state, graphic_attributes)
 
     def get_renderer(self) -> GraphicRenderer:
         raise NotImplementedError()
@@ -1265,26 +1270,53 @@ class GraphicRenderer:
     """
 
     def __init__(self, graphic: Graphic) -> None:
+        self.graphic_type = graphic.type
         self.graphic_uuid = graphic.uuid
         self.modified_count = graphic.modified_count
         self.stroke_color = graphic.stroke_color
-        self.used_stroke_style = graphic.used_stroke_style
-        self.used_stroke_width = graphic.used_stroke_width
-        self.used_fill_style = graphic.used_fill_style
+        self.stroke_width = graphic.stroke_width
+        self.fill_color = graphic.fill_color
+        self.used_role = graphic.used_role
+        self.is_shape_locked = graphic.is_shape_locked
+        self.is_position_locked = graphic.is_position_locked
+        self.is_rotation_locked = graphic.is_rotation_locked
         self.label = graphic.label
         self.label_padding = graphic.label_padding
         self.label_font = graphic.label_font
-        self.is_position_locked = graphic.is_position_locked
-        self.is_shape_locked = graphic.is_shape_locked
-        self.is_rotation_locked = graphic.is_rotation_locked
+        self.graphic_stylesheet = GraphicStylesheet.default()
+        default_interaction_state = GraphicInteractionState.default()
+        default_graphic_attributes = GraphicAttributes(
+            position_locked=self.is_position_locked,
+            shape_locked=self.is_shape_locked,
+            rotation_locked=self.is_rotation_locked,
+        )
+        # Keep these values for legacy call sites that do not pass style context.
+        self.used_stroke_style = self.resolve_shape_stroke_color(default_interaction_state, default_graphic_attributes)
+        self.used_stroke_width = self.resolve_shape_stroke_width(default_interaction_state, default_graphic_attributes)
+        self.used_fill_style = self.resolve_shape_fill_color(default_interaction_state, default_graphic_attributes)
 
-    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, is_selected: bool, is_focused: bool) -> None:
+    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> None:
         raise NotImplementedError()
 
     def label_position(self, mapping: CoordinateMappingLike, font_metrics: UISettings.FontMetrics, padding: float) -> typing.Optional[Geometry.FloatPoint]:
         return None
 
-    def draw_label(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike) -> None:
+    def resolve_shape_stroke_color(self, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> str:
+        if self.stroke_color is not None:
+            return self.stroke_color
+        return self.graphic_stylesheet.resolve(self.graphic_type, self.used_role, {"part": "shape"}, PROP_STROKE_COLOR, graphic_state, graphic_attributes)
+
+    def resolve_shape_fill_color(self, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> str:
+        if self.fill_color is not None:
+            return self.fill_color
+        return self.graphic_stylesheet.resolve(self.graphic_type, self.used_role, {"part": "shape"}, PROP_FILL_COLOR, graphic_state, graphic_attributes)
+
+    def resolve_shape_stroke_width(self, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> float:
+        if self.stroke_width is not None:
+            return self.stroke_width
+        return self.graphic_stylesheet.resolve(self.graphic_type, self.used_role, {"part": "shape"}, PROP_STROKE_WIDTH, graphic_state, graphic_attributes)
+
+    def draw_label(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> None:
         label = self.label
         if label:
             padding = self.label_padding
@@ -1305,7 +1337,7 @@ class GraphicRenderer:
                     ctx.close_path()
                     ctx.fill_style = "rgba(255, 255, 255, 0.6)"
                     ctx.fill()
-                    ctx.stroke_style = self.used_stroke_style
+                    ctx.stroke_style = self.resolve_shape_stroke_color(graphic_state, graphic_attributes)
                     ctx.stroke()
                     ctx.font = font
                     ctx.text_baseline = "middle"
@@ -1328,7 +1360,7 @@ class MissingGraphic(Graphic):
 
 class MissingGraphicRenderer(GraphicRenderer):
 
-    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, is_selected: bool, is_focused: bool) -> None:
+    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> None:
         pass
 
 
@@ -1557,15 +1589,17 @@ class RectangleGraphic(RectangleTypeGraphic):
 
 class RectangleGraphicRenderer(RectangleTypeGraphicRenderer):
 
-    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, is_selected: bool, is_focused: bool) -> None:
+    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> None:
         bounds = self.bounds
         rotation = self.rotation
-        used_stroke_width = self.used_stroke_width
-        used_stroke_style = self.used_stroke_style
-        used_fill_style = self.used_fill_style
-        is_shape_locked = self.is_shape_locked
-        is_position_locked = self.is_position_locked
-        is_rotation_locked = self.is_rotation_locked
+        used_stroke_width = self.resolve_shape_stroke_width(graphic_state, graphic_attributes)
+        used_stroke_style = self.resolve_shape_stroke_color(graphic_state, graphic_attributes)
+        used_fill_style = self.resolve_shape_fill_color(graphic_state, graphic_attributes)
+        is_selected = graphic_state.selected
+        is_focused = graphic_state.focused
+        is_shape_locked = graphic_attributes.shape_locked
+        is_position_locked = graphic_attributes.position_locked
+        is_rotation_locked = graphic_attributes.rotation_locked
         # origin is top left
         origin = mapping.map_point_image_norm_to_widget(bounds.origin)
         size = mapping.map_size_image_norm_to_widget(bounds.size)
@@ -1608,7 +1642,7 @@ class RectangleGraphicRenderer(RectangleTypeGraphicRenderer):
                 ctx.stroke_style = used_stroke_style
                 ctx.stroke()
                 draw_circular_marker(ctx, rotation_point, is_focused, is_rotation_locked)
-        self.draw_label(ctx, ui_settings, mapping)
+        self.draw_label(ctx, ui_settings, mapping, graphic_state, graphic_attributes)
 
     def label_position(self, mapping: CoordinateMappingLike, font_metrics: UISettings.FontMetrics, padding: float) -> typing.Optional[Geometry.FloatPoint]:
         bounds = Geometry.FloatRect.make(self.bounds)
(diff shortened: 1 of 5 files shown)


nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+        attributes: GraphicAttributes,
+    ) -> T:
+        """Resolve *key* against a selector fragment.
+
+        Axes inside the fragment are unordered constraints.
+        """
+        result = key.default
+        fragment_part = selector_fragment.get("part")
+        fragment_axes = {k: v for k, v in selector_fragment.items() if k != "part" and v is not None}
+        for rule in self._rules:
```
Tiomat85, line 1198: This is going to scale badly as it grows, you have O(N) here as you are checking all rules for each check.  A general list of rules seems to be quite antithetical to the premise of CSS of repeated placement of hierarchical partial style lists.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+          graphic[role="fourier_mask"]::part(shape)                      { stroke-color: #F08; fill-color: rgba(255,0,127,0.1) }
+        """
+        rules: list[GraphicStyleRule] = []
+
+        # base default: all graphics, all parts are always visible
+        base_properties = GraphicStyleProperties()
+        rules.append(GraphicStyleRule(GraphicStyleSelector(), base_properties, order=1))
+
+        # base stroke/width defaults (matches Graphic._default_stroke_color and used_stroke_width)
+        base_shape_properties = GraphicStyleProperties()
```
Tiomat85, line 1222: This list of rules method is incredibly brittle, as this grows this will become unmanageable and the styling is entirely dependent on getting the ordering here completely correct.  If the order is wrong, you will get in a tangle with incorrect styling and have to manually untangle/reorder to get the 'rules' applied in the right order.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+
+        Rule mapping (reproduces existing hard-coded logic, last match wins):
+          graphic                                                        { visibility: always }
+          graphic                                                        { stroke-color: #F80; stroke-width: 1.0 }
+          interval-graphic::part(shape)                                  { stroke-color: #F00 }
+          channel-graphic::part(shape)                                   { stroke-color: #F00 }
+          graphic[role="mask"]::part(shape)                              { stroke-color: #00F; fill-color: rgba(0,0,255,0.1) }
+          graphic[role="fourier_mask"]::part(shape)                      { stroke-color: #F08; fill-color: rgba(255,0,127,0.1) }
+        """
+        rules: list[GraphicStyleRule] = []
```
Tiomat85, line 1215: Why is this hard-coded imperative implementations rather than a style-sheet like the discussed objective where you completely separate out the actual styling from the code that uses that.  These defaults should be in an explicit defaults.nss or similar file completely distinct from the implementation of how those defaults are used.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1006,301 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+        return self.state_selector.matches(state) and self.attributes_selector.matches(attributes)
+
+
+class GraphicStyleRule:
+    """A selector paired with a declaration (set of properties) and a source order."""
+
+    def __init__(self, selector: GraphicStyleSelector, declaration: GraphicStyleProperties, order: int = 0) -> None:
+        self.selector = selector
+        self.declaration = declaration
+        self.order = order
```
MattRoyle, line 1169: The resolver does not ever use the order field so it seemingly does nothing.

nion/swift/model/Graphics.py
```diff
@@ -1265,26 +1270,53 @@ class GraphicRenderer:
...
+        self.is_rotation_locked = graphic.is_rotation_locked
         self.label = graphic.label
         self.label_padding = graphic.label_padding
         self.label_font = graphic.label_font
-        self.is_position_locked = graphic.is_position_locked
-        self.is_shape_locked = graphic.is_shape_locked
-        self.is_rotation_locked = graphic.is_rotation_locked
-
-    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, is_selected: bool, is_focused: bool) -> None:
+        self.graphic_stylesheet = GraphicStylesheet.default()
```
Tiomat85, line 1286: Why are we mixing and matching approaches here?  If we are trying to use a stylesheet way of storing display properties, then that should propagate through entirely rather than having local imperative properties as well as a default stylesheet.
If we want to go ahead with a stylesheet approach then we should move all storage and querying of display properties into that approach, and if the architecture for the stylesheets doesn't support that then it should be adapted so we can use a singular way of handling this.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
         return False
 
 
+T = typing.TypeVar('T')
+
+
+class GraphicInteractionState:
```
Tiomat85, line 1007: Why are these graphic specific?  Surely the stylesheets should be applicable to all UI elements, and selected/focussing are common to most elements not just Graphics objects.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+    def __init__(self, selected: bool | None = None, focused: bool | None = None) -> None:
+        self.selected = selected
+        self.focused = focused
+
+    @staticmethod
+    def wildcard() -> GraphicInteractionStateSelector:
+        """Return a selector with no interaction-state constraints."""
+        return GraphicInteractionStateSelector()
+
+    def matches(self, graphic_state: GraphicInteractionState) -> bool:
```
Tiomat85, line 1049: Is this going to get unwieldy if we start adding additional interaction state? Hover comes to mind immediately.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+        self.position_locked = position_locked
+        self.shape_locked = shape_locked
+        self.rotation_locked = rotation_locked
+
+    @staticmethod
+    def wildcard() -> GraphicAttributesSelector:
+        """Return a selector with no attribute constraints."""
+        return GraphicAttributesSelector()
+
+    def matches(self, graphic_attributes: GraphicAttributes) -> bool:
```
Tiomat85, line 1074: Ditto unwieldyness if we start adding additional attributes.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+    def matches(self, graphic_state: GraphicInteractionState) -> bool:
+        """Return True if this selector matches the given interaction state."""
+        if self.selected is not None and self.selected != graphic_state.selected:
+            return False
+        if self.focused is not None and self.focused != graphic_state.focused:
+            return False
+        return True
+
+
+class GraphicAttributesSelector:
```
Tiomat85, line 1058: Rather than 1 class to handle all attribute selection, would a sequence of selectors make more sense?

So if you had one that applies if position_locked = True and shape_locked = False,  you pass in something akin to:
[PositionLockedAttributeSelector(True), ShapeLockedAttributeSelector(False)] and it cycles through those conditions.  Essentially turning it into a filter.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+        self.__storage[key.name] = value
+
+
+# Built-in property keys. Stylesheet identifiers use dashes; Python names use underscores.
+PROP_STROKE_COLOR: GraphicStylePropertyKey[str] = GraphicStylePropertyKey("stroke-color", str, "#F80")
+PROP_FILL_COLOR: GraphicStylePropertyKey[str] = GraphicStylePropertyKey("fill-color", str, "")  # empty string = no fill
+PROP_STROKE_WIDTH: GraphicStylePropertyKey[float] = GraphicStylePropertyKey("stroke-width", float, 1.0)
+
+
+class GraphicStyleSelector:
```
Tiomat85, line 1119: Would it make sense to consider how this applies to the full UI styling rather than just focussing on the Graphics? It makes it a larger scoped project, but given we agreed this was going to be a longer term project anyway we might as well decide on the end-goal specifications rather than just the short-term ones for a specific issue.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1001,254 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+        part: str | None = None,
+        axes: dict[str, str] | None = None,
+        state_selector: GraphicInteractionStateSelector | None = None,
+        attributes_selector: GraphicAttributesSelector | None = None,
+    ) -> None:
+        self.graphic_type = graphic_type
+        self.role = role
+        self.part = part
+        self.axes = axes if axes is not None else dict()
+        self.state_selector = state_selector if state_selector is not None else GraphicInteractionStateSelector.wildcard()
```
Tiomat85, line 1138: Why do we need these wildcards, when all it turns into is 'True'?

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1006,301 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+        for rule in self._rules:
+            if rule.selector.matches(graphic_type, role, fragment_part, state, attributes, fragment_axes) and rule.declaration.has(key):
+                result = rule.declaration.get(key)
+        return result
+
+    @staticmethod
+    def default() -> GraphicStylesheet:
+        """Return a stylesheet that encodes the current built-in defaults as explicit rules.
+
+        Rule mapping (reproduces existing hard-coded logic, last match wins):
```
MattRoyle, line 1214: > last match wins

Intentional or not this is not how CSS works. Without a specification to go off of to define how this style-sheet should work it is hard to repute fundamental design decisions however this approach has drawbacks. 
CSS cascade does use order precedence as a way to determine the value, but that is the last step used in the cascade algorithm only used when all the above ones don't resolve. The most useful part of CSS is the specificity which it seems is is attempted with the type role, part etc but is not prioritized over the list order so if you defined a more generic rule after more specific rules the generic one would override the property.
See [Cascade order](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Introduction#cascading_order).

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1006,301 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+        self.graphic_type = graphic_type
+        self.role = role
+        self.part = part
+        self.axes = axes if axes is not None else dict()
+        self.state_selector = state_selector if state_selector is not None else GraphicInteractionStateSelector.wildcard()
+        self.attributes_selector = attributes_selector if attributes_selector is not None else GraphicAttributesSelector.wildcard()
+
+    def matches(self, graphic_type: str, role: str | None, part: str | None, state: GraphicInteractionState,
+                attributes: GraphicAttributes, axes: dict[str, str] | None = None) -> bool:
+        if self.graphic_type is not None and self.graphic_type != graphic_type:
```
MattRoyle, line 1150: This appears to be attempting to implement the specificity of a cascade algorithm but the inclusion of role, part, and axes is makes this very inflexible. What if more specificity is needed, then another will have to be added here which could effect every single graphic property. Some of these are also overspecific to graphics like an axis which limits the ability for this to be expanded to other parts of the software that don't have an axis.
CSS specificity works on ID, Class and Type with each having a weight added for the levels. Then each selector targeting an element will be have the weights compared with ID priority, then if the ID weights of the selector is the same then the Class is checked in the same way and finally the Type. See [Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity#selector_weight_categories)

nion/swift/model/Graphics.py
```diff
@@ -1017,33 +1317,88 @@ class GraphicRenderer:
...
+        self.is_rotation_locked = graphic.is_rotation_locked
         self.label = graphic.label
         self.label_padding = graphic.label_padding
         self.label_font = graphic.label_font
-        self.is_position_locked = graphic.is_position_locked
-        self.is_shape_locked = graphic.is_shape_locked
-        self.is_rotation_locked = graphic.is_rotation_locked
-
-    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, is_selected: bool, is_focused: bool) -> None:
+        self.graphic_stylesheet = GraphicStylesheet.default()
```
MattRoyle, line 1333: This has a negative performance scaling. With each additional rule you want to add to GraphicStylesheet you are increasing the number of rules checked per resolve call since every rule is checked in the resolve call. Each resolve call is done for every graphic when redrawn. Additionally whenever a new graphics renderer is instantiated a whole new style sheet is also created by the default() call even though they are all identical and should probably be a global level object. Intervals on live line plot displays like the ZLP create a new GraphicRenderer every frame, so this class will be initialized every frame for every interval graphic on every live line plot. I'm not sure if the same occurs with 2D graphics.

nion/swift/model/Graphics.py
```diff
@@ -1001,6 +1006,301 @@ def test_label(self, ui_settings: UISettings.UISettings, mapping: CoordinateMapp
...
+                role="measurement",
+                part="label",
+                axes={"category": "details", "type": "width"},
+            ),
+            measurement_properties,
+            order=4,
+        ))
+
+        # interval-graphic shape default stroke color
+        measurement_interval_shape_properties = GraphicStyleProperties()
```
MattRoyle, line 1292: This is just duplicated. It highlights the drawback of how unreadable this imperative approach is.

MattRoyle (commented):
I think there is a need for a proper specification, this is very far from what I understood from the discussion. This is very specific to the Graphics and tied in to how that currently works. Even with just five properties built in the GraphicsStylesheet is hard to follow exactly what the outcome will be to such a degree that the docstring includes a css style declarative version, that is what I imagined adding a property would look like, the parsing of which would be done once.

nion/swift/model/LinePlotDisplay.py
```diff
@@ -25,14 +25,18 @@
...
     channels: typing.Tuple[float, float]
-    selected: bool
     index: int
     left_text: str
     right_text: str
     middle_text: str
     label: typing.Optional[str]
     style: typing.Optional[str]
     color: typing.Optional[str]
+    graphic_stylesheet: Graphics.GraphicStylesheet
```
MattRoyle, line 35: From @cmeyer in reference to #1883:
> RegionInfo is a data only class, so adding a function callback is a sneaky way of changing its intrinsic nature. So as written, I would reject this PR on that point only.

Here the class graphic_stylesheet is added which has the resolve method which is used *as a callback* to get the state of specific properties:
https://github.com/nion-software/nionswift/pull/1900/changes#diff-fc43e746fd3782b16ec910753de2c1975a18d9ada87e062f480b68fc50f6680dR668

More info in: https://github.com/nion-software/nionswift/pull/1883/changes#r3551311953

cmeyer: The callbacks in #1883 point to functions (which should be properties, btw) on the `Graphic`. The `Graphic` is mutable and may change between calls, leading to race conditions. In the stylesheet implementation, there shouldn't be any callbacks. There is an immutable stylesheet object that gives the capability to resolve the value of a property based on rules for a given rendering pass. If you think there is a callback (reference to Graphic or something else immutable) that I missed, please give me the line in the PR version of the code so I can look at it. I would consider that to be a bug.

nion/swift/model/Graphics.py
cmeyer: We might want to cache the default stylesheet and we might be able to streamline the code, but I don't see how this makes anything unreadable - it's expressing complicated conditions of styling in a single location.

Also, what do you mean by "imperative approach"?

nion/swift/model/Graphics.py
cmeyer: Working code first. Optimization second as long as the initial approach isn't too bad. I don't think this is a significant performance problem. Caching the default stylesheet and snapshotting when needed is a trivial improvement, but I don't think it has any measurable effect in this version.

nion/swift/model/Graphics.py
cmeyer: I debated the specific implementation of this and it's arguable to integrate 'role' and 'part' into 'axes' (<-- an unfortunate conflicting name that I'll update) - but 'axes' serves to allow unlimited expansion in specificity.

nion/swift/model/Graphics.py
cmeyer: I'll provide the design doc in the next revision. I'll also update that comment. "the cascade decides first; source order decides last" or something similar.

cmeyer:
> I think there is a need for a proper specification, this is very far from what I understood from the discussion. This is very specific to the Graphics and tied in to how that currently works. Even with just five properties built in the GraphicsStylesheet is hard to follow exactly what the outcome will be to such a degree that the docstring includes a css style declarative version, that is what I imagined adding a property would look like, the parsing of which would be done once.

This is a fair assessment - but the advantage of this, even though it's more complicated from a code perspective, is that it allows us to centralize the style logic and make the logic declarative.

Right now I coded the stylesheet rather than adding a parser (coding probably being what you mean by 'imperative') but I think that's a good stepping stone. I included the text equivalent to understand where we might go with it.

I could easily be convinced to separate the first commit into its own file, eliminate anything graphics-specific, and add a parser. I decided to focus on graphics first to work out the details.

nion/swift/model/Graphics.py
```diff
@@ -1017,26 +1270,53 @@ class GraphicRenderer:
...
+        self.used_fill_style = self.resolve_shape_fill_color(default_interaction_state, default_graphic_attributes)
+
+    def draw(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> None:
         raise NotImplementedError()
 
     def label_position(self, mapping: CoordinateMappingLike, font_metrics: UISettings.FontMetrics, padding: float) -> typing.Optional[Geometry.FloatPoint]:
         return None
 
-    def draw_label(self, ctx: DrawingContextLike, ui_settings: UISettings.UISettings, mapping: CoordinateMappingLike) -> None:
+    def resolve_shape_stroke_color(self, graphic_state: GraphicInteractionState, graphic_attributes: GraphicAttributes) -> str:
```
Tiomat85, line 1304: This feels a little odd as it is mixing the renderer directly having a colour assigned, or asking the styling system for the colour as opposed to just having one of those systems handling it.  This means lots of boiler-plate functions like this that have to check a specific local property before then going to the styler to get what it should be when instead we should be able to encapsulate all of the styling determination elsewhere and re-use without needing all of these specific functions.

Tiomat85 (changes requested):
I think before we go down this route we should probably have a sit down and try to determine what the end-goals are with this.  My understanding of what we were trying to achieve was to have styling completely distinct from the ui code, with a common re-usable style handler that can handle the various aspects of modern UI styling (inheritence, composition, state) that makes up the cascading part of CSS.  Additionally we want to think of what we want the other end to look like, to make sure that we don't end up with multiple solutions that control half of the styling each which just leads to increased difficulty in future when my view would be that a single UI element can essentially query a StyleManager with it's own current state to get individual stylistic properties.

Since this is going to be such a big change and refactor, we need to make sure that we get what is most useful for future development fully in place and designed so that we don't need to change course on it too much, and plan how a full migration of all Styling of the UI is going to work with the new system so we don't end up with multiple parallel, incompatible systems.

cmeyer:
Close pending further discussion.

PR #1903, opened 2026-07-10
Fix issue with data not updating when internally modified.

### LINKED ISSUE #1902: Data updates are displayed intermittently
Make a 16x16 image. Assign it r01. Run the script below, which updates all pixels in the image. Intermittently, the image will only partially update.

```
import random

def c():
    data_item = r01.data_item._data_item
    with data_item.data_ref() as data_ref:
        data = data_ref.data
        data[0, 0] = random.random()
        data_ref.data = data
    with data_item.data_ref() as data_ref:
        data = data_ref.data
        data[0, 1] = random.random()
        data_ref.data = data

c()
```

```
import random
import time

def a(data_item, x):
    with _model.item_transaction(data_item):
        with data_item.data_ref() as data_ref:
            data = data_ref.data
            for index in range(data.shape[0]):
                data[index, x] = random.random()
            data_ref.data = data

def b():
    data_item = r01.data_item._data_item
    for x in range(data_item.data_shape[0]):
        a(data_item, x)

b()
```

cmeyer:
Fixes #1902

### DIFF
--- a/nion/swift/model/DisplayItem.py
+++ b/nion/swift/model/DisplayItem.py
@@ -1838,15 +1838,21 @@ class DisplayPropertiesLayersGraphics:
 
 @dataclasses.dataclass(frozen=True)
 class DisplayDataChannelsAndCalibrationStyle:
+    # must be non-equal when data changes, so data modified counts are included.
+    # equality is used to check for stream changes.
     display_data_channels: typing.Sequence[DisplayDataChannel | None]
     calibration_style_id: str | None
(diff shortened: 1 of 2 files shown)


cmeyer:
Acknowledging that I'm doing this in the wrong order - I merged this so I could build candidate installers this weekend. @KRLango can you review anyway after the fact?

KRLango:
Code looks good to me. I have also ran the same tests that originally showed this bug and they are now working correctly.

PR #1909, opened 2026-07-14
Fix interval drawing regression (l/r/mid appearing even when deselected).

cmeyer:


cmeyer:
I merged this directly for internal testing. @Tiomat85 I'd still like a review for a double check. Thanks.

Tiomat85 (commented):
Code looks fine, the rejigging makes it clearer.

PR #1915, opened 2026-07-16
storage system cleanup

### LINKED ISSUE #1914: Persistent object system does not correctly store properties internally
In process data migration takes on-disk format and converts it to in-memory format by transforming on-disk to in-memory at read and un-transforming in-memory to on-disk at write.

However, the properties are sometimes stored un-transformed in the persistent objects themselves and not properly un-transformed at write. This works because they just get re-transformed at read again; but it is not possible to use the in-memory format for anything useful since the actual format is not known.

Proper in-memory format is required for upcoming computation work.

cmeyer:
Fixes #1914

- **Eliminate explicit support for large_format flags. Part of a storage refactoring.**
- **Fix storage issue to store transformed properties in dict.**

---

- Eliminates the 'properties' property on data and display items. Callers must now explicitly use persistent_storage.get_item_properties(). This makes the code more clear.
-  Eliminates the poorly defined `PersistentStorageSystem.get_storage_properties` and replaces it with `get_properties`.
- Ensures the `__properties` stored in `PersistentStorageSystem` are the transformed properties, which they were not before. This ensures the 'in memory' state matches the current version of the class.
- Explicitly labels the reader/writer methods to be clear that they are reading/writing untransformed data.
- Renames `read_project_properties` to `get_combined_properties` for clarity.
- Add transform/untransform virtual methods to be used while reading and writing.
- Split the normalization part of `load_properties` into `normalize_project_data_path` and add a test for that code path.
- Other minor cleanup and test updates.

See also:
- [persistent object model design document](https://github.com/nion-software/nionswift/blob/master/design/persistent_object_model.md)

### DIFF
--- a/nion/swift/model/DataItem.py
+++ b/nion/swift/model/DataItem.py
@@ -677,11 +677,6 @@ class DataItem(Persistence.PersistentObject):
             self.dynamic_title = dynamic_string
 
     @property
-    def properties(self) -> typing.Optional[Persistence.PersistentDictType]:
-        """ Used for debugging. """
-        return self.get_storage_properties()
-
-    @property
     def is_live(self) -> bool:
         """Return whether this library item represents live acquisition."""
         return self.__is_live
--- a/nion/swift/model/DataItem.py
+++ b/nion/swift/model/DataItem.py
@@ -184,11 +184,9 @@ class DataItem(Persistence.PersistentObject):
     storage_version = 13
     writer_version = 13
 
-    def __init__(self, data: typing.Optional[_ImageDataType] = None, item_uuid: typing.Optional[uuid.UUID] = None,
-                 large_format: bool = False) -> None:
+    def __init__(self, data: typing.Optional[_ImageDataType] = None, item_uuid: typing.Optional[uuid.UUID] = None, large_format: bool = False) -> None:
         super().__init__()
         self.uuid = item_uuid if item_uuid else self.uuid
-        self.large_format = large_format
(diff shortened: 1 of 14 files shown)


nion/swift/model/DataItem.py
```diff
@@ -184,11 +184,9 @@ class DataItem(Persistence.PersistentObject):
     storage_version = 13
     writer_version = 13
 
-    def __init__(self, data: typing.Optional[_ImageDataType] = None, item_uuid: typing.Optional[uuid.UUID] = None,
-                 large_format: bool = False) -> None:
+    def __init__(self, data: typing.Optional[_ImageDataType] = None, item_uuid: typing.Optional[uuid.UUID] = None, large_format: bool = False) -> None:
```
Tiomat85, line 187: Can we remove large_format from the initializer here, it is now unused.

Tiomat85 (approved):
LGTM

nion/swift/model/DataItem.py
cmeyer: I don't think so. We can't change all caller sites that might use this, so I think it's safer to leave it in there.

Tiomat85: Ok, makes me twitch but it's a minor thing not worth changing a hundred caller sites currently.

cmeyer: I changed a lot of the internal uses of this - but even though we don't publish this API, it's one of the most likely to be used externally. I will have follow-ups for the remaining internal uses after the computation branch merges, which itself will take care of a bunch of them.

PR #1918, opened 2026-07-16
Use annotated array name, define coordinate mappings, update iteration section.

cmeyer:
This is a design document PR.

The best way to review it is to [look at the file as a whole](https://github.com/cmeyer/nionswift/blob/bb54dc19e98243a6933688407cb4fe166de7a3c2/design/data_model.md), not the diff.

The intention here is to provide guidance for upcoming computation, calibrations, and coordinate system work, not necessarily to set things permanently.

~~Open question: do 'coordinate system transformations' belong in this document?~~ Using 'extensions' to handle this, see design document.

### DIFF
--- a/design/data_model.md
+++ b/design/data_model.md
@@ -16,51 +16,37 @@ Each axis has a single calibration associated with it. The calibration may be an
 
 ### Planned Implementation
 
-The planned annotated array model specifies an ordered list of axis groups, each with a rank, and a value type.
+An annotated array contains an underlying array, array descriptor, and array metadata. An array header packages the descriptor and metadata with the storage data type.
 
-Value types are scalar, complex, RGB, RGBA, or fixed-length vector (1/2/3).
+The array metadata contains a creation time, free-form attributes, and structured extension records. An extension record is a versioned, typed container for an opaque encoded payload whose semantics and access interface are defined outside the annotated array model. Extension records track things like computation provenance and coordinate transforms.
 
-An annotated array always contains at least one axis group. All axis groups must have at least one axis except the final axis group may have zero.
+The array descriptor contains an ordered list of axis groups and a value type. At least one axis group is required; only the final group may have rank zero.
 
-Coordinate systems are physical coordinate spaces typically representing an orientation of axes in space.
-
-Coordinate mappings are mappings from the index coordinate space of an axis group into calibrated coordinates.
-
-Each axis group has zero or more associated coordinate mappings. One coordinate mapping may be designated the primary coordinate mapping.
-
-Each axis group designates the coordinate system of its index coordinate space. The designated coordinate system may be null, indicating that the coordinate system is unknown.
-
-Axis groups in different annotated arrays are correlated when they reference the same coordinate system.
-
-Within a coordinate mapping, each axis has exactly one associated calibration or coordinate array.
-
-An annotated array contains an underlying array, array descriptor, and array metadata.
-
-The array descriptor contains the intrinsic fields used to interpret the array.
-
-The array metadata contains a creation time, extension records, and free-form metadata.
-
-An array header packages an array descriptor, storage data type, and array metadata.
-
-Future revisions may expand the value types and/or support variable length (ragged) value types.
+Each axis group may have coordinate mappings and may reference the coordinate system in which its mapped coordinates are interpreted. One coordinate mapping may be primary.
 
 ### Terms and Definitions
 
 An **annotated array** is one complete unit of data: an underlying array, array descriptor, and array metadata.
 
-An **array header** is a utility value containing an array descriptor, storage data type, and array metadata without an underlying array.
-
 An **array descriptor** contains the intrinsic fields required to interpret an underlying array.
+
+**Array metadata** is contextual information that travels with an annotated array but is not required to interpret its underlying array.
+
+An **array header** is a utility value containing an array descriptor, storage data type, and array metadata without an underlying array.
 
 An **intrinsic field** is information whose semantics and invariants are part of the annotated array model, such as axis groups, coordinate mappings, coordinate system references, and intensity calibrations.
 
-**Array metadata** is contextual information that travels with an annotated array but is not required to interpret its underlying array.
+The **storage data type** specifies how values are represented in the underlying array. It is distinct from the value type, which describes the meaning of one logical value.
+
+The **creation time** is the time at which an annotated-array value is created.
 
 An **extension record** is versioned, code-interpreted array metadata whose semantics are owned outside the fundamental annotated array model. Examples are acquisition coordinate context and computation provenance.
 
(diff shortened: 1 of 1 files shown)


design/data_model.md
```diff
@@ -1,81 +1,99 @@
...
+Within a coordinate mapping, each axis has exactly one associated calibration or coordinate array.
+
+The annotated array has zero or more coordinate system transforms, which are affine transforms from one coordinate system to another, recorded at acquisition time. They normally only apply to isotropic coordinate mappings.
 
 Future extensions may expand the value types and/or support variable length (ragged) value types.
 
-### Glossary
+### Terms and Definitions
+
+An **annotated array** is one complete unit of data: an underlying array together with its annotations (axis groups, coordinate mappings, coordinate system transforms, and metadata). The implementation class name is `AnnotatedArray`. *Note on nomenclature:* "annotations" here means the structural and calibration metadata attached to the array - not display graphics such as rectangles, lines, or region markers drawn on a displayed image.
```
Tiomat85: Quite a bit of overlap here with the section above, I don't know how tight we are trying to be with tokens for these files.

design/data_model.md
```diff
@@ -1,81 +1,99 @@
...
+
+An **axis group** is an ordered group of axes in an annotated array. The rank of an axis group is the number of axes in the group.
+
+The **signal axis group** is the final axis group in an annotated array. *Note on nomenclature:* "signal" is not a perfect term - axes outside the signal axis group may also carry meaningful signal - but it is the least bad of the candidates considered and a clear improvement over the previous term "datum," which read as surveying jargon and was an awkward singular of "data."
+
+A **value type** is one of the supported value types: scalar, complex, RGB, RGBA, or vector.
+
+A **coordinate system** is a physical coordinate space, typically representing an orientation of axes in space, consisting of an ordered list of coordinate system axes; it is the node type of the coordinate system graph. Each coordinate system has a unique identifier; identity is by identifier, so two coordinate systems with identical coordinate system axes are still distinct systems. Coordinate systems are shared: axis groups in one or more annotated arrays may designate the same coordinate system.
```
Tiomat85: This one is quite difficult to read.  Asking Copilot to reword it I got:

A coordinate system represents a physical coordinate space, typically defined by an orientation of axes, and contains an ordered list of coordinate axes. Each coordinate system has a unique identifier that solely determines its identity; systems with identical axes are distinct if their identifiers differ. Coordinate systems are shared objects and may be referenced by axis groups in multiple annotated arrays.

Tiomat85 (approved):

design/data_model.md
cmeyer: Good catch - the "node type" thing has to go. That was an LLM thing that got inserted at one point.

cmeyer:
I pushed more changes - notably to simplify the classes, but at the same time add a 'extension record' capability, which is metadata but managed by an extension class. e.g. computation provenance, acquisition info, or microscope state such as coordinate transforms.

bentps:
The key change I'm seeing here is:

"Each axis group has zero or more associated coordinate mappings. One coordinate mapping may be designated the primary coordinate mapping."

Yes, great. More than one coordinate mapping is great.

Are there other new use cases that this document is designed to address? Do you think you could write a paragraph about what problems you are trying to solve here? I might have more comments if I had that.

"Graphics are drawn in graphics coordinates — a fixed per-axis scaling of index coordinate space in which each axis spans 0 to 1 — independent of any coordinate mapping. A graphic may instead be defined in stage coordinates: its points are converted to fractional indices through inverse coordinate mapping. "

How do we decide if a graphic is defined by graphics coordinates or in a coordinate mapping? Can we make it so graphics are always defined in the active coordinate mapping, and if the graphic is copied to another data item with compatible coordinates, the calibrated coordinates of the graphic are preserved, rather than the "graphics coordinates"?

This document touches on some things which are concerns of mine, but I don't know if this is the right place to bring them up, like the flexibility of our data structure. I don't see a reason in principle to restrict the number of data or collection axes.

For viewing data in Swift, it would be great if we could make the axis indexing match python conventions, so you can translate a square graphic location coordinates in pixels in the inspector directly to array slice indices in python.

cmeyer:
> The key change I'm seeing here is:
> 
> "Each axis group has zero or more associated coordinate mappings. One coordinate mapping may be designated the primary coordinate mapping."
> 
> Yes, great. More than one coordinate mapping is great.
> 
> Are there other new use cases that this document is designed to address? Do you think you could write a paragraph about what problems you are trying to solve here? I might have more comments if I had that.

I added an introductory sentence:

The planned model expands the current representation so it can support more axis groups, multiple calibrations per axis group, and extensions such as provenance and coordinate transformations without reshaping the core structure. It also leaves room for richer coordinate-system references, non-uniform coordinates, and additional value types as the model evolves.

> 
> "Graphics are drawn in graphics coordinates — a fixed per-axis scaling of index coordinate space in which each axis spans 0 to 1 — independent of any coordinate mapping. A graphic may instead be defined in stage coordinates: its points are converted to fractional indices through inverse coordinate mapping. "
> 
> How do we decide if a graphic is defined by graphics coordinates or in a coordinate mapping? Can we make it so graphics are always defined in the active coordinate mapping, and if the graphic is copied to another data item with compatible coordinates, the calibrated coordinates of the graphic are preserved, rather than the "graphics coordinates"?

This is noted and appears in a separate graphics design document I will publish soon.

> 
> This document touches on some things which are concerns of mine, but I don't know if this is the right place to bring them up, like the flexibility of our data structure. I don't see a reason in principle to restrict the number of data or collection axes.
> 

This is part of the purpose of the redesign: no restriction on the data or collect axes.

> For viewing data in Swift, it would be great if we could make the axis indexing match python conventions, so you can translate a square graphic location coordinates in pixels in the inspector directly to array slice indices in python.

I will follow up on this separately.

Please feel free to post any other questions here and I'll either file them into separate issues, take them offline, or add them to the design document.

cmeyer:
Squashed and merged. 2d6cb124af0e0723f4bbdf18bfba071c5cd40672

tylerharvey:
Looks like I'm late here, but I'll add:
> A calibration is a structure containing scale, offset, and units used to map index values to calibrated values using the formula x' = x * scale + offset. A calibration is a linear transform on a single axis. A calibration exists within a coordinate mapping, one per axis. Calibrations support sampled (uniformly spaced) axes.
>
> A coordinate array is a structure containing arrays that map index values to coordinate values. Coordinate arrays support non-sampled (non-uniformly spaced) axes.

Log-scale coordinate axes should be supported as well as linear-scale axes rather than needing to be implemented as an array. This is the way matplotlib does it. The array approach is extremely clunky for something that can be fully represented by a scale, an offset, a unit, and `scale_type='log'` rather than a giant floating point array.

PR #1921, opened 2026-07-22
Data model design: introduce persistent format.

cmeyer:
Add persistent storage format description.

### DIFF
--- a/design/data_model.md
+++ b/design/data_model.md
@@ -24,7 +24,7 @@ The array metadata contains a creation time, free-form attributes, and structure
 
 The array descriptor contains an ordered list of axis groups and a value type. At least one axis group is required; only the final group may have rank zero.
 
-Each axis group may have coordinate mappings and may reference the coordinate system in which its mapped coordinates are interpreted. One coordinate mapping may be primary.
+Each axis group may have coordinate calibrations and may reference the coordinate system in which its calibrated coordinates are interpreted. One coordinate calibration may be primary.
 
 ### Terms and Definitions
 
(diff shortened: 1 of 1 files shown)


design/data_model.md
```diff
@@ -94,11 +92,153 @@ Lossless serialization and exact copying preserve all extension records, includi
...
+        ],
+        "coordinate_system_id": "camera",
+        "coordinate_calibrations": {
+          "spatial": {
+            "label": "Real-space (camera)",
+            "calibrations": [
+              {
+                "scale": 0.1,
+                "offset": 0.0,
+                "unit": "nm"
```
Brow71189, line 140: I have a question about how we store calibrations here: There has been criticism in the past that Swift is not aware of metric prefixes of the the units used. Couldn't we store the unit (`m`) and the metric prefix or the corresponding scaling factor separately?
Or actually we should probably always store them in SI units, so the example above would become

```
{
  "scale": 1e-10,
  "offset": 0.0,
  "unit": "m",
  "preferred_display_scaling": 1e-9
``` 

`"preferred_display_scaling"` can be used to display the units with a fixed metrix prefix all the time, so that the units are not jumping back and forth when you cange the FOV on your scan for example.

Tiomat85: I like this as it opens up significantly easier handling of actual unit intelligence where computations can follow fixed rules on how to calculate the resultant calibration.

cmeyer: This PR is for persistent format, and primarily at a high level. So I don't think this change falls into this PR.

But this is a good point. I'll propose an improved calibration object. The standard is to measure a "quantity" which is a value, unit, and "dimension" where dimension is a vector of exponents over the seven SI base "dimensions" — length, mass, time, electric current, thermodynamic temperature, amount of substance, and luminous intensity — with rational exponents.

cmeyer: Filed follow-up to-do:
- https://github.com/nion-software/nionswift/issues/1930

PR #1922, opened 2026-07-23
Preliminary refactoring to prepare for upcoming computation rework

cmeyer:
This PR contains preliminary refactoring of computations to prepare for the upcoming computation registration and execution rework supporting iteration of any axis group and improved computations.

This PR should not change any behavior in the existing UX.

The last commit in the PR is special handling for existing computations in experimental and eels-analysis, to be used only until all computations get migrated to the new system.

As with many of the refactoring PR's, this one has a lot of complicated detailed work to prepare for a future PR. When evaluating this PR, I'm looking for outright mistakes; but also comments on clarity of the code and other subjective comments are welcome, although subjective comments may be addressed in a future PR instead of this one.

- **Allow computation processors to be registered and used to execute computations.**
- **Merge/simplify processor lists.**
- **Eliminate ProcessingBase/Processing.py by merging functionality into new computations.**
- **Refactor computation processor to combine 'sources' and 'parameters' into 'inputs'.**
- **Change ComputationProcessorRequirement to validate using data_metadata rather than data_item.**
- **Change computation output handling to upcast numpy array compatible to DataAndMetadata.**
- **Separate computation processor variable type from computation variable type.**
- **Add computation processor list/map handling.**
- **Add special handling for existing registered computations.**


nion/swift/model/Symbolic.py
```diff
@@ -3568,76 +3677,270 @@ def needs_update_for_event(self, event_type: BoundDataEventType) -> bool:
...
+            return ComputationProcessorListInput.from_dict(d)
+        if "list" in input_name or "shifts" in input_name:
+            return ComputationProcessorListInput.from_dict(d)
+        if "graphic" in input_name or "region" in input_name:
+            return ComputationProcessorValueInput(ComputationProcessorVariableType.REGION_2D, input_name, input_label)
+        if "gamma" in input_name or "weight" in input_name or "rotation" in input_name:
+            return ComputationProcessorValueInput(ComputationProcessorVariableType.REAL, input_name, input_label)
+        if "index" in input_name:
+            return ComputationProcessorValueInput(ComputationProcessorVariableType.INTEGER, input_name, input_label)
+        if "enabled" in input_name or "flip" in input_name or "show" in input_name or "crop_to_valid":
```
Tiomat85: crop_to_valid being used as truthy here.

nion/swift/model/Symbolic.py
```diff
@@ -3568,76 +3677,270 @@ def needs_update_for_event(self, event_type: BoundDataEventType) -> bool:
...
+        self.old_built_in = False
+
+    @staticmethod
+    def _from_source(d: PersistentDictType) -> ComputationProcessorInput:
+        # handle backward compatibility until all clients updated
+        input_name = typing.cast(str, d["name"])
+        input_label = typing.cast(str | None, d.get("label"))
+        if input_name == "src" or "data_item" in input_name:
+            return ComputationProcessorDataInput(input_name, input_label, "xdata", tuple(), tuple(), False)
+        if "regions" in input_name:
```
Tiomat85, line 3872: This repeating pattern feels like it should have a better solution:

```
def _containing(s: str, options:typing.Sequence[str]):
    return any(k in s for k in options)
```

or if they are supposed to be == instead of just in, a match statement with sets would be cleaner.

nion/swift/model/DocumentModel.py
```diff
@@ -2483,7 +2469,7 @@ def _get_builtin_processors(cls) -> dict[str, Symbolic.ComputationProcessor]:
                 "control_type": "slider"}
             vs["gaussian-blur"] = {"title": _("Gaussian Blur"), "expression": "xd.gaussian_blur({src}.cropped_display_xdata, sigma)",
```
Tiomat85, line 2470: It is a bit nitpicky but since we are refactoring can we try make these easier to follow? Something like:

 ```
             vs["fft"] = {"title": _("FFT"),
                         "expression": "xd.fft({src}.cropped_display_xdata)",
                         "sources": [{"name": "src",
                                      "label": _("Source"),
                                      "croppable": True,
                                      "data_type": "cropped_display_xdata"}]}
            
            vs["inverse-fft"] = {"title": _("Inverse FFT"),
                                 "expression": "xd.ifft({src}.xdata)", 
                                 "sources": [{"name": "src",
                                              "label": _("Source"),
                                              "data_type": "xdata"}]}
```

With the amount of them, it might even be worth considering moving them to their own file just containing the default computations.

File: nion/swift/model/Symbolic.py [hunk 4]
```diff
@@ -3394,8 +3411,14 @@ def __init__(self, values: typing.Sequence[int]) -> None:
     def from_dict(cls, d: PersistentDictType) -> ComputationProcessorRequirementDataRank:
         return cls(d["values"])
 
-    def is_data_item_valid(self, data_item: DataItem.DataItem) -> bool:
-        return data_item.datum_dimension_count in self.values
+    def to_dict(self) -> dict[str, typing.Any]:
+        return {
+            "type": "datum_rank",
```
Tiomat85, line 3416: This type differing from the name of the class feels awkward. Can it at least be data_rank?  
Also, since type appears to be 'mandatory' for all of these could we push that up to the ComputationProcessorRequirement as a property, maybe even with it automatically putting them in the base classes to_dict in a standard way and each derivation appending relevant stuff to the dictionary returned from super().to_dict()

nion/swift/model/DocumentModel.py
```diff
@@ -3025,43 +3049,15 @@ def close(self) -> None:
...
-
-_handle_processing_component_registered_listener = Registry.listen_component_registered_event(handle_processing_component_registered)
-# listening to processing-component is used in two places, so do not use `fire_existing_...`
-for component in Registry.get_components_by_type("processing-component"):
-    handle_processing_component_registered(component, {"processing-component"})
+def _register_processors() -> None:
+    for processing_id, computation_processor in DocumentModel._get_builtin_processors().items():
+        # temporary hack to handle mapped scalar and window functions
+        if not processing_id.startswith("mapped-") and not processing_id.endswith("-window"):
+            computation_processor.old_built_in = True
```
Tiomat85, line 3056: Positive logic is generally easier to understand intent, maybe make it more expressive?

```
is_new_style = processing_id.startswith("mapped-") or processing_id.endswith("-window")
if not is_new_style:
    computation_processor.old_built_in = True
```

nion/swift/model/DocumentModel.py
Tiomat85, line 2470: It is a bit nitpicky but since we are refactoring can we try make these easier to follow? Something like:

 ```
             vs["fft"] = {"title": _("FFT"),
                         "expression": "xd.fft({src}.cropped_display_xdata)",
                         "sources": [{"name": "src",
                                      "label": _("Source"),
                                      "croppable": True,
                                      "data_type": "cropped_display_xdata"}]}
            
            vs["inverse-fft"] = {"title": _("Inverse FFT"),
                                 "expression": "xd.ifft({src}.xdata)", 
                                 "sources": [{"name": "src",
                                              "label": _("Source"),
                                              "data_type": "xdata"}]}
```

With the amount of them, it might even be worth considering moving them to their own file just containing the default computations.

File: nion/swift/model/Symbolic.py [hunk 6]
```diff
@@ -2074,6 +2079,15 @@ def __init__(self, expression: typing.Optional[str] = None) -> None:
         # if the computation is running and needs to be deleted, this can delay deletion until the computation finishes.
         self.is_deleted = False
 
+    def close(self) -> None:
+        self.__variable_changed_event_listeners = typing.cast(typing.Any, None)
```
Tiomat85, line 2083: This is ugly, why not either just `.clear()` on the list, or allocate to a new list.  If not, then the type-change should be in the `__init__` making them optional to indicate that it has the possibility of being None and you don't need this code-smell of a cast.

nion/swift/model/Symbolic.py
```diff
@@ -2618,10 +2626,11 @@ def evaluate(self, api: typing.Any) -> typing.Optional[ComputationExecutor]:
         needs_update = self.needs_update
         self.needs_update = False
         if needs_update:
-            if self.expression:
-                executor = ScriptExpressionComputationExecutor(self, api)
-            else:
+            use_registered_executor = not self.expression or (self.computation_processor is not None and not self.computation_processor.old_built_in)
```
Tiomat85, line 2629: Similarly here I'd tweak the logic to make it a bit clearer:

```
has_new_registered_processor = self.computation_processor is not None and not self.computation_processor.old_built_in

if self.expression and not has_new_registered_processor:
    executor = ScriptExpressionComputationExecutor(self, api)
else:
    executor = RegisteredComputationExecutor(self, api)
```

nion/swift/model/Symbolic.py
```diff
@@ -3568,76 +3677,270 @@ def needs_update_for_event(self, event_type: BoundDataEventType) -> bool:
...
+                return ComputationVariableType.BOOLEAN
+            case ComputationProcessorVariableType.INTEGER:
+                return ComputationVariableType.INTEGRAL
+            case ComputationProcessorVariableType.REAL:
+                return ComputationVariableType.REAL
+            case ComputationProcessorVariableType.COMPLEX:
+                return ComputationVariableType.COMPLEX
+            case ComputationProcessorVariableType.STRING:
+                return ComputationVariableType.STRING
+            case ComputationProcessorVariableType.REGION_2D:
```
Tiomat85, line 3727: What is the distinction between a region_2D and a rectangle?

Tiomat85 (changes requested):
General comments, some of which stretch the scope a bit.

Should the DocumentModel really be the interface point for computations, with all the computation logic etc or should it be extracted into a distinct object, perhaps that DocumentModel inherits from, just to try split out the roles for clarity.  DocumentModel is a huge class that does lots and lots of things so each individual role it is trying to do becomes diluted amongst all of the other parts.

Also, is there scope with this for extending to different types of Executors - multithreaded compiled libraries, GPU accelerated and the like?

nion/swift/model/Symbolic.py
cmeyer: I'll fix the 'crop_to_valid' code, but overall this section is throwaway code intended to be a bridge until `nionswift-experimental` is refactored to the new API.

nion/swift/model/DocumentModel.py
cmeyer: Again: throwaway code. All of this will change with annotated arrays.

nion/swift/model/DocumentModel.py
cmeyer: More throwaway code.

nion/swift/model/DocumentModel.py
cmeyer: More throwaway. Indeed they will be moved out of this file and into JSON + library.

nion/swift/model/Symbolic.py
cmeyer: 2D is a region applicable to 2D data. A rectangle is a rectangle. i.e. a 'line' qualifies as a REGION_2D; an 'interval' does not. I can add more documentation about this, but it is also likely to change in a future PR.

nion/swift/model/Symbolic.py
cmeyer: The intention is to (a) garbage collect; (b) cause issues if it is used after `close`. A `clear` or assignment of a new `list` doesn't accomplish (b).

cmeyer:
> Should the DocumentModel really be the interface point for computations, with all the computation logic etc or should it be extracted into a distinct object, perhaps that DocumentModel inherits from, just to try split out the roles for clarity. DocumentModel is a huge class that does lots and lots of things so each individual role it is trying to do becomes diluted amongst all of the other parts.

This will be addressed in a follow-up PR. I don't expect any registration to take place directly in `DocumentModel`.

> Also, is there scope with this for extending to different types of Executors - multithreaded compiled libraries, GPU accelerated and the like?

The first pass will not try to address threading or pipelining. The first pass is primarily to implement general iteration and mapping of computations applying to a single target axis group of 1d or 2d data. e.g. apply an FFT to every point in a sequence of collections of 2d signal data.

nion/swift/model/Symbolic.py
cmeyer: Is this comment referring to an intermediate comment? It seems correct to me: `ComputationProcessorRequirementDatumRank` -> `"datum_rank"`

Either way, I expect that this will be updated in a future PR to be match the annotated array and processing operation designs, i.e. "datum" will be replaced with "signal".

nion/swift/model/Symbolic.py
cmeyer: Your change gives clearer logic, but I want the newer `RegisteredComputationExecutor` first in the `if`. `ScriptExpressionComputationExecutor` may go away entirely. So I'm going to leave this alone for now.

nion/swift/model/Symbolic.py
cmeyer: Fixed.

nion/swift/model/Symbolic.py
Tiomat85:
```
class ComputationProcessorRequirementDataRank(ComputationProcessorRequirement):
    def __init__(self, values: typing.Sequence[int]) -> None:
        self.values = list(values)

    @classmethod
    def from_dict(cls, d: PersistentDictType) -> ComputationProcessorRequirementDataRank:
        return cls(d["values"])

    def to_dict(self) -> dict[str, typing.Any]:
        return {
            "type": "datum_rank",
            "values": self.values
        }
```

Classname is DataRank, type string is "datum_rank".  It is a minor thing, but those slight naming mismatches get annoying over time.

nion/swift/model/Symbolic.py
Tiomat85: Then surely the correct solution is to define the type to be Optional?  Using an ugly, incorrect cast to force the wrong type into an object is not something we should do.

nion/swift/model/Symbolic.py
Tiomat85: Then just flip the final if statement:
`if has_new_registered_processor or not self.expression:`

If you agree the logic is cleaner, and clarity of logical intent is important what is the justification for not adjusting this?

nion/swift/model/Symbolic.py
Tiomat85: Some documentation at least here would be useful, as at initial glance a rectangle is a 2D shape on 2D data, and the Region_2D sounded the same so the distinction in exactly when each is appropriate to be used would be good.  
It also feels like there might be some stretching of the enum a bit that it is trying to encode two different things in a single enum? What type is the parameter, and what is it applicable to?

nion/swift/model/Symbolic.py
cmeyer: It turns out I've already fixed this in a later commit stacked on this PR. But the fix was to change `ComputationProcessorRequirementDataRank` to `ComputationProcessorRequirementDatumRank`. And the further-down-the-stream fix is `ComputationProcessorRequirementSignalRank`.

I'll look into pulling that bit of the future commit into this PR.

nion/swift/model/Symbolic.py
cmeyer: This idiom is used throughout the code base. After close, an object is invalid. The cast means use-after-close fails at the point of misuse. `Optional` turns that into a defensive `if` that silently does nothing, in hundreds of places.

nion/swift/model/Symbolic.py
cmeyer: I've decided to leave this alone. `ComputationProcessorRequirementDataRank` is misnamed, not the `type` entry in the `dict`. And `ComputationProcessorRequirementDataRank` has been misnamed for two years - it'll be updated naturally as these requirements are generalized for `AnnotatedArray`.

cmeyer:
This PR is ready to be merged, but I am delaying merging until the follow-up PR is ready to minimize disruption to experimental computations.

cmeyer:
Merged 87aa81e4246846375365133eaa490f1d59057f23..200206e1107b0dd30dae37e3a4f22c12dbeb5a48

PR #1926, opened 2026-07-29
Name project dialog

### LINKED ISSUE #1896: New Project doesn't show warning that project already exists
This happens more likely when creating 2 projects on the same day.
If a user attempts to create a project in a specific folder which matches the name of another project it will silently fail.

Loading multi-acquire settings from C:\Users\Jonathan.Feito\AppData\Local\Nion\Nion Swift\multi_acquire_settings.json (if it exists)
Loading multi-acquire parameters from C:\Users\Jonathan.Feito\AppData\Local\Nion\Nion Swift\multi_acquire_spectrum_parameters.json (if it exists)
Using cache C:\Users\Jonathan.Feito\AppData\Roaming\Nion\Nion Swift\Cache\proj_Y41XHTEZSWDHTB5ZI2M5JXFYO.nscache
Loading project C:\Users\Jonathan.Feito\Desktop\Regression display.nsproj
Loaded project 1s (data items: 15, display items: 15, data structures: 0, computations: 1, connections: 0, data groups: 1, workspaces: 1)
Loading multi-acquire settings from C:\Users\Jonathan.Feito\AppData\Local\Nion\Nion Swift\multi_acquire_settings.json (if it exists)
Loading multi-acquire parameters from C:\Users\Jonathan.Feito\AppData\Local\Nion\Nion Swift\multi_acquire_spectrum_parameters.json (if it exists)
  File "C:\NionApps\Developer\main\nionui\nion\ui\UserInterface.py", line 1526, in handle_clicked
    self.on_clicked()
    ~~~~~~~~~~~~~~~^^
  File "C:\NionApps\Developer\main\nionui\nion\ui\Dialog.py", line 141, in on_clicked
    do_close = on_clicked_fn()
  File "C:\NionApps\Developer\main\nionswift\nion\swift\Application.py", line 1034, in handle_new
    app.create_project_reference(pathlib.Path(self.directory), self.__project_name_field.text or "untitled")
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\NionApps\Developer\main\nionswift\nion\swift\Application.py", line 850, in create_project_reference
    project_reference = profile.create_project(directory, project_name)
  File "C:\NionApps\Developer\main\nionswift\nion\swift\model\Profile.py", line 631, in create_project
    self.append_project_reference(project_reference)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "C:\NionApps\Developer\main\nionswift\nion\swift\model\Profile.py", line 655, in append_project_reference
    assert project_reference.project_reference_id not in {project_reference.project_reference_id for project_reference in self.project_references}
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError

MattRoyle:
Fixes #1896 
Refactors the new project dialog to use a declarative name project dialog which is intended to also be used in project renaming.
This requires adding a function to the project storage system that can check if the given name is available.

### DIFF
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -59,6 +59,7 @@ from nion.utils import Geometry
 from nion.utils import ListModel
 from nion.utils import Model
 from nion.utils import Registry
+from nion.utils import Stream
 
 if typing.TYPE_CHECKING:
     from nion.swift.model import DisplayItem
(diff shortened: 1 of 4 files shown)


nion/swift/Application.py
```diff
@@ -1001,148 +1002,50 @@ def create_handler(self, component_id: str, container: typing.Any = None, item:
...
-
-            def __init__(self, ui: UserInterface.UserInterface, app: Application, event_loop: asyncio.AbstractEventLoop, profile: Profile.Profile) -> None:
-                super().__init__(ui, title=_("New Project"), app=app, persistent_id="new_project_dialog")
-
-                self.directory = self.ui.get_persistent_string("project_directory", self.ui.get_document_location())
+    @staticmethod
+    def get_project_base_name(directory: str) -> str:
+        """Get a default project base name using the current datetime.
 
-                project_base_name = _("Nion Swift Project") + " " + datetime.datetime.now().strftime("%Y%m%d")
```
MattRoyle, line 997: The old new project dialog had the functionality to append a index to the end of the default name if it already existed, however this code can't work because it is using creating checking a **path** with that name doesn't exist when it should be checking a **file** doesn't exist.

I modified this to use the check_project_name_is_available function and changed it to enclose the index in brackets as is standard for windows, although I think macOS doesn't add curly braces around the increment.

nion/swift/model/FileStorageSystem.py
```diff
@@ -642,7 +661,15 @@ def _read_library_properties(self, migration_stage: ProjectStorageSystemMigratio
     @abc.abstractmethod
     def normalize_project_data_path(self) -> None: ...
 
-    #
+    @staticmethod
+    def check_project_name_is_available(name: str, directory: str) -> ProjectNameResult:
```
cmeyer: `ProjectStorageSystem.check_project_name_is_available` is not used. Can it be removed?

nion/swift/Application.py
cmeyer: I don't see a change in functionality - am I missing something?

Also, I'm ok with the "file (1)" style - but this has gotten direct complaints about not using parenthesis in filenames. What do others (in the office) think about this? (not a blocker, just something to look at)

nion/swift/Application.py
```diff
@@ -1001,148 +1002,50 @@ def create_handler(self, component_id: str, container: typing.Any = None, item:
...
-                project_base_index = 0
-                project_base_index_str = ""
-                while os.path.exists(os.path.join(self.directory, project_base_name + project_base_index_str)):
-                    project_base_index += 1
-                    project_base_index_str = " " + str(project_base_index)
-
-                self.project_name = project_base_name + project_base_index_str
+        If a project already exists with that name in the directory then try appending an index to the name until a unique name is found.
+        The max index checked is 999, so in that case the project name will not be unique.
```
cmeyer, line 1012: Very minor thing, but we could just add a unique generated string to the filename if it gets > 99 (or 9?). Generate. Check. Continue. Up to 1000 checks, but almost certainly will never get to 102.

```
import secrets

ALPHABET = "34679ACDEFHJKMNPRTVWXY"  # 22 symbols, no visual pairs

def new_id(length: int = 4) -> str:
    return "".join(secrets.choice(ALPHABET) for _ in range(length))
```

cmeyer (changes requested):
One removal (extraneous `check_project_name_is_available`) required.

One question (use parenthesis or not on unique filenmame).

One suggestion (use trivial unique id generator after 100 (or 10) unique file tries).

nion/swift/Application.py
jamesrussell216: I'm OK with either. Chris is right we have had objections to parenthesis in the past, but it is a Windows standard so I think still defensible.

nion/swift/Application.py
MattRoyle: Both macos and windows use an increment when saving a file that already appears on the disk. I don't think using a random value in the filename is a good user experience. Also you would need to run this multiple times in case the choice happens to already be used. 
This is just the default filename when the popup is created, which has the date added to it year month day. 
Someone would need to make 1000 projects in a single day for this to be non-unique. I could increase the max index to 9999.
The default name already being used doesn't lead to any software breaking side effects it just means the popup will start with an invalid filename error saying the name already exists.

nion/swift/Application.py
MattRoyle: The old dialog had the code for this increment but it wasn't checking the project path so it was always skipped even when the name already exists. This fixes that bug by checking the project path.

nion/swift/model/FileStorageSystem.py
MattRoyle: In the rename projects PR #1927, the instance of a storage_system is used to pass the check_project_name_is_available which is why it is in the base ProjectStorageSystem so other types of storage systems can implement the specific function to check it is available.

cmeyer: The multiple uses of `check_project_name_is_available` make the logic hard to follow: same name, different contexts and parameters, even with one static method.

This is standing in for an object that manages the file storage system files themselves, and substituting an ad-hoc inheritance system for `check_project_name_is_available`.

My preference is a separate PR introducing `ProjectStorageSystemManager` with overridable methods to 'make the storage system' and to 'check whether the name is available'. This would be traditional object inheritance. Then put this PR on top of that one.

Or: drop `ProjectStorageSystem.check_project_name_is_available` from this PR and we can pick up the discussion in the follow-up.

nion/swift/Application.py
cmeyer: The problem is your algorithm fails at 1000.

MattRoyle: No it just gives a non-unique name, which is a case that is already handled. I could increase the max index to the max int or some other large number which negates the likelihood of that being reached.

nion/swift/model/FileStorageSystem.py
MattRoyle: We need to keep both this PR and the project renaming PR independent of any future idea of a ProjectStorageSystemManager. It is a unspec'd refactor that could go in a number of directions and will take a significant chunk of time to get right. For instance when I spoke to Kev about it, he and I had very different ideas of what it will be and how much it will be in charge of. I would say that the profile already "manages" the project storage implicitly in open_project and the add_project_index folder etc. So either it would be refactoring such areas to use the new manager or adding additional handling there. 
I will remove the ProjectStorageSystem.check_project_name_is_available and instead put the logic in a function outside of the storage system class.

cmeyer (approved):
This looks ready to go to me. I'll wait for you to mark it as ready (it's draft now).

Ion-e:
On testing: check what happens when user deletes project from filepath, and creates a new project with the same name.

nion/swift/Application.py
```diff
@@ -1175,6 +1078,180 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+
+        self.accept_button_enabled.value = is_valid
+        if is_valid:  # The name is valid and doesn't already exist.
+            self.project_name_status_label.value = str()
+        elif errors:  # Display the error messages.
+            error_str = "\n".join(errors)
+            self.project_name_status_label.value = error_str
+
+
+class NameProjectDialog(Declarative.Handler):
```
Tiomat85: Should probably have an overridden close function to handle closing the ValueStreamActions you create (__accept_button_enabled_action and __accept_button_tooltip_action)

```
def close(self) -> None:
    self.__accept_button_enabled_action.close()
    self.__accept_button_enabled_action = typing.cast(typing.Any, None)
    self.__accept_button_tool_tip_action.close()
    self.__accept_button_tool_tip_action = typing.cast(typing.Any, None)
    super().close()
```

nion/swift/Application.py
```diff
@@ -1175,6 +1078,180 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+            project_name_header_row,
+            u.create_spacing(8),
+            project_name_row,
+            u.create_spacing(4),
+            project_name_status_row,
+            u.create_stretch(),
+            u.create_spacing(16)
+        )
+
+        self.dialog = typing.cast(Dialog.ActionDialog, Declarative.construct(document_controller.ui, document_controller, u.create_modeless_dialog(column, title=title), self))
```
Tiomat85: It might be worth extracting the creation of column into a 'create_dialog_ui' function to try keep the __init__ a bit shorter.

nion/swift/Application.py
```diff
@@ -1175,6 +1078,180 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+        self.viewmodel.update_project_status_label()
+        self.dialog.show()
+        assert self._project_name_line_edit is not None
+        self._project_name_line_edit.focused = True
+        self._project_name_line_edit.select_all()
+
+    def handle_project_name_changed(self, _widget: UserInterface.LineEditWidget, text: str) -> None:
+        self.viewmodel.update_project_status_label(text)
+
+    def choose_directory(self, _widget: UserInterface.PushButtonWidget) -> None:
```
Tiomat85: Maybe change this to handle_choose_directory to match the others?

Tiomat85 (approved):
One little potential for a resource leak, and a couple of tidyness comments.

nion/swift/Application.py
Tiomat85: I think that is a perfectly reasonable limitation to accept, as if we are hitting the point of thousands of projects with identical names per day then a new issue can be raised and we look at how users are getting into that much of a tangle.  Performance at higher numbers might be noticeable but again you are needing hundreds of uniquely named projects per day so that is way into the realm of not-a-realistic-problem.

Tiomat85 (approved):
LGTM

Ion-e (approved):
LGTM. Nice one

Issue #492, opened 2020-07-28
Switching projects should show progress bar when loading new project


cmeyer:
Currently, any existing window is closed and the new project loads without giving any feedback to the user. This can be confusing for the duration of the load, which can be several seconds for large projects.

MattRoyle:
I would like implement this as a follow up to project renaming #1927. The gap where the current project is closed and renamed then reopened can take several seconds which can easily be confused for failure. 
Does this need a progress bar or could it just be a splash screen to reduce complexity?
Many applications have a progress message like "Building shaders" or "loading project data", this could be the last message in the terminal.

cmeyer:
A modal dialog would be a great start.

MattRoyle:
We discussed in backlog refinement the notes are below:
This could be expand to “Loading projects” although in the build version there is already a splash screen.
Potentially show last log message as a indicator of progress.
Having a *throbber* (animated loading icon) probably wont work even if it updated on a separate thread. It also wouldn't be as helpful as a progress bar in terms of informing the user as to the progress of switching/loading.
Might be useful to have stages of loading, i.e. loading plugins, reading data files, creating display items etc.
Loading data items may be a place that has more granular progress to show loaded out of total items.
Also show current data item being loaded so we know which it has failed on
SPIKE: check we can do anything here other than 0 & 100% for a progress bar.

This could be done in stages of having the initial modal splash screen before moving on to implementing user feedback of loading/progress.

PR #1927, opened 2026-07-29
Rename project on disk

### LINKED ISSUE #504: Add ability to rename projects on disk

MattRoyle:
Fixes #504. ~Requires #1926.~
This PR was split from #1874 since the commit history and PR comments made it hard to discuss/review properly.
This requires the dialog and check function from #1926.

The current project can be renamed via a new menu item in File "Rename Project". This closes the current project, performs the renaming, then reopens the project. Other projects can also be renamed via the "Choose Project" menu. You can right click one of the projects in the list and select "Rename Project" to rename the project (this does not close the existing project or open the renamed one).

When renaming the current project any active tasks running may be interrupted since the project is closed then reopened. This is not prevented and any issues caused by this are out of the scope of this PR.

### DIFF
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -939,6 +939,41 @@ class Application(UIApplication.BaseApplication):
 
         self.event_loop.call_later(command_delay or 0.0, execute_next_command, commands, list(commands), command_delay or 0.0, repeat or 1)
 
+    def rename_project(self, project_reference: Profile.ProjectReference, name: str) -> None:
+        """Rename the project reference to the given name.
+
+        If the project reference is the current project, it will be closed and reopened after renaming to update the title.
+        """
+        with self.prevent_close():
+            old_title = project_reference.title
+            renaming_current_project = self.profile.last_project_reference == project_reference.uuid
+            document_controller = self.document_controllers[0]
+            assert document_controller is not None
+            if renaming_current_project:
+                document_controller.request_close()
+            error_message: str | None = None
+            new_project_reference: Profile.ProjectReference | None = None
+            try:
+                new_project_reference, rename_result = self.profile.rename_project(project_reference, name)
+                error_message = "\n".join(rename_result.error_messages)
+            except Exception as e:
+                # Catch and log any exceptions during the rename process
+                error_message = _("Exception during renaming project")
+                logging.exception(e)
+                new_project_reference = project_reference
+            finally:
+                # Reopen the project only when renaming the current project and when the project reference is returned
+                if renaming_current_project and new_project_reference is not None:
+                    self.switch_project_reference(new_project_reference)
+                    assert new_project_reference.project is not None
+                    new_project_reference.project.title = name
+                    self.profile.last_project_reference = new_project_reference.uuid
+
+                if error_message:
+                    self.__show_project_error_dialog(_("Error Renaming Project"), error_message)
+                else:
+                    logging.info(f"Renamed Project \"{old_title}\" to \"{name}\"")
+
 
 def get_root_dir() -> str:
     root_dir = os.path.dirname((os.path.dirname(os.path.abspath(__file__))))
--- a/nion/swift/Application.py
+++ b/nion/swift/Application.py
@@ -1212,7 +1212,7 @@ class NameProjectViewModel:
 
     def update_project_status_label(self, project_name: str | None = None, directory: str | None = None) -> None:
         project_name = project_name or self.filename.value or str()
-        directory = directory or self.directory.value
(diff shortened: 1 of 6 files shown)


nion/swift/model/FileStorageSystem.py
```diff
@@ -1163,6 +1173,48 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
...
+            self.__project_path = old_project_path.rename(new_project_path)
+        except Exception as e:
+            logging.exception(_("Failed to rename project. Renaming Project File gave an exception:\n") + str(e))
+            return ProjectNameResult([_("Exception while renaming the Project File")], self.__project_path)
+
+        error_messages = []
+        try:
+            # The directory must be checked to exist because load_properties will set the data path to the default even when the directory does not exist
+            if old_data_path is not None:
+                self.__project_data_path = old_data_path.rename(old_data_path.parent / f"{name} Data")
```
KRLango: I would define `new_data_path` with `old_data_path`, `old_project_path` and `new_project_path` rather than inline here. That way all paths are treated the same.

nion/swift/model/FileStorageSystem.py
```diff
@@ -1163,6 +1173,48 @@ def __find_storage_handlers(self, directory: typing.Optional[pathlib.Path], *, s
...
+        new_project_path = old_project_path.parent / f"{name}.nsproj"
+        try:
+            self.__project_path = old_project_path.rename(new_project_path)
+        except Exception as e:
+            logging.exception(_("Failed to rename project. Renaming Project File gave an exception:\n") + str(e))
+            return ProjectNameResult([_("Exception while renaming the Project File")], self.__project_path)
+
+        error_messages = []
+        try:
+            # The directory must be checked to exist because load_properties will set the data path to the default even when the directory does not exist
```
KRLango: None of the code here is checking if a directory exists. This comment should be amended or removed.

nion/swift/Application.py
```diff
@@ -1058,6 +1101,68 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+        super().__init__()
+        self.accept_fn = accept_fn
+        self.project_reference = project_reference
+
+    action_id = "project.rename_project"
+    action_name = _("Rename Project")
+    action_parameters = [
+        UIWindow.ActionStringProperty("name")
+    ]
+    check_project_name_is_available = FileStorageSystem.check_file_project_name_is_available  # The FileProjectStorageSystem is used as the default project storage system
```
KRLango: This is a very long line, I would move the comment to the line above

nion/swift/Application.py
```diff
@@ -1058,6 +1101,68 @@ def invoke(self, context_: UIWindow.ActionContext) -> UIWindow.ActionResult:
...
+
+    def execute(self, context: UIWindow.ActionContext) -> UIWindow.ActionResult:
+        assert self.project_reference is not None
+        project_name = self.get_string_property(context, "name")
+        assert project_name is not None
+        application = typing.cast(Application, context.application)
+        application.rename_project(self.project_reference, project_name)
+        self.project_reference = None  # Clear the now invalid project_reference so if it is invoked again it get the current project reference again
```
KRLango: Another long line where the comment should be moved to the line above. Also can it be reworded - the grammar is very strange at the moment. Something like: `Clear the now invalid project_reference so if this is invoked again it gets the latest project reference`

KRLango (commented):
A few minor bits, but otherwise looking good to me.

MattRoyle:
When renaming the current project, the swift window is closed then the project is renamed, then project is reopened via the switch project mechanism. I've noticed that in certain projects this can take several seconds, upwards of 10 in some cases. This was confusing enough for me to think something was wrong and end the pycharm instance. I imagine users may also think something has gone wrong and start fiddling.
I think having a [loading screen](https://github.com/nion-software/nionswift/issues/492) or splash screen will be a necessary follow up to this work to mitigate this.

Ion-e:
Bug is:
Choose project > right click > rename 
Actual: observe that the edit box is highlighted but it's not possible to type directly into it, user has to click on it

Expected: either not having it "automatically highlighted" or having the focus to actually be in that edit box

jamesrussell216:
@Ion-e discussed this with @MattRoyle 

I think that's a minor issue at this stage. Happy to have it in an issue for follow up but it's a deeper problem than just this PR so I think this should go in despite that.

Issue #1718, opened 2025-12-09
Improve line plot rendering with pixel density >> 1


Brow71189:
Take the following example code:

```
d = np.zeros(2048)
d[100] = 1
show(d)
d = np.zeros(1024)
d[100] = 1
show(d)
d = np.zeros(512)
d[100] = 1
show(d)
```

This results in the images below (from left to right). You would expect that the y-values of all three plots are 1, but they seem to scale with how much the plot has to be compressed in x to fit into the display panel.

<img width="2323" height="739" alt="Image" src="https://github.com/user-attachments/assets/f834c83a-c1f3-47c2-8054-8eb61ae7ac35" />

cmeyer:
The _scale_ of the y-axis is the same in all three screenshots, but the actual plot differs because of a binning optimization.

The graph algorithm is here [algorithm for drawing the graph](https://github.com/nion-software/nionswift/blob/960e9576351296077d547cf44935cfcca17b75fc/nion/swift/LineGraphCanvasItem.py#L430).

Here is an alternative, not tested for correctness or performance:

```
diff --git a/nion/swift/LineGraphCanvasItem.py b/nion/swift/LineGraphCanvasItem.py
index 1a777a6b..996ea24e 100644
--- a/nion/swift/LineGraphCanvasItem.py
+++ b/nion/swift/LineGraphCanvasItem.py
@@ -542,7 +542,7 @@ def calculate_line_graph(plot_height: int, plot_width: int, plot_origin_y: int,
         baseline = max(plot_origin_y, baseline)
         # rebin so that uncalibrated_width corresponds to plot width
         calibrated_data = calibrated_xdata._data_ex
-        binned_length = int(calibrated_data.shape[-1] * plot_width / uncalibrated_width)
+        binned_length = calibrated_data.shape[-1]
         did_draw = False
         if binned_length > 0:
             binned_data = Image.rebin_1d(calibrated_data, binned_length, rebin_cache)
@@ -551,7 +551,7 @@ def calculate_line_graph(plot_height: int, plot_width: int, plot_origin_y: int,
             # draw the plot
             last_py = baseline
             for i in range(0, plot_width):
-                px = plot_origin_x + i
+                px = plot_origin_x + i * plot_width / binned_length
                 binned_index = binned_left + i
                 if binned_index >= 0 and binned_index < binned_length and not binned_data_is_nan[binned_index]:
                     data_value = binned_data[binned_index]
```

PR #1931, opened 2026-08-04
Adds in graph_style options for Line Graphs

Tiomat85:
Adds a graph_style property to LineGraphs.  graph_style can be either None (defaulting to Bar), "bar" which is the stepped bar graph, or "line" which provides a simple line graph between the provided data points.
A selector is added to the Inspector to allow for switching between these types.

### DIFF
--- a/nion/swift/model/DisplayItem.py
+++ b/nion/swift/model/DisplayItem.py
@@ -1623,6 +1623,7 @@ class DisplayLayer(Schema.Entity):
     stroke_color = Schema.EntityAttribute[typing.Optional[str]]()
     fill_color = Schema.EntityAttribute[typing.Optional[str]]()
     stroke_width = Schema.EntityAttribute[typing.Optional[int]]()
+    graph_style = Schema.EntityAttribute[typing.Optional[str]]()  # "bar" | "line" | "scatter"
     display_data_channel = Schema.EntityAttribute[typing.Optional[DisplayDataChannel]]()
 
     def __init__(self, display_layer_properties: typing.Optional[Persistence.PersistentDictType] = None) -> None:
@@ -1684,6 +1685,7 @@ class DisplayLayerInfo:
     stroke_color: str | None
     fill_color: str | None
     stroke_width: int | None
+    graph_style: str | None = None  # line-plot style: "bar" | "line" | "scatter" (None = default "bar")
 
 
 @dataclasses.dataclass
@@ -2700,7 +2702,8 @@ class DisplayItem(Persistence.PersistentObject):
                 label=label,
                 stroke_color=display_layer.stroke_color,
                 fill_color=display_layer.fill_color,
-                stroke_width=display_layer.stroke_width
+                stroke_width=display_layer.stroke_width,
+                graph_style=display_layer.graph_style,
             )
             display_layer_list.append(display_layer_info)
         return display_layer_list
@@ -2721,7 +2724,7 @@ class DisplayItem(Persistence.PersistentObject):
     @display_layers_list.setter
     def display_layers_list(self, value: list[DisplayLayer]) -> None:
         assert len(value) == len(self.display_layers)
-        properties = ["data_row", "fill_color", "stroke_color", "label", "stroke_width"]
+        properties = ["data_row", "fill_color", "stroke_color", "label", "stroke_width", "graph_style"]
         for index, (display_layer, new_display_layer) in enumerate(zip(self.display_layers, value)):
             for property in properties:
                 property_value = getattr(new_display_layer, property)
--- a/nion/swift/model/DisplayItem.py
+++ b/nion/swift/model/DisplayItem.py
@@ -1623,6 +1623,7 @@ class DisplayLayer(Schema.Entity):
     stroke_color = Schema.EntityAttribute[typing.Optional[str]]()
     fill_color = Schema.EntityAttribute[typing.Optional[str]]()
     stroke_width = Schema.EntityAttribute[typing.Optional[int]]()
+    graph_style = Schema.EntityAttribute[typing.Optional[str]]()  # "bar" | "line"
     display_data_channel = Schema.EntityAttribute[typing.Optional[DisplayDataChannel]]()
 
     def __init__(self, display_layer_properties: typing.Optional[Persistence.PersistentDictType] = None) -> None:
@@ -1684,6 +1685,7 @@ class DisplayLayerInfo:
     stroke_color: str | None
     fill_color: str | None
     stroke_width: int | None
+    graph_style: str | None = None  # line-plot style: "bar" | "line" (None = default "bar")
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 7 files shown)


cmeyer:
Can you file an issue to go along with this with motivation for the change? Also, I think scatter plots are fundamentally different than line plots of this style since they take a different data shape, right? If so, there are a couple references to scatter plots in this PR that should be removed.

Tiomat85:
Will do on the Issue.  Scatter plots are odd, since depending on how they all go it could be almost a one line change in here (where the x coordinate for each point is calculated).  I will remove all references to scatter for this though as they aren't relevant but when we come around to trying to create a scatter implementation the line segment calculator would be a good start point.

Tiomat85:
The more I think about Scatter, the more I think it is mixing two concepts in one. 

Concept one is basically just the graph_style (or whatever alternative name we go if we come up with something better) that is essentially how you draw the lines between the data points.  Bar mode draw a horizontal line from the data point, to the right, until it reaches the X value of the next data point. Then draws a vertical line to that data point.  Line mode draws a straight line from a data point to the next data point to the right.  Other modes could be added as extensions (Point only, spline for smoothly varying curve etc etc).

Concept two is actually the non-uniform X axis that has been discussed elsewhere and, when mixed with a point-only gives the 'natural' scatter plot.  This can be an additional follow-on task but there would need to be some discussion on whether there are any additional questions on that.  A quick one that comes to mind for that would be for the bar plot whether there is any world where you would actually be plotting the area of the bar instead of the height.  Will make a new issue relating to this, but not plan on doing this in the immediate term.

Brow71189:
I 100% agree with that. This is a common distinction in other software as well, for example Excel. They call it "bar graph", "line graph" and "scatter graph", although I'm not saying that these are the names we should be using, just that the categories make sense.

Tiomat85:
Meeting discussion points:
Bar+Line graph alignment matters
Trial offsetting the bar graph by half a pixel such that the bar graph is then cantered on the point rather than being a line to the right from the point.
Investigate feasibility of doing a per-layer selection rather than per-graph so you can plot a line graph on top of a bar graph.

Feature Creep:
Function based plotting rather than just from data items - Is this just a graph annotation?

cmeyer:
Follow-up on the function-based plotting (this really belongs in the issue, but I'll put it here for now): we can do that with a generative computation instead. No need to have explicit graph support for it.

nion/swift/model/DisplayItem.py
```diff
@@ -1623,6 +1623,7 @@ class DisplayLayer(Schema.Entity):
     stroke_color = Schema.EntityAttribute[typing.Optional[str]]()
     fill_color = Schema.EntityAttribute[typing.Optional[str]]()
     stroke_width = Schema.EntityAttribute[typing.Optional[int]]()
+    graph_style = Schema.EntityAttribute[typing.Optional[str]]()  # "bar" | "line"
```
KRLango: I assume it is being done to match the surrounding code, but shouldn't these be `str | None`?

nion/swift/model/LinePlotDisplay.py
```diff
@@ -390,18 +390,20 @@ def __init__(self,
                  axes: typing.Optional[LineGraphAxes],
                  fill_color: typing.Optional[Color.Color],
                  stroke_color: typing.Optional[Color.Color],
-                 stroke_width: typing.Optional[float]) -> None:
+                 stroke_width: typing.Optional[float],
+                 graph_style: typing.Optional[str] = None) -> None:
```
KRLango: I assume it is being done to match the surrounding code, but shouldn't these be `float | None` and `str | None`?

Tiomat85:
As discussion alignment now is centered in the bar, with it not extending axes so the first bar is half-width with default rendering.  Type is at layer level so you can have some lines and some bars.

I have ignored the function version for the time being as I think that is a very different question and not within scope of this.

nion/swift/model/LinePlotDisplay.py
```diff
@@ -606,6 +614,13 @@ def axes(self) -> LineGraphAxes:
             left_channel = typing.cast(int, min(left_channel, right_channel))
             right_channel = typing.cast(int, max(left_channel, right_channel))
 
+            # if any display layer uses the centered "bar" style, extend the drawn range 0.5 channel to the
+            # left so the leftmost bar (which is centered on its sample position, spanning [c-0.5, c+0.5)) is
+            # fully visible rather than being clipped at the left edge of the plot.
+            uses_bar_style = any(display_layer.graph_style == "bar" for display_layer in self.display_layers[0:MAX_LAYER_COUNT])
+            drawn_left_channel: float = left_channel #left_channel - 0.5 if uses_bar_style else left_channel
+            drawn_right_channel: float = right_channel #right_channel + 0.5 if uses_bar_style else right_channel
+
```
cmeyer: Can we just always extend the drawn range?

cmeyer:
This code eliminates the decimation algorithm and switches to sampling. Here is a demonstrator file, compare the displays old main branch vs. this PR when displayed around 500px wide.

AI assisted:
```
def make_decimation_test_data(display_width_px: int = 500, samples_per_pixel: int = 40) -> numpy.ndarray:
    """Build a 1D array sized so ~samples_per_pixel data points map to every screen pixel
    (display_width_px total), specifically to expose the difference between an
    area-weighted rebin (one blended value per pixel -- "keeps" every pixel) and a
    per-sample renderer that emits one vertex per sample with no decimation ("throws out"
    pixel-level aggregation, relying on overdraw).
    Four equal-length regions:
      1) flat baseline      -- sanity check: both renderers should render this identically.
      2) single-sample spike -- one sample at 50x baseline amid otherwise flat data.
                                A rebin will smear/attenuate it by ~1/samples_per_pixel
                                (nearly invisible). A raw per-sample renderer draws it as
                                an explicit vertex -- it may show full-height or vanish
                                depending on stroke overdraw order, either way diverging
                                from the rebin's smoothed result.
      3) near-Nyquist square wave -- alternates +A/-A every single sample. A rebin
                                correctly averages this down to ~baseline (there's no way
                                to see oscillation this fast at this zoom level). A raw
                                per-sample renderer instead draws thousands of full-amplitude
                                overlapping segments into a handful of pixel columns
                                -- classic aliasing, and a visibly different (much "busier"
                                or solid-looking) band compared to the rebin output.
      4) slow sine          -- varies slowly relative to the pixel grid; both renderers
                                should produce nearly identical output here, serving as a
                                control region.
    """
    n_total = display_width_px * samples_per_pixel
    quarter = n_total // 4
    baseline = numpy.full(quarter, 10.0)
    impulse = numpy.full(quarter, 10.0)
    impulse[quarter // 2] = 500.0  # single-sample spike, 50x baseline
    square = 10.0 + 10.0 * numpy.where(numpy.arange(quarter) % 2 == 0, 1.0, -1.0)
    x = numpy.linspace(0.0, 6.0 * numpy.pi, quarter)
    slow_sine = 10.0 + 8.0 * numpy.sin(x)
    return numpy.concatenate([baseline, impulse, square, slow_sine])
```

cmeyer:
There are pixel edge problems:

From console:
```
show(numpy.random.randn(4))
```

cmeyer:
This PR is temporarily on hold while we review requirements beyond the `make_decimation_test_data` and evaluate those requirements against the various possible algorithms. Restated: we need to figure out what is the correct behavior - I'm not sure the old algorithm is desired; but equally not sure what a new algorithm should be.

Roughly, as I read it, the old algorithm does decimation but can lose contributions from narrow peaks (i.e. filtering); the new algorithm uses all points and may suffer performance issues. Also, my proposed "all contributions" decimation in #1718 is another alternative to evaluate against requirements. There are other possible algorithms (currently exploring those, too).

PR #1934, opened 2026-08-20
Add iterable computation support (FFT only for now)

### LINKED ISSUE #430: Add pixel by pixel processing to map 0d, 1d and 2d operations to collections/sequences
e.g. take FFT of every item in collection.

We might consider options to apply along a row/column and reduce.

- [ ] Ensure rotated crop can be performed [efficiently] on sequences/collections

### See Also
- https://github.com/nion-software/nionswift/issues/304

cmeyer:
Fixes #430

- **Move updated window and scalar functions to JSON files.**
- **Add method to create data item from AnnotatedArray.**
- **Model: add per-input operation semantics and cross-source iteration planning**
- **UI: add input-operation selector for iterable computation sources**
- **Add project migration to/from existing window/scalar functions.**
- **Extract executor iteration into functions.**
- **Add computation API v1 for host/plugin executor registration and typed parameter access.**
- **Add computation v1 executor.**
- **Add FFT based on computation API v1 and annotated array.**

This PR introduces iterable computations by adding per-input operation (describes how each source is reduced for processor compatibility), cross-source iteration planning, and a new computation API/executor stack, then uses that stack to ship the first concrete operation: FFT on computation API v1 + annotated arrays. It also adds the inspector UI for choosing input operations on iterable sources, refactors executor iteration logic to support the new planning model, and includes migration/plumbing commits to move existing window/scalar function definitions and support transition with project migration behavior.

This PR can be tested by creating higher order datasets (i.e. a sequence of 2d images) and applying the FFT function to that data. Then use Edit Computation to edit which axis group (or display) the FFT applies to. In order for ease of testing, add a crop to the sequence of 2d images and do the FFT of the crop. Then the crop can be moved to trigger updates.

~~NOTE: Type checking is failing. The review and testing are still worth it right now. Please proceed. I will fix the type checking in a patch before merging this.~~ Succeeds now.


nion/swift/model/Symbolic.py
```diff
@@ -525,6 +530,112 @@ class ComputationVariableType(enum.Enum):
...
+    m["datum"] = datum_data_metadata
+    return m
+
+
+def is_iterable(data_input: ComputationProcessorDataInput, data_metadata: DataAndMetadata.DataMetadata) -> bool:
+    """Determine if the data described by the data metadata is iterable according to the requirements of the computation processor data input."""
+
+    axis_set_data_metadata_list = get_axis_set_data_metadata_list(data_metadata)
+
+    # if there are fewer than 2 axis sets, then it can't be iterable according to the requirements of the computation
```
Brow71189, line 623: That is a limitation we should remove. Why can't we iterate over one of the data axis for examle? An important application for this would be for angle-reolved EELS, where we need to align the rows in a single 2D image. Currently this is implemented as a special case in EELS->Align ZLP, but it would be great if we could get rid of this special case and cover this usecase with the general iteration system.

Brow71189 (changes requested):
The UI for selecting the iteration axis is a bit confusing:

Image

Which of these options iterate and which ones don't? Also what does "Input operation" mean?

I think it would be better to explicitly offer a dropdown menu called "map to" or "iterate along" and then list the available iteration axis

What could also be useful would be to show the already processed data. Right now it processes everything in the background and only in the end you get a result. Think about a computation where you have to fine-tune some paramters: Users might want to see what they get before the entire thing has finished running and then cancel the computation if something needs changing.
Not showing intermediate results removes one of the key advantages of this new system: Allowing the user to change parameters and check if they are correct without waiting for the entire computation to run through.

Next, performance:
I ran an FFT along the collection axis of a larger dataset (3.5 GB). It runs the FFT within a reasonable amount of time (45 s on my machine), which is also what the Computation editor panel shows as execution time. But then, it starts writing the data to disk, and that took another 3-4 minutes. From observing task manager it looks like it accumulates the entire data in memory first, then does something where the memory usage drops again (is it writing to disk here? Maybe, because there is disk usage going on). Then memory spikes again, so there is another copy happening and then it slowly starts dropping again and eventually swift starts being responsive again and everything is done. And keep in mind that this was a 3.5 GB dataset, so what happens if you do that for 35 GB or 100 GB?

Below is the screenshot of my task manager. The red line close to the left edge is the point where processing was done, and the circle in the middle shows the second memory spike. And then shortly after the arrow on the right edge of the screenshot Swift was done with everything.

Image

I think I found the location of the extra data copy: https://github.com/cmeyer/nionswift/blob/1a6c37506e4e2a24bb5a81b14573845732136d1f/nion/swift/Facade.py#L1015

And [this line](https://github.com/nion-software/niondata/blob/4546a878e599b3e87cf0389ed24264d765a42a00/nion/data/annotated_array/_implementation.py#L654) can also produce another copy, depending on the input data.

cmeyer:
These are all forward thinking comments, but I don't think they change anything in this PR.

"Input Operation" because it is the best term I could think up. It's open for discussion, but "map to" is wrong - you're selecting the axis group which is input to the operation and then collected along the other axis groups; and "iterate along" is closer, but also wrong when using a reduction of the data (i.e. the "Displayed Data") as the input.

We can debate the naming of that UI element offline - but, for now, I'd like to get this PR merged so that I can move on to the other issues you raise, like performance and live updating.

Limiting the "input operation" to an axis group is a simplification to get the initial version implemented. It also simplifies the already-difficult-to-understand UI. There are a few solutions going forward, one of which is to relax that limitation; another is to have pre-operations that can be stacked to split apart axis groups. That's another offline discussion...

cmeyer:
One other consideration... before release I can put this behind a feature flag until we're happy with it.

cmeyer:
Added follow-up issues:
- https://github.com/nion-software/nionswift/issues/1941
- https://github.com/nion-software/nionswift/issues/1942
- https://github.com/nion-software/nionswift/issues/1943
- https://github.com/nion-software/nionswift/issues/1944

cmeyer:
This has been open for several days and is blocking other work, so I'm planning to merge on 2026-08-24. If you see something critical, say so before then - for smaller comments, I'll pick them up in one of the follow-up PR's.

Brow71189:
I would say this definitely needs to go behind a feature flag until we have to most importat problems fixed (performance and UI).

I did not propose to just replace the string "Input operation" with "map along" of course. This would be a different way of specifying what to do. I think it would be clearer for users if they could select between "Do not map" and "Map along collection axis" than "Process display data" and "Process data". Obviously selecting "map along collection" would run your "process datum axis" option.

We also need to fix the performance problems I highlighted soon, otherwise this will be unusable. I made another performance test in a fresh Swift project with the same dataset, just to see if that made a difference. No the save performance is a lot better but the read performance is terrible now in certain situations, so maybe there was some caching going on in my initial test. Anyway, the new performance tests hare summarized here: https://github.com/nion-software/nion-internal/issues/359

I made an attempt at an improved chunking algorithm here: https://github.com/nion-software/nionswift/pull/1945 which is something we should consider when going for iterable computations.

cmeyer:
I would not put it behind a feature flag until a release.

PR #1938, opened 2026-08-21
Compressed nhdf export

Brow71189:
This PR adds a new nhdf export option which allows to store the output file as lz4 compressed file. Because the inital performance with just comression enabled was terrible, it also adds an improved handling of hdf5 chunking: If the input data item is already backed by and hdf5 file and uses chunks, the output file will use the same chunking layout. This dramatically improves compressed  export performance and even slightly improves uncompressed export performance.
Finally this PR adds logging output for nhdf imports so that it becomes easier to compare compressed vs. uncompressed imports.
I will open a follow-up issue for improving import performance by improving how chunks are handled.

Benchmark results for a ~37 GB dataset with real Dectris data (~10% of the pixels have counts in them):

Without https://github.com/nion-software/nionswift/pull/1938/commits/0d561a578263809bcaae32ca203d4e5d675f8483 (chunking improvements):

```
Export 138s (D:\4D Data for march 1st heterostructure 4d .nhdf) C(512 x 512) x D(192 x 192), Real (32-bit)
Export 768s (D:\4D Data for march 1st heterostructure 4d compressed.nhdf) C(512 x 512) x D(192 x 192), Real (32-bit)
```

With chunking improvements:

```
Export 127s (D:\4D Data for march 1st heterostructure 4d .nhdf) C(512 x 512) x D(192 x 192), Real (32-bit)
Export 126s (D:\4D Data for march 1st heterostructure 4d compressed.nhdf) C(512 x 512) x D(192 x 192), Real (32-bit)
```

Which means that with chunking improvements, there is no speed penalty for compressed export, but the data size goes from 37 GB down to 1.5 GB, which is a 25x file size reduction!

A note on [hdf5plugin](https://hdf5plugin.readthedocs.io/en/stable/index.html): Yes, this introduces a new dependency, but it is a small library with no additional dependencies itself. It is mostly a convenience library because it compiles and registers the HDF5 compression filters and provides convenience classes for setting up compression. It does not develop or host the compression libaries themselves, they are all hosted by the hdf5group [here](https://github.com/HDFGroup/hdf5_plugins). That means, that even if `hdf5plugin` causes us issues in the future, it is easy to compile and register the relevant filter plugins ourselves and use them from h5py. However, right now the risk of using hdf5plugin is very low in my opinion and it makes development a lot easier if we want to experiment with different compression algorithms.

### DIFF
--- a/nion/swift/model/HDF5Handler.py
+++ b/nion/swift/model/HDF5Handler.py
@@ -313,10 +313,17 @@ class HDF5HandlerFactory(StorageHandler.StorageHandlerFactoryLike):
 
 
 def _create_dataset(data_group: typing.Any, dataset_id: str, data: numpy.typing.NDArray[typing.Any], filter: h5py.filters.FilterRefBase | None = None) -> typing.Any:
+    chunks: tuple[int, ...] | None | bool = getattr(data, "chunks", True)
     data_shape = data.shape
 
     # create the dataset, preallocate space.
-    ds = data_group.create_dataset(dataset_id, shape=data_shape, dtype=data.dtype, compression=filter)
+    ds = data_group.create_dataset(dataset_id, shape=data_shape, dtype=data.dtype, compression=filter, chunks=chunks)
+
+    if callable(getattr(data, "iter_chunks", None)):
+        # if the data has an iter_chunks method, use it to write the data in chunks.
+        for selection in getattr(data, "iter_chunks")():
+            ds[selection] = data[selection]
+        return ds
 
     # search for the chunk size by iterating backwards through the shape and finding the
     # largest chunk size that is less than 64MB.
--- a/nion/swift/model/HDF5Handler.py
+++ b/nion/swift/model/HDF5Handler.py
@@ -6,9 +6,11 @@ from __future__ import annotations
 import datetime
 import io
 import json
+import logging
 import os
 import pathlib
 import threading
+import time
 import typing
 import uuid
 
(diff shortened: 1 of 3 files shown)


cmeyer:
How does the option get presented to the user?

cmeyer:
This PR is addressing this issue:
- https://github.com/nion-software/nionswift-io/issues/39

We can discuss requirements in that issue before evaluating this PR.

Brow71189:
> How does the option get presented to the user?

I just added a new file format to the export dialog for now, since that was the easiest way of getting this option in:

grafik

As far as I know we don't have an export options UI so a new entry in the dropdown menu seemed like a good solution for now.

KRLango (approved):
Everything looks good to me, but I would be wary of adding this without the improved chunking based on the timings.

cmeyer:
The extra format in the menu makes the choice explicit.

I continue to be highly uncomfortable with adding core dependencies on small Python libraries. `h5py` itself is already problematic (search their issues for bugs filed by me) and adding secondary libraries like `hdf5plugin` on top of that is equally problematic.

I propose making this initial compression capability into a plug-in instead of directly embedded in nionswift. I can do that work - it will essentially be adding the proper hooks to nionswift and then packaging the code from this PR into the plug-in.

Any objection to that path, at least until we gain experience with compression compatibility and that library in particular? The net effect will be identical - but doing it as a package gives us more flexibility on the installer side.

PR #1950, opened 2026-09-03
Change SVG export dialog to support bitmap exports w/ PPI

### LINKED ISSUE #1852: Exporting with scale bar to bitmap format (PNG, TIFF)
We've had a request to be able to export display objects to png format in a way that includes the scale bar embedded. This would produce a similar output to the export SVG, but obviously in png format and not vectorised so any graphics would have to be embedded in. Primary motivation is to be able to easily share such images in a standard image format.

cmeyer:
Fixes #1852

- **Add export_bitmap_file function; allow font-size override.**
- **Change SVG export dialog to support bitmap exports w/ PPI.**

AI assisted (dialog).

This is a step to update the existing dialog; follow-ups will consolidate the various export workflows, make exporting modeless (and threaded), etc.

Note: the font size is fixed at 1/6 PPI in this version. PPI for SVG is 96.

The menu item is renamed from Export SVG... to Export SVG/Bitmap... in this PR.

image

### DIFF
--- a/nion/swift/DisplayPanel.py
+++ b/nion/swift/DisplayPanel.py
@@ -3503,3 +3503,23 @@ def preview(drawing_metrics: UISettings.DrawingMetrics, display_style: UISetting
                 display_canvas_item.update_canvas_items()
                 display_canvas_item.repaint_immediate(drawing_context, pixel_shape)
     return drawing_context
+
+
+def get_drawing_context_for_export(ui_settings: UISettings.UISettings, display_item: DisplayItem.DisplayItem, display_shape: Geometry.IntSize, ppi: float, font_ppi: int) -> DrawingContext.DrawingContext:
+    # take a snapshot so to modify the display properties for the proper image zoom, position, and canvas mode.
+    # ensure that the snapshot is closed when finished.
+    display_item_snapshot = display_item.snapshot()
+    try:
+        # update the zoom, position, and canvas mode for the export. these are neutral values that ensure it will
+        # draw at 1:1 scale and centered in the canvas.
+        display_properties = display_item_snapshot.display_properties
+        display_properties["image_zoom"] = 1.0
+        display_properties["image_position"] = (0.5, 0.5)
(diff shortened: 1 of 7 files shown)


MattRoyle (commented):
LGTM

Ion-e:
Issue and special mentions:

1. No persistence in default selection (Consistently), so making several exports of the same size can be tedious. The only attributes that remain persistent are "SVG/Bitmap image", and "PNG/BMP/JPEG", all other ones reset to default values.
2. When exporting to SVG, the "File format" drop down remains shown, which is misleading as it states the format will be either PNG/JPEG/BMP. There is an inconsistent use in the UI regarding the drop down enable/disable functionality. A disabled drop down should indicate that the values are not editable, not that the values are not going to be used. This behaviour is inconsistent in this same window between the aforementioned File Format drop down and the PPI drop down values.
3. Sometimes when swapping from Bitmap to SVG options, you have to click twice which makes me think the cursor loses focus.
4. It's not clear UI to have the "Custom" option as an option in the "PPI", and then having 2 PPI edit boxes. This should be reworked.
5. The "Units" expander is not intuitive to be referring to "Width/Height", the UI should be reworked to reflect this, potentially placing the units first as a user is more likely to first select the desired units then select the unit values as standard workflow.
6. Units drop down and PPI dropdown have different lengths.
7. Sometimes, when swapping from SVG to Bitmap, the PPI drop down doesn't become enabled automatically, the user has to manually change the Units drop down for it to become enabled.
8. Export as and File format dropdowns could be merged to "Export as: SVG/BItmap PNG/Bitmap BMP/Bitmap JPEG"
9. We could change the "Custom" option in the drop down for aspect ratio by adding the Superscan Width/Height button:

Image

10. How were the Default PPI values chosen?
11. The custom PPI edit box and PPI drop down could be merged into 1.
12. The "Default (5:3)" can be changed to "(5:3)".


After this issues have been addressed, a new round of testing should be done on this feature. Thanks

PR #1963, opened 2026-09-14
Migrate window functions to new computations; add inspector variant support

### LINKED ISSUE #322: Add computation variant group for grouping similar functionality
The UI would allow users to swap a computation to another member of the variant group without destroying the computation tree. This requires that the only difference between variants is in parameters, not in inputs or outputs.

An example would be filters - the user should be able to switch between Gaussian, Sobel, Laplace without redoing the computation.

Image

cmeyer:
Fixes #322

- **Add executors and use niondata primitives for window functions.**
- **Add variant-group switching for interchangeable window function processing operations**

image


cmeyer:
Merged f874ab48aa7fb4f0c28495851f30982e8d6d21b8 609e29a734de50e81b41c96ef5cdf4af23dc8f94

PR #1958, opened 2026-09-09
Add detail browser option to data panel and display panels

cmeyer:
Fixes #297

- **Add details browser to data panel.**
- **Add detail browser view style to display panel**

This is an initial, minimal version of this feature.

image

A follow-up is to remember the choices for the data panel.

Another possible consideration for the future is to make the details more table-like (adjustable widths, more column options, sorting options, etc.) when there is enough space.

This PR attempts to work within the bounds of what is possible now.

PR #1967, opened 2026-09-22
Improve data panel performance with DisplayItemsModel and lazy browser views

cmeyer:
- **Fix stale ThumbnailSource docstring that no longer matches its lifecycle**
- **Improve data panel performance by replacing FilteredListModel use with bespoke DisplayItemsModel**
- **Lazily build data panel grid view to avoid duplicating list view cost**
- **Make data panel list view lazy too, so no browser type is a special case**
- **Store data panel browser views in dicts keyed by type instead of per-type ivars**

Also makes data panel choice persistent.

This is mostly performance and code simplification; but it now serves as a baseline for
- #1958


Ion-e:
Does this needs testing?

